PR #45, opened 2026-05-21
fix(ci): unblock x86 clippy gate

timtoole02:
## Summary
- allow unused x86 AMX FFI declarations while the retained route is not wired
- allow Clippy MSRV lint on the AVX512-VNNI helper that is runtime feature-gated and built with the repo toolchain
- scope the mac-only scheduler counter env test to macOS aarch64

## Verified
- Ubuntu x86 same-host: `cargo fmt --all -- --check`
- Ubuntu x86 same-host: `cargo clippy --all-targets --all-features -- -D warnings`
- Ubuntu x86 same-host: `cargo test --all-targets --all-features`
- Ubuntu x86 same-host: `cargo doc --no-deps --all-features`

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -43,6 +43,7 @@ use crate::{
 use crate::tensor::Q8_0AmxPackedBlock;
 
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+#[allow(dead_code)]
 unsafe extern "C" {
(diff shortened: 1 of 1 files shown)


timtoole02:
Closing as superseded. The cleaner clippy fix is already on `main`, and the latest integrated work landed through #47 with green CI.

PR #86, opened 2026-05-22
perf(q8): route FFN decode chain through VNNI down

timtoole02:
## Summary
- Lets the default-off x86 FFN decode-chain path use the existing FFN-down VNNI decode consumer when that VNNI gate is also enabled.
- Preserves fallback to the packed-rows4 decode consumer for missing CPU support, bad widths, or missing VNNI sidecar storage.
- Adds a Linux x86_64 unit test proving the chain records and uses the VNNI down route, while existing VNNI parity tests keep covering rows4 equivalence.

## Ubuntu x86_64 validation
Canonical host fresh clone under `/home/ubuntu/work/camelid-cron-0719640b-ffn-chain-vnni-down`, head `a53fb525bf6564ddae82bace47586c972a09639c`:

```text
Linux x86_64
cargo 1.95.0 (f2d3ce0bd 2026-03-21)
cargo fmt --check
cargo test q8_ffn_down_vnni --lib -- --nocapture
  6 passed; 0 failed
cargo test x86_q8_ffn_decode_chain_can_use_vnni_down_consumer --lib -- --nocapture
  1 passed; 0 failed
rc=0
```

No same-host llama.cpp timing comparison was run for this PR; this is a default-off code and parity/control-plane slice, not a throughput promotion.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -8469,15 +8469,98 @@ fn try_x86_q8_ffn_decode_chain_path(
     let down_started = Instant::now();
     let decode_group_chunking = mac_q8_ffn_down_decode_group_chunking_enabled()
         || x86_q8_ffn_down_decode_group_chunking_enabled();
-    let output = q8_0_packed_rows4_single_input_projection_with_decode_chunking(
-        down_route.packed,
-        &quantized_activated.blocks,
-        down_route.output_width,
-        down_name,
-        decode_group_chunking,
-    )?;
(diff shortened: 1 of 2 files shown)


timtoole02:
Closing for PR queue cleanup. This VNNI-down FFN decode-chain slice is non-mergeable against current main and should be reintroduced as a fresh PR only after the route/resolver base is current and same-host validation is planned.

PR #88, opened 2026-05-22
refactor(inference): move attention trace diagnostics

timtoole02:
## Summary
- move attention trace DTOs and trace reconstruction helpers into `src/inference/attention_trace.rs`
- keep the public `inference::LlamaAttentionTrace` API via re-exports
- leave attention/runtime/kernel behavior unchanged

## Validation
- `cargo test --lib attention_trace`
- `cargo test --lib`
- `cargo clippy --lib -- -D warnings`

Note: cargo emitted filesystem hard-link cache warnings in the shared target dir, but all commands passed.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -17,6 +17,7 @@ use serde::Serialize;
 use crate::execution_plan::MAC_Q8_PREFILL_I8MM_MIN_ROWS;
 use crate::metal;
 
+mod attention_trace;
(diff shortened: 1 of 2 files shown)


timtoole02:
Closing for PR queue cleanup. This refactor is non-mergeable against current main after the inference/module churn; reapply as a fresh branch if we still want the attention-trace extraction.

PR #99, opened 2026-05-22
perf(q8): label VNNI rawptr telemetry route

timtoole02:
## Summary
- Record a distinct Q8 schedule route name when the default-off FFN-down VNNI raw-pointer kernel is selected.
- Keep the existing non-rawptr VNNI route name for scalar/non-rawptr runs.
- Add Linux x86_64 test coverage so telemetry proves rawptr route selection.

## Validation
- Linux x86_64 focused route-selection tests passed on the PR branch before closure.
- Same-host bounded parity/timing guardrails passed and proved the rawptr route telemetry key was emitted.
- The measured same-host result rejected any throughput, default-on, or support-promotion claim; this remains default-off route/parity evidence only.

## Closure
The same change was ported onto current `main` after #95-#98 landed and pushed directly as `f1433ed` (`perf(q8): label VNNI rawptr route`). This stale pre-refactor PR branch is closed unmerged.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -8887,6 +8887,16 @@ fn record_q8_ffn_down_vnni_decode_reject(
     );
 }
 
+fn q8_ffn_down_vnni_decode_route_name() -> &'static str {
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+    {
+        if x86_q8_vnni_decode_rawptr_enabled() {
+            return "x86_vnni_decode_rawptr_consumer";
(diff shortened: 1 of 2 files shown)


timtoole02:
Ported the rawptr route telemetry change onto current `main` after #95-#98 landed and pushed it directly as `f1433ed` (`perf(q8): label VNNI rawptr route`). Closing this PR because its branch was based on the old pre-refactor main and the current-main version is now checked in.

PR #103, opened 2026-05-22
perf(q8): make FFN decode chain self-contained

timtoole02:
## Summary
- Make the default-off x86 FFN decode-chain gate imply its gate-up and down decode consumers.
- Add a focused env-plan regression for the implied consumer behavior.
- Retain Llama 3.2 3B Q8 parity/timing evidence for the self-contained chain gate.

## Evidence
- CI run #680 passed.
- One-token parity against llama.cpp passed for the checked prompt.
- Timing remains a negative result: Camelid still trails llama.cpp materially on same-host bounded wall metrics.
- Route telemetry proves the self-contained chain path is actually taken.

No default-on gate, support expansion, or throughput claim is made.


timtoole02:
Cron 95495a91 update (2026-05-23 06:08 UTC): rebased this PR onto current `origin/main` and force-pushed `codex/ffn-chain-implied-consumers-95495a91` from `1bb18bdd566f7c9d875766877f4ca55a4080399b` to `984b866fe54b73d1441930589386a086233506e5`. PR is now mergeable against current main.

CI: run #696 completed successfully for `984b866fe54b73d1441930589386a086233506e5`.

Remote Linux x86_64 benchmark/build validation could not be re-run during this cycle because the validation host was unavailable. No Darwin build/test/benchmark substitute was used, and no fresh same-host timing/parity claim is made.

timtoole02:
Closing during PR queue cleanup. This branch is now conflict-heavy against current `main` after the Q8/refactor/chat readiness merges, and the retained lesson is already represented in later default-off Q8 planning work. If the self-contained FFN decode-chain behavior is still needed, respin as a small current-main PR with fresh validation.

PR #120, opened 2026-05-24
perf(q8): record baseline attention-output route telemetry

timtoole02:
## Summary
- record route telemetry when the baseline `attention_output` packed-rows4 matmul prefill path runs
- populate both route-level and layer-scoped telemetry buckets for the baseline path
- add a focused unit test so future same-host Q8 artifacts can prove which baseline route executed

## Validation
- Remote Linux x86_64: `cargo fmt --all -- --check`
- Remote Linux x86_64: `cargo test --lib q8_attention_output_packed_rows4_matmul`

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -10766,7 +10766,19 @@ fn try_x86_q8_attention_output_packed_rows4_matmul_path(
         return Ok(None);
     }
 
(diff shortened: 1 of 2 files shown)


timtoole02:
Closing as superseded by the newer Q8 route telemetry work, especially #141 and #144.

PR #128, opened 2026-05-24
perf(q8): trace x86 qkv prefill route

timtoole02:
## Summary
- manage the x86 QKV prefill consumer gate through the execution plan so clean worktrees preserve or clear it predictably
- add Q8 scheduler route telemetry and denial coverage for the x86 attention QKV decode and packed-rows4 prefill routes
- add unit coverage for the managed gate and the new QKV route telemetry behavior

## Validation
- `cargo fmt --all --check`
- `cargo test ubuntu_experimental_validated_gates_select_rust_avx2_q8_path --lib`
- `cargo test ubuntu_experimental_preserves_explicit_x86_q8_gate_opt_ins --lib`
- `cargo test planner_env_apply_clears_stale_x86_q8_decode_consumer_flags --lib`
- `cargo test q8_attention_qkv_prefill_route_resolver_records_route_and_denials --lib`
- `cargo test q8_attention_qkv_packed_rows4_matmul_matches_runtime_packed_baseline_for_prefill --lib`
- `cargo build --release`

## Same-host evidence
- Baseline retained AVX2 lane on the Ubuntu x86_64 validator stayed at about `519.55 ms` TTFT / `519.81 ms` total for the repeated-prompt harness.
- The shared-QKV prefill gate initially looked only slightly slower there, but the new telemetry showed the prefill route fired during warmup only; the measured runs stayed decode-only because the harness reused the same prompt shape.
- A unique-prompt rerun forced the prefill route into the measured set and rejected it clearly:
  - baseline: about `3301.72 ms` TTFT / `3302.10 ms` total / `2796.67 ms` backend first-content
  - candidate: about `4243.23 ms` TTFT / `4243.63 ms` total / `3726.00 ms` backend first-content
  - measured route attribution: candidate recorded `84` `attention_qkv.packed_rows4_matmul_prefill` calls across measured runs; baseline recorded none
- Guardrail marker passed on every measured Camelid and llama.cpp run in both A/Bs.

## Decision
- Keep `CAMELID_X86_Q8_ATTENTION_QKV_PREFILL_CONSUMER` default-off.
- Retain the telemetry slice for future scheduler/owner experiments because it turns the old route-attribution gap into concrete measured evidence.


timtoole02:
Closing as superseded by #144 and #141, which keep the corrected QKV prefill owner gate and named denial telemetry as the current review targets.

PR #137, opened 2026-05-24
perf(q8): hoist packed rows4 avx2 decode accumulators

timtoole02:
## Summary
- route packed rows4 decode projections through a decode-specific AVX2 hoist helper
- reuse the same in-register AVX2 accumulate/reduce helper for the packed rows4 matmul hoist path
- keep the route structure and default-off gate posture unchanged while removing per-block scalar lane updates from the hoisted AVX2 path

## Validation
- `cargo fmt --check`
- `cargo test --lib x86_q8_avx2_packed_rows4_hoisted_matmul_matches_scalar_dot`
- `cargo test --lib x86_q8_avx2_packed_rows4_decode_hoist_projection_matches_scalar_dot`

## Notes
- Same-host Linux x86_64 timing/parity is still pending for this branch.
- This PR is a bounded backend slice only; it does not widen support claims or change default behavior.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -13,6 +13,11 @@ use std::sync::OnceLock;
 use rayon::prelude::*;
 use serde::Serialize;
 
+#[cfg(target_arch = "x86")]
+use std::arch::x86::{__m128i, __m256i};
+#[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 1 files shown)


timtoole02:
Closing as superseded by #138, which carries the retained AVX2 decode-hoist evidence on a cleaner branch.

PR #140, opened 2026-05-25
fix(q8): wire FFN decode-chain planner and packed-output diagnostics

timtoole02:
## Summary

- enable `CAMELID_X86_Q8_FFN_GATE_UP_DECODE_CONSUMER` and `CAMELID_X86_Q8_FFN_DOWN_DECODE_CONSUMER` automatically when `CAMELID_X86_Q8_FFN_DECODE_CHAIN=on` is selected by the Linux x86 experimental planner
- let output-projection diagnostics decode runtime-packed tied-output Q8 rows so the experimental packed-output routes can still be inspected precisely
- count `ffn_gate_up_decode_consumer_taken` so decode-chain and split gate/up consumer runs leave explicit schedule telemetry

## Why this change exists

The FFN decode-chain fast path already hard-requires both `runtime_plan.q8.ffn_gate_up_decode_consumer` and `runtime_plan.q8.ffn_down_decode_consumer`, but the Linux x86 experimental planner previously treated the decode-chain flag and those required decode-consumer legs as unrelated knobs. That made the decode-chain opt-in silently miss the intended route unless callers also knew to set the extra dependent gates.

The runtime-packed output diagnostics slice closes a separate observability gap on the same lane: once tied output is materialized as runtime-packed Q8 storage, output-projection diagnostics need to be able to reconstruct token rows from that packed representation instead of requiring dense values or file backing.

The follow-up telemetry slice closes a smaller proof gap on top of that work: once the planner implies the gate/up leg and the decode chain executes, schedule telemetry should record that the gate/up decode consumer actually fired instead of leaving only the decode-chain aggregate counter.

## Validation

- `cargo test --lib ubuntu_experimental_ffn_decode_chain_enables_required_gate_up_and_down_legs`
- `cargo test --lib x86_q8_ffn_decode_chain_is_default_off_and_matches_split_consumers`
- `cargo test --lib q8_ffn_gate_up_consumer_matches_runtime_packed_baseline`
- `cargo test --lib output_projection_diagnostics_support_runtime_packed_tied_output_rows`
- Existing Ubuntu Linux x86_64 same-host Llama 3.2 3B Q8_0 branch evidence remains bounded route/timing evidence:
  - Camelid TTFT `3137.54 ms`, total `3138.06 ms`
  - llama.cpp TTFT `316.21 ms`, total `504.97 ms`
  - marker guard passed
  - route proof: `ffn_decode_chain_taken=168`, `ffn_down_decode_consumer_taken=168`, `logits.x86_output_decode_owner.calls=6`

## Public-repo privacy check

- No private hostnames, SSH commands, user home paths, key paths, or local validation details included

## Support-contract check

- This change does not widen support claims
- Evidence remains bounded to the exact Ubuntu Linux x86_64 Llama 3.2 3B Q8_0 same-host lane

## Evidence / artifacts

- Focused planner and diagnostics tests for the decode-chain planner wiring, runtime-packed tied-output inspection, and gate/up decode-consumer telemetry
- Existing Ubuntu same-host timing artifact for the decode-chain-only opt-in case

## Docs impact

- None

### DIFF
--- a/src/execution_plan.rs
+++ b/src/execution_plan.rs
@@ -517,17 +517,28 @@ fn select_linux_x86_q8_plan(
         "CAMELID_X86_Q8_FFN_GATE_UP_DECODE_PAIRED_DOT",
         optional_x86_q8_gate("CAMELID_X86_Q8_FFN_GATE_UP_DECODE_PAIRED_DOT"),
     );
+    let ffn_decode_chain_enabled = env_flag_enabled("CAMELID_X86_Q8_FFN_DECODE_CHAIN");
     env_updates.insert(
         "CAMELID_X86_Q8_FFN_DECODE_CHAIN",
-        optional_x86_q8_gate("CAMELID_X86_Q8_FFN_DECODE_CHAIN"),
+        if ffn_decode_chain_enabled {
+            Some("on")
+        } else {
+            Some("off")
+        },
     );
     env_updates.insert(
         "CAMELID_X86_Q8_FFN_GATE_UP_PACKED_ROWS4_MATMUL",
         optional_x86_q8_gate("CAMELID_X86_Q8_FFN_GATE_UP_PACKED_ROWS4_MATMUL"),
     );
+    let ffn_down_decode_consumer_enabled =
+        ffn_decode_chain_enabled || env_flag_enabled("CAMELID_X86_Q8_FFN_DOWN_DECODE_CONSUMER");
     env_updates.insert(
         "CAMELID_X86_Q8_FFN_DOWN_DECODE_CONSUMER",
-        optional_x86_q8_gate("CAMELID_X86_Q8_FFN_DOWN_DECODE_CONSUMER"),
+        if ffn_down_decode_consumer_enabled {
+            Some("on")
+        } else {
+            Some("off")
+        },
     );
     env_updates.insert(
         "CAMELID_X86_Q8_FFN_DOWN_PACKED_ROWS4_MATMUL",
@@ -574,6 +585,13 @@ fn select_linux_x86_q8_plan(
         "CAMELID_X86_Q8_OUTPUT_DECODE_OWNER",
         optional_x86_q8_gate("CAMELID_X86_Q8_OUTPUT_DECODE_OWNER"),
     );
+
+    if ffn_decode_chain_enabled && !env_flag_enabled("CAMELID_X86_Q8_FFN_DOWN_DECODE_CONSUMER") {
+        reasons.push(
+            "FFN decode-chain opt-in also enables the required FFN-down decode consumer gate"
+                .into(),
+        );
+    }
     reasons.push("validated Ubuntu/Linux x86_64 Rust Q8 runtime repack enabled".into());
     reasons.push("validated Rust AVX2 Q8 packed rows4 kernel selected".into());
     reasons.push(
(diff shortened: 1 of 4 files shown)


timtoole02:
Validation note: the fresh Ubuntu same-host artifact on this branch is route/timing evidence only. `ffn_decode_chain_taken=112` and TTFT improved versus the prior missing-gate run, but the recorded marker guard stayed false, so this should not be read as deterministic-output parity or any support-contract promotion.

timtoole02:
Additional scope note: this branch fixes the FFN decode-chain planner dependency only. The same-host run kept the output-owner env enabled but recorded `output_decode_owner_calls=0`, so any output-owner routing follow-up remains separate work.

timtoole02:
Fresh Ubuntu Linux x86_64 rerun on branch head `4def58c`:

- focused planner regression passed: `cargo test --lib ubuntu_experimental_ffn_decode_chain_only_opt_in_also_enables_down_consumer`
- shared-target release rebuild passed
- same-host Llama 3.2 3B Q8_0 rerun (`warmup=1`, `repeats=2`, `max_tokens=4`, unique prompt) stayed route/timing evidence only:
  - Camelid TTFT `3046.76 ms`, total `3047.05 ms`
  - llama.cpp TTFT `314.90 ms`, total `428.38 ms`
  - marker guard stayed false in both measured Camelid runs and both measured llama.cpp runs

Additional exact blocker from a follow-up dense-diagnostics probe on the same head: the request failed with `output_projection_diagnostic_failed` because the output-projection diagnostic path rejected the tied/file-backed output tensor shape `[3072, 128256]` with zero in-memory values. So this branch still should not be read as deterministic-output parity or any support/perf promotion, and dense output-route proof needs a separate diagnostics fix before it can be used as public evidence.

timtoole02:
Follow-up on branch head `8f2e87b`:

- pushed `fix(q8): support runtime-packed output diagnostics`
- focused Ubuntu Linux x86_64 checks passed on the pushed head:
  - `cargo test --lib ubuntu_experimental_ffn_decode_chain_enables_required_down_leg`
  - `cargo test --lib output_projection_diagnostics_support_runtime_packed_tied_output_rows`
  - shared-target release build
- the dense one-token parity probe that previously died with `output_projection_diagnostic_failed` now completes on the exact Llama 3.2 3B Q8_0 row:
  - prompt tokens match: `true`
  - generated tokens match: `true` (`[66]` vs `[66]`)
  - generated text match: `true` (`"c"`)
  - output-projection diagnostics returned `16` rows; first row layout was `token_major`
- fresh same-host Ubuntu x86_64 rerun on the same head (`warmup=1`, `repeats=2`, `max_tokens=8`, unique prompt) remains route/timing evidence only:
  - Camelid TTFT `3137.54 ms`, total `3138.06 ms`
  - llama.cpp TTFT `316.21 ms`, total `504.97 ms`
  - marker guard passed
  - route proof in the measured run: `ffn_decode_chain_taken=168`, `ffn_down_decode_consumer_taken=168`, `logits.x86_output_decode_owner.calls=6`

This turns the earlier dense-diagnostics crash into real output-projection evidence, but it does not change the support or performance posture: the branch still trails llama.cpp materially on the same host.

timtoole02:
Merge blocker: the latest bounded same-host evidence in this PR still proves `ffn_decode_chain_taken`, `ffn_down_decode_consumer_taken`, and `logits.x86_output_decode_owner.calls`, but it does not yet record `ffn_gate_up_decode_consumer_taken` on the branch head that adds that counter.

Because this PR explicitly adds `ffn_gate_up_decode_consumer_taken` to close the gate/up proof gap, the Ubuntu same-host rerun should show that exact counter firing on the latest head before this merges. As currently published, the branch proves planner wiring plus output-diagnostics recovery, but not the new gate/up telemetry claim.

timtoole02:
`ffn_gate_up_decode_consumer_taken` is no longer measuring what its name and the PR body claim.

Exact issue:
- `origin/main` only records the split gate/up decode-consumer route via `output_projection_by_route["ffn_gate_up.decode_consumer"]`.
- This branch adds `add_q8_schedule_counter(&Q8_SCHED_FFN_GATE_UP_DECODE_CONSUMER_TAKEN, 1)` in the split route, which is reasonable.
- But it also adds the same counter increment inside `try_x86_q8_ffn_decode_chain_path(...)` immediately before logging `output_projection_by_route["ffn_gate_up.decode_fused_activation"]`.
- The new test change makes that conflation explicit by changing the decode-chain telemetry assertion from only `ffn_decode_chain_taken == 1` on `origin/main` to also expecting `ffn_gate_up_decode_consumer_taken == 1` on this branch.

Why that blocks acceptance:
- The counter name and PR summary say this is counting gate/up decode-consumer hits.
- The decode-chain path is a different route and is already separately counted as `ffn_decode_chain_taken`, with its gate/up leg logged under `ffn_gate_up.decode_fused_activation`.
- After this patch, any evidence bundle that reports `ffn_gate_up_decode_consumer_taken` will overcount by including fused decode-chain activations, so the metric can no longer distinguish split consumer usage from decode-chain usage.

Please either:
1. keep `ffn_gate_up_decode_consumer_taken` exclusive to `decode_consumer`, or
2. rename/restructure the telemetry so the split consumer and fused decode-chain gate/up leg are tracked as separate metrics.

Until that distinction is preserved, the new telemetry claim in this PR is not trustworthy evidence.

timtoole02:
Resolved by merged follow-up #165.

The fix keeps `ffn_gate_up_decode_consumer_taken` exclusive to the split `ffn_gate_up.decode_consumer` route and adds `ffn_gate_up_decode_fused_activation_taken` for the decode-chain fused gate/up leg, so the telemetry claim that blocked this PR is now represented as two separate counters instead of one conflated metric.

PR #156, opened 2026-05-25
perf(q8): trace attention qkv packed route

timtoole02:
## Summary
- add QKV route telemetry names for the x86 decode-consumer and packed-rows4 prefill paths
- record denial reasons when the shared QKV route fails closed before a same-host run
- record successful packed-route telemetry so the Ubuntu same-host lane can prove the specialized prefill route executed

## Validation
- `cargo test q8_attention_qkv_route_policy_records_denials --lib -- --nocapture`
- `cargo test q8_attention_qkv_packed_rows4_matmul_records_route_telemetry --lib -- --nocapture`
- `cargo test q8_attention_qkv_ --lib`

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -7103,6 +7103,15 @@ enum X86Q8AttentionQkvRouteKind {
     PackedRows4Matmul,
 }
 
+impl X86Q8AttentionQkvRouteKind {
+    fn telemetry_name(self) -> &'static str {
+        match self {
+            Self::Decode => "decode_consumer",
+            Self::PackedRows4Matmul => "packed_rows4_matmul_prefill",
+        }
+    }
+}
+
 struct X86Q8AttentionQkvRoute<'a> {
     q_packed: &'a Q8_0PackedRows4,
     k_packed: &'a Q8_0PackedRows4,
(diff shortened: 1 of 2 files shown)


timtoole02:
I found a telemetry regression in this patch.

`try_x86_q8_attention_qkv_decode_consumer_path(...)` and `try_x86_q8_attention_qkv_packed_rows4_matmul_path(...)` now record one elapsed route under `attention_qkv.*` using only `q` / `q.name`. PR149 was emitting separate route hits for `attention_q`, `attention_k`, and `attention_v` with each weight name.

That matters to the same-host evidence path: `scripts/summarize-same-host-stream-timing.mjs` correlates `projection_layer_routes` back to layer hotspots by `route.role`, and `hotspotRolesForProjectionRole(...)` only special-cases `ffn_gate_up`; everything else must already match the hotspot role name exactly. With an aggregate `attention_qkv` role, the summarizer can no longer line up those route hits with the recorded `attention_q` / `attention_k` / `attention_v` hotspots, so the Ubuntu lane loses the per-role proof it needs for Q/K/V route execution.

I would keep the shared denial bucket if you want it, but restore elapsed route records for all three roles separately, using the Q/K/V weight names again.

PR #281, opened 2026-06-14
qwen3-thinking: deterministic full-trace parity investigation (negative result, no claim change)

timtoole02:
## Summary — negative result (no claim change)

Investigates whether Qwen3 thinking-mode **full-trace** token-parity vs the pinned llama.cpp reference can be promoted from the shipped *leading-trace* claim by putting both sides on a shared **deterministic reduction order** (the requested path (a)).

**Verdict: not achievable as a configuration.** The block-level reduction order is already shared and stable on both sides; the residual divergence is fixed per-kernel f32 arithmetic, which is not a reduction-order knob. The leading-trace claim is retained unchanged. **No path-(b) semantic/text-level fallback was used.**

## Evidence (`qa/evidence-bundles/qwen3-1.7b-q8-thinking-deterministic-investigation-*`)

**H1 / mechanism (confirmed, with the kill-condition nuance):**
- **Within-engine reduction-order noise ≈ 0.** The ggml reference's top-2 logit gaps at the divergence positions are **bit-identical at `-t 1` and `-t 8`**; camelid `--deterministic` is bit-exact across runs/threads by design (DECISIONS.md D9). So the divergence is **not** within-reduction-order noise.
- `--deterministic` reproduces the **exact same divergence points (73/27/205/26)** as the cpu_reference path in the shipped bundle — a deterministic shared-order config does not move the cross-engine divergence (block order already shared).
- The divergent-position gaps are small near-ties (reference top-2 gap **0.022–0.180 logit**) but the perturbation flipping them is accumulated cross-kernel f32 difference **far larger than reduction-order noise** (e.g. capital@73: camelid prefill ranks " landmarks" over " sites" by +0.706 vs the reference's +0.166).

**Two mechanisms (teacher-forced at each divergence):**
- **Dominant — camelid decode-vs-prefill (3/4 probes).** Fed the agreed prefix, camelid **prefill reproduces the reference argmax**, but camelid incremental **decode flips it**. camelid's forward pass (prefill) is correct; its autoregressive decode path diverges.
- **Cross-implementation (1/4 — primary-color@27, 0.022-logit razor-tie).** camelid prefill *also* differs from ggml here: fixed camelid-CPU-dot vs ggml-`vec_dot` rounding.

**Why not a config:** the residual is (a) camelid decode kernels ≠ camelid prefill kernels and (b) camelid CPU dot ≠ ggml `vec_dot` — fixed, ISA/implementation-dependent f32 arithmetic. Eliminating it is reimplementation, not configuration. Even unifying decode==prefill leaves the cross-impl razor-tie at primary-color@27, so full-trace parity is unreachable by config regardless.

**Actionable lead** (the real discrepancy the kill-condition points at): camelid **decode-vs-prefill numerical inconsistency** (~0.7 logit by token 73, accumulated over decode steps), which affects all long greedy generation, not only thinking mode. Aligning camelid's decode numerics to its prefill path is the engineering follow-up — out of scope for a reduction-order config.

## Scope

No functionality or parity claim changed. Ledgers add a one-line provenance pointer to this investigation under the existing "full-trace thinking-mode token-parity" rejection. All four public-evidence/scrub checks pass.


timtoole02:
**Correction pushed (544cb58).** A follow-up artifact-free re-measurement falsifies this PR's *decode-vs-prefill* mechanism attribution (the verdict is unchanged and strengthened).

The original "camelid prefill matches the reference, decode flips it (~0.706 logit at capital@73, 3/4 probes)" was a **measurement artifact**: the prefill side was fed a detokenized prefix via `/v1/completions` text input (`parse_special=false`), so `<think>`/`<|im_start|>` tokenized as literal characters → a different context than the real decode KV.

Re-measured with **exact token ids** (`camelid_prompt_token_ids`, bypasses tokenization): **camelid decode == camelid prefill** — capital@73 both pick `" sites"` with a **0.003-logit** gap to `" landmarks"` (not 0.706); primary-color@27 and hello@26 also agree (3/3 captured; 2+2@205 not captured, box flakiness).

**There is no decode-vs-prefill bug.** The thinking-trace divergence vs llama.cpp is entirely camelid-vs-ggml cross-implementation rounding at near-ties — the out-of-scope item. The verdict (full-trace parity not achievable via a shared-order config) stands without a decode-vs-prefill confound. See `correction-decode-vs-prefill-artifact.json`.

PR #406, opened 2026-07-08
feat: hardware-aware model fit advisor (capacity axis)

karan68:
## Summary

- Adds a hardware-aware **model fit advisor** — a *capacity* signal estimating whether the local machine can load/run each catalog model, surfaced across API, WebUI, and CLI. Strictly separate from support/parity.
- `src/fit.rs`: `FitVerdict` + pure `assess()`, reusing `cuda_vram::evaluate` for the VRAM branch and a host-RAM budget **modeled on** the `kv_cache` `max(80% available, 25% total)` policy (independent reimplementation over `HardwareProfile`; documented platform divergences). Degrades to `unknown` (never a failure) when memory can't be probed.
- **Exact pre-download sizing:** `src/fit_dims.rs` (`DimsResolver`) reads each model's real GGUF dimensions from a **12 MiB header range-fetch** — no weight download — and computes an exact KV-cache footprint. Rows carry `fit_confidence: exact | approx | unknown`; the WebUI shows a solid chip for `exact` and a dashed `~` estimate chip for the coarse `approx` fallback (non-dense/unparseable archs, e.g. Gemma-4 MoE, are honestly left `approx`).
- `DimsResolver` is production-hardened: in-flight de-dup, bounded concurrency (semaphore) with curl `--connect-timeout/--max-time`, an RAII drop-guard so a panicked/early-returned fetch can't pin a slot, a debounced single-writer, and a **versioned + TTL + LRU** disk cache with negative caching. `GET /api/models/catalog` is side-effect-free for curated rows (warmed once at startup); HF-search rows schedule ≤5 bounded background fetches.
- WebUI capacity chip (blue "fits" / red "too big" + CPU-chip icon) with advisory "best for" tags; CLI `pull` gains a `FIT (this host)` column; `POST /api/models/load` fail-fasts with a typed **`422 model_too_large_for_host`** (overridable via `CAMELID_SKIP_FIT_CHECK=1`). The load guard runs on a blocking thread (off the async executor).
- **Capacity axis only** — a "fits" verdict never appears in `model_compatibility` and cannot promote any row.

## Why this change exists

Camelid reports which models are supported/runnable but never whether your machine can hold one, so a user can download an 8B and only discover it won't fit via a mid-load OOM or a mid-generation KV abort. This gives an honest, up-front capacity signal (before download and at load time) so users pick a model that actually runs — without touching the support ledger. The runtime `VramShortfall`/KV guards remain the authoritative safety net; this layer only informs.

## Compatibility note

`POST /api/models/load` can now fail-fast with `422 model_too_large_for_host` on a `WontFit` verdict from a **probed** host, where it previously always attempted the load. It never fires on an unprobed host (`Unknown` is never `WontFit`), and `CAMELID_SKIP_FIT_CHECK=1` restores the unconditional load-attempt behavior. The authoritative `VramShortfall`/`KvCache` guards remain the hard net after the load begins.

## Validation

- `cargo fmt --all -- --check` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test --lib` → **710 passed, 0 failed** (incl. `fit::tests`, `api::catalog_fit_tests`, `fit_dims::tests`, and a hermetic GGUF header-prefix parse test)
- `cd frontend && npm run build` — clean
- Gated live network test `CAMELID_TEST_REMOTE_DIMS=1` (Qwen3-0.6B, Llama-3.2-1B) verifies the header fetch end-to-end.

## Evidence / artifacts

- Live on an RTX 4060 (8 GB) / 16 GB host: catalog warmed **11 exact / 4 approx** (the 4 approx are Gemma-4 E2B/E4B/12B/26B-A4B — non-dense metadata, correctly negative-cached, not network failures). A restart served **11 exact from the disk cache in ~1.5 s with zero re-fetch**. Small/mid rows → "fits", 7–8B → "fits (offload)", 12B/26B → "too big" — consistent across API JSON, CLI column, and the WebUI chip.
- Behavior is advisory only; the badge is a static capacity hint (cached startup snapshot), while the load guard re-probes live — documented as an intentional hint-vs-authoritative split.

## Public-repo privacy check

No private hostnames, SSH commands, user home paths, key paths, or raw infrastructure output are included. Remote validation blockers are summarized generically. `bash scripts/check-public-scrub.sh`; `node audit-evidence-bundle-privacy.mjs --strict`.

## Support-contract check

No support wording changed. TinyLlama Q8_0 remains the full-support gate; 1B/3B/8B rows stay within their documented exact-row smoke envelopes. The only ledger edit is a new "Capacity rule" in COMPATIBILITY.md stating the fit verdict is not a support claim.

## Docs impact

- `docs/architecture/MODEL_FIT_ADVISOR_PLAN.md` — end-to-end design/plan (Slices 1–8, incl. the DimsResolver redesign and review-hardening).
- `ROADMAP.md` — new "Model fit advisor (capacity axis)" lane.
- `COMPATIBILITY.md` — new "Capacity rule" in Governing rules.

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -86,6 +86,7 @@ Two rules keep this matrix honest across docs, API signals, and UI copy:
 
 - **Support rule:** nothing adjacent inherits support across model size, quantization, tokenizer lane, API surface, or frontend state.
 - **Credit rule:** visible llama.cpp / ggml acknowledgement and the MIT notice remain part of parity-backed release claims.
+- **Capacity rule:** the model-fit advisor is a *capacity* signal (can this host load/run a row), reported separately from support. A "fits" verdict never implies parity, validation, or support; it does not appear in `model_compatibility` and cannot promote any row. See [`docs/architecture/MODEL_FIT_ADVISOR_PLAN.md`](docs/architecture/MODEL_FIT_ADVISOR_PLAN.md).
 
 README, `STATUS.md`, `/api/capabilities`, and frontend readiness copy should continue to mirror this exact ledger. `/api/capabilities` exposes the same compatibility rows as `model_compatibility`; read each row literally. Metadata parsing does not imply tokenizer parity, tokenizer parity does not imply generation, tensor loading does not imply safe execution, and one supported row must never lend support to adjacent model sizes or quantizations.
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -272,6 +272,14 @@ Portability and packaging should remain explicit:
 - no implied portable model-path assumptions without documentation
 - no release packaging claim before reproducible setup instructions exist
 
+### Model fit advisor (capacity axis)
+
+A **capacity** signal, deliberately separate from the support ledger: it estimates whether the *local machine* can load/run a catalog row, and never implies parity or support. See [`docs/architecture/MODEL_FIT_ADVISOR_PLAN.md`](docs/architecture/MODEL_FIT_ADVISOR_PLAN.md).
+
+- The verdict (`fits_resident` / `fits_with_offload` / `cpu_only_ok` / `wont_fit` / `unknown`) is derived from the detected `HardwareProfile` and the row's on-disk weight footprint; it is advisory, never a support claim, and degrades to `unknown` (never a failure) on hosts whose memory cannot be probed.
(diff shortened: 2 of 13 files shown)


karan68:
@timtoole02  please review

timtoole02:
Merged via #416, which carries this branch's full diff plus two crash-safety fixes from the adversarial review (the 25%-of-total RAM floor overcommit in usable_host_ram_bytes, and the GPU FitsResident branch not consulting host RAM for load-time staging). Your commits are credited via co-author trailer — thanks for the advisor; the DimsResolver range-fetch design in particular is a nice piece of work.

PR #419, opened 2026-07-09
feat: JSON Schema constrained decoding for response format (structured outputs)

karan68:
## **Summary**

Adds JSON-Schema-constrained decoding: `POST /v1/chat/completions` now honors `response_format: {"type":"json_schema", "json_schema":{"schema":{…}}}`, guaranteeing the model output matches a supported subset of JSON Schema — the structured-output capability Ollama/vLLM have and Camelid lacked.

New `SchemaState` in `grammar.rs`: a schema-directed byte-level PDA that reuses the existing `json_object` masking path (same `accepts`/`advance`/`is_done` contract) via a new `ConstraintState` enum, so the decode loop is backend-agnostic.

Fail-closed by design: any schema feature Camelid can't enforce byte-for-byte returns a typed `400` naming the offending keyword — never a silently dropped constraint.

Generalizes the existing bool `json_object_mode` plumbing to `Option<ConstraintSpec>` (compiled/validated at request time, so schema errors surface as a `400` before generation); `json_object` behavior is unchanged.

---

## **Why this change exists**

Structured output (guaranteed-valid, schema-conforming JSON) is table-stakes for agent/tooling/RAG-extraction workflows and is the one high-frequency capability Ollama (structured outputs) and vLLM (guided_json) have that Camelid did not. It's implemented in Camelid's native style — a small, fail-closed, heavily-tested byte automaton that reuses the proven `json_object` constrained-decoding machinery rather than adding a parallel subsystem — so it's a correctness guarantee on the output, matching the project's differentiation.

Supported subset (root object/array; typed properties + required + `additionalProperties:false`; array items; string `enum`/`const`; scalar types). Everything outside it (type unions, `anyOf`/`allOf`/`oneOf`/`$ref`, `pattern`/`format`/`min*`/`max*`, non-string enum, etc.) is rejected with a precise `400`. Tool-call constraining was investigated and deliberately deferred (documented in `docs/architecture/feature.md`): tool-call output format is model-specific and parsed post-hoc, so forcing one shape would regress non-Llama models — it needs per-model grammars, a separate lane.

---

## **Validation**

- `git diff --check` — clean (exit 0, committed range vs main)
- `cargo fmt --all -- --check` — clean
- `cargo test --all-targets --all-features` — `cargo test --lib` → **747 passed / 0 failed / 55 ignored**; `cargo clippy --all-targets -- -D warnings` → **0 warnings**; full `--all-features` (CUDA) run finishing (change is CPU-only grammar/API logic — no CUDA/kernel/inference paths touched)
- `cd frontend && npm run build` — N/A (no frontend changes)
- Other evidence attached below

---

## **Public-repo privacy check**

- No private hostnames, SSH commands, user home paths, key paths, local validation details, raw SSH stderr, or raw infrastructure failure output are included
- Remote validation blockers are summarized generically — N/A (no remote validation involved; all checks ran locally with no infra details included)
- `bash scripts/check-public-scrub.sh` — exit 0
- `node audit-evidence-bundle-privacy.mjs --strict` — N/A (no evidence bundle added by this PR)

---

## **Support-contract check**

- This PR does not overclaim support
- TinyLlama Q8_0 remains the current full-support gate; exact Llama 3.2 1B/3B and Llama 3 8B Q8_0 rows stay limited to their documented exact-row smoke envelopes unless exact new evidence is included — unchanged; no support/parity rows touched
- Any 1B / 3B / 8B wording stays aligned with `COMPATIBILITY.md`, `STATUS.md`, and `/api/capabilities` — N/A (this is a capacity/output-format feature, model-agnostic; adds no model-support claims)

---

## **Evidence / artifacts**

- `cargo test --lib`: **747 passed / 0 failed / 55 ignored** (rebased on current main).
- New unit tests: `grammar::tests` (compile-rejection matrix, required + value-type enforcement, bool/null/number, string enum/const, nested objects+arrays, non-mutating `accepts` masking incl. multibyte tokens, object-key trie, `constraint_spec_builds_and_validates`) and `api::tests::response_format_maps_to_constraint`.
- `cargo clippy --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- `git diff --check`: clean.
- Net diff vs main: `src/grammar.rs` **+1237**, `src/api/mod.rs` **+145/−144**, `docs/architecture/feature.md` **+293**.

---

## **Docs impact**

Adds `docs/architecture/feature.md` — the lane's design doc: authoritative supported-subset contract, invariants preserved (EOG-iff-done, spec/fast-lane disabled under a constraint, `stream:true+constrained → 400`), and the rationale for deferring tool-call constraining.

No `README` / `COMPATIBILITY.md` / `STATUS.md` / `ROADMAP.md` changes — this is a capacity/output-format feature and makes no support or parity claims.


timtoole02:
## Adversarial review (35-agent verified): NOT mergeable as-is — 3 blockers, 6 majors

First: the core automaton is genuinely good work. The `num_advance`/`str_advance` sub-grammar extraction is byte-identical to the old `JsonState` inline logic (cross-checked by test), the enum prefix-free completion reasoning is correct, compile-time 400s are well-shaped, and the decode-loop seam (spec lane + GPU fast lane gated on `grammar.is_none()`, EOG-iff-done, typed error on an all-false mask) holds structurally. Trial-merge onto main `b5567fac` is clean; clippy/fmt/tests all green locally.

But the integration edges break the PR's own headline promise ("guaranteeing the model output matches the schema", "never a silently dropped constraint") in several reachable ways. Every finding below survived an independent adversarial verification pass against the merged code.

### Blockers

**B1 — Prompt-prefix cache bypasses the constraint entirely** (`src/api/mod.rs:9923`, key at `:9723`). The cache-hit gate checks `!collect_dense_diagnostics && !want_execution_trace && !resident_cuda_active` — not `constraint_active` — and the cache key (model_id, model_path, token_ids, sampling) omits the constraint. On a warm hit the first token is sampled from the RAW cached logits with no mask, and the grammar is never advanced over it → output like `Sure{...}` with 200 OK. Retry/regenerate with the same prompt is the *common* warm-cache case, so the guarantee is broken on the second identical request, every time, on the CPU lane. (Pre-existing for `json_object`; this PR makes it the enforcement path for `json_schema`, where clients programmatically depend on shape.) Fix: add `constraint_active` to the cache-bypass gate (cheapest), or key the cache on the constraint.

**B2 — Reachable dead-end: exhausted object accepts `,` then `"` into a state where NO byte is legal** (`src/grammar.rs:1092`, `:832`). `advance_after_value` accepts `,` inside an object unconditionally and `Key` accepts `"` unconditionally, without checking that an unused declared property remains. Once all properties are used, the very common BPE token `,"` is mask-legal — and the next step's mask is all-false → `apply_token_mask` errors → the whole request 503s mid-generation. Alternative outcome: the model pads whitespace to `finish_reason: length` with invalid JSON. Repro: `{"type":"object","additionalProperties":false,"properties":{"a":{"type":"integer"}},"required":["a"]}`, model emits `{"a":1` then `,"`. Fix: reject `,` (and `"` in Key mode) when no unused property remains — this can't lock the object out, since no-unused-props implies `required_satisfied`, so `}` is always available.

**B3 — Decode-time DoS: attacker-controlled schema dimensions multiply into the per-step full-vocab scan** (`src/grammar.rs:651`, `:626`). Property count, property-name length, and enum candidate count/size are uncapped (only bound = body size limit), and each multiplies the pre-existing ~50k–150k `accepts()` calls per decode step: `used: Vec<bool>` is deep-cloned per candidate token (10k-prop schema ≈ ~1.5 GB memcpy + 150k allocs *per step*), and the key feasibility scan is O(props × keylen) per byte per candidate. This runs on the single engine worker, starving every other request. Fix: compile-time caps (e.g. ≤256 properties, ≤64-byte names, ≤256 enum members × ≤256 bytes) with typed 400s — consistent with the fail-closed philosophy.

### Majors

- **M1** (`src/grammar.rs:509`): `enum`/`const` with a contradictory sibling `type` compiles silently — `{"type":"integer","enum":["a"]}` forces the output `"a"`, violating the declared type (the schema is unsatisfiable per JSON Schema conjunctive semantics and should be a 400). Fix: when `type` accompanies enum/const, require it to be exactly `"string"`.
- **M2** (`src/api/mod.rs:7460–7496`): the gemma4 / runnable / DiffusionGemma serve lanes short-circuit BEFORE `constraint_from_response_format` runs and never read `response_format` → silent 200 with unconstrained output on those env-gated lanes. Fix: hoist the constraint parse above the lane dispatch and 400 from lanes that can't enforce it.
- **M3** (`src/api/mod.rs:7684`, parser at `:10577`): `tools` + `json_schema` whose schema declares a `name` string property → `parse_tool_calls` deterministically reclassifies the constrained output as a tool call (content emptied, fabricated ToolCall). OpenAI explicitly allows tools + response_format together. Fix: skip tool-call parsing when a constraint is active, or 400 the combination.
- **M4** (`src/api/mod.rs:7990`): `camelid_receipt` + constraint seals a `reproducible:true` receipt whose replay hardcodes `constraint: None` → an honestly generated receipt deterministically FAILS verification. Fix: 400 the combination (mirroring the stream/n>1 guards) or stamp + replay the constraint.
- **M5**: the PR body claims `docs/architecture/feature.md` (+293) in three places — the file is **not in the diff** (changedFiles=3). The authoritative supported-subset contract isn't delivered; please push it.
- **M6**: none of the three claimed invariants is enforced by a test; `stream:true`+constrained → 400 has zero coverage anywhere. The updated pinned vertical-slice test also *replaced* (rather than extended) the json_schema-without-payload → 400 case.

### Minors (fix-with, not gating)

lone UTF-16 surrogate `\uD800` escapes accepted in constrained strings (strict parsers, incl. serde_json, reject the output); non-ASCII enum members / property names can starve the mask via the lossy per-token byte cache (`decode(&[id], false)` → U+FFFD for byte-fallback tokens) — suggest rejecting non-ASCII enum/property bytes at compile time for now; `std::mem::replace(mode, Done)` makes every `Err` path fail *open* into "complete" (should fail sticky-closed — add a `Failed` variant); per-candidate clone is 3–5× heavier than `json_object` with no first-byte early-out; O(props × required) compile scan; stale doc comment on `ChatCompletionRequest.response_format` and the now-ambiguous "not supported yet" unsupported-fields message.

### Verdict

The `SchemaState` core is solid and worth landing, but B1 breaks the headline guarantee on every warm-cache retry, B2 makes ordinary requests 503 or emit invalid output, and B3 is a one-request DoS on the engine worker. With those three + M1–M4 fixed (all have small, local fixes sketched above), the missing design doc pushed, and tests pinning the three claimed invariants, this is a strong addition.

timtoole02:
Integration pass complete — the review findings are all closed on this branch and the PR is ready to merge once CI is green.

First: the `SchemaState` automaton core is genuinely good work, and the F-series responsiveness (F4 fixing the enum/type conflict, F5.2's error-taxonomy rationale) made this integration straightforward. The core ships as-is; every fix below is at the integration edges. @karan68's 8 commits remain the base of the branch; the hardening commits sit on top.

| Finding | Fix commit | Pinned by |
|---|---|---|
| **B1** prompt-prefix cache bypasses the constraint | `1b82d26f` — the cache never interacts with constrained decoding (gate inside lookup/store, both directions) | `constrained_generation_skips_prompt_prefix_cache_lookup`, `constrained_generation_does_not_store_prompt_prefix_cache` (both verified to fail pre-fix) |
| **B2** reachable dead end (`,"` after last property) | `2cb29d23` — `,` and key-opening `"` gated on an unused declared property remaining | `exhausted_object_rejects_comma_and_key_quote`, `empty_object_schema_rejects_key_quote`, `partial_object_still_accepts_comma` |
| Fail-open `advance` (promoted from minor) | `2cb29d23` — sticky `SchemaMode::Failed`; a rejected byte can never surface as a clean "stop" | `advance_error_is_sticky_failed` |
| **B3** schema-dimension DoS | `da56edf8` — compile-time caps (256 props / 64B names / 256 enum members / 256B literals / depth 32), typed 400s naming keyword + cap; O(required×props) scan → HashSet | `schema_dimension_caps_reject_oversized_and_accept_at_cap` |
| **M1** enum + conflicting `type` | Already fixed on your head (F4, `9de41c2`) — no change needed; exact repro pinned in `5f79b4ed` | `const_or_enum_with_conflicting_type_is_rejected` (now includes `{"type":"integer","enum":["a"]}`) |
| **M2** env-gated lanes silently ignore the constraint | `1d83a114` — constraint parse hoisted above lane dispatch; gemma4/runnable/DiffusionGemma return a typed 400 for constrained requests | Pre-existing before-runtime 400 test passing post-hoist + inspection (lanes are env-gated, not constructible in CI) |
| **M3** tool-call cannibalization | `1d83a114` — `should_parse_tool_calls(tools_active, constraint_active)`; constrained output is content by definition, tools still render into the prompt | `constrained_output_is_never_reclassified_as_tool_call` |
| **M4** receipts replay unconstrained | `2caeb413` — `ReceiptRequest.response_format` (digest-stable when absent, digest-bound when present); replay re-compiles it. Digest-safety check passed → full-fix path, not the 400 fallback | `receipt_request_without_constraint_serializes_unchanged`, `receipt_stamp_records_response_format` |
| **M5** missing design doc | `5993cebf` — `docs/architecture/STRUCTURED_OUTPUTS.md` (SCREAMING_SNAKE per the sibling docs' convention, not the `feature.md` placeholder): subset contract, caps table, 9 invariants each naming its test, relaxations, error taxonomy | scrub-guard clean |
| **M6** untested invariants | `e6b4674d` + per-phase tests above | `stream_true_with_json_object_is_rejected`, `stream_true_with_json_schema_is_rejected`, reinstated `chat_completion_rejects_json_schema_without_schema_payload` alongside the out-of-subset case |
| Stale doc comment + error copy | `5f79b4ed` — `response_format` field doc reflects json_schema; unsupported-fields message reworded route-specifically | — |
| Lone surrogates | Decided, not drifted: `str_advance` is genuinely shared with `JsonState`, so a Schema-only rejection isn't cleanly separable; left as-is and documented as a known relaxation with a follow-up note (same relaxation as the shipped json_object path) | `STRUCTURED_OUTPUTS.md` §3 |

Also merged current `main` into the branch (conflict-free). Local gates across every phase: fmt / clippy `--all-targets --all-features -D warnings` / lib suite 752→762 passing / vertical slice 84→87 / all-targets compile / `git diff --check` / public-scrub — all green at every step.

PR #447, opened 2026-07-14
Models: streamline activation and add guarded local deletion

karan68:
## Summary

- turn curated supported-model acquisition into one `Download and start` workflow: download, inspect, authoritative fit/load, active-model confirmation, readiness confirmation, then Chat
- keep arbitrary Hugging Face rows download-only and runnable smoke admission separate from serving
- add Windows-first in-app deletion for inactive local GGUFs with exact file identity, same-origin management authorization, and lifecycle exclusion across load/smoke/download/generation
- keep the Models coordinator mounted across navigation, serialize disk scans, preserve immediate cancellation, and prevent competing acquisitions, settlements, and loads

## Model lifecycle safety

- load and unload are serialized transitions (a dedicated transition mutex); publication and teardown can no longer interleave
- every prepared generation holds an owned file lease through decode, so a model cannot be unloaded or deleted while a request is still using it
- deletion takes the exclusive lease and fails closed with `409 model_operation_in_progress` when any scan, smoke, load, or generation is active (never queues behind them)

## Guarded deletion contract

- deletion is unavailable while any model is loaded or any load, smoke check, unload, or model download is active
- the local scan issues a process-bound opaque token over filename, size, high-resolution write time, NTFS volume/file index, and link count
- Windows deletion reopens the exact file with `OPEN_REPARSE_POINT`, rejects directories/reparse points/hardlinks, revalidates identity, denies write/delete sharing, and applies deletion to that same handle
- files that are in use return `409 model_in_use`; stale/replaced files fail closed with `409 model_changed`
- linked/reparse-point GGUFs stay listed but receive no delete token (they are no longer hidden from the scan)
- requests require a loopback-bound server, a loopback `Host`, and an `Origin` authority that exactly matches `Host`; cross-origin scans receive no deletion token
- non-Windows deletion returns `501 model_delete_unsupported` and scans expose no deletion token

## Download lifecycle

- duplicate catalog id or destination filename returns a typed `409`; cancel-after-completion returns `409 download_already_completed`
- terminal (completed/failed) download records are retained and non-destructive on read, so a reloaded or second surface can rejoin the exact continuation (persisted download/smoke/start mode)
- an explicit acknowledgement endpoint clears a terminal record after the continuation succeeds, rejects still-running downloads, and is idempotent — so a remount cannot replay a completed continuation

## UI behavior

- loaded model: show `Unload`; hide delete action
- unloaded authorized model: show `Use for chat` plus a trash action; unauthorized/non-Windows clients see no delete control
- confirmation names the exact filename, friendly size, and exact byte count
- catalog acquisition ownership is lifted to the parent and pinned across live searches, so a stale search response can never unmount the active row or strand the lock
- automatic post-download checking/loading counts as a page-level model operation, so deletion stays blocked during it
- switching the API base clears and regenerates model state so stale snapshots from a previous backend cannot apply

## Validation

- Rust library suite: 786 passed, 0 failed, 55 ignored
- guarded deletion + lifecycle HTTP/authorization tests: 10 passed
- download install/cancel/ack/retention tests: 8 passed
- frontend catalog-activation, deletion, and UI smokes passed; production Vite build passed
- Windows all-target Clippy and Linux-target `--lib` Clippy both pass with warnings denied
- local model deletion is intentionally Windows-first; other platforms receive an explicit unsupported response and no authorization token
- no model support, parity, or backend capability claim is widened by this PR

model deleted
model delete sure dialogue box
model loading
model unloaded
model ready


karan68:
@timtoole02  please review this

PR #473, opened 2026-07-17
Web Workspace: durable read-only conversations and grounded local analysis

karan68:
## Summary

This PR adds **Web Workspace** as an additive, read-only preview for durable conversations grounded in one user-selected local folder. It does not replace Chat, the CLI/TUI agent, model management, or the OpenAI-compatible API.

Workspace provides:

- durable SQLite/FTS5 threads with idempotent follow-up messages;
- explicit resume, delete, compact, and undo operations;
- exact 4,096-token prompt-plus-generation budgeting;
- bounded evidence persistence with SHA-256 integrity checks;
- exactly three server-enforced tools: `read_file`, `list_dir`, and literal-content `search`;
- deterministic explicit file inventories without trusting model-retyped filenames;
- a responsive transcript, context inspector, follow-up composer, and bounded activity history.

The feature remains **preview / not production-promoted**.

## Review Closure

Commits `821ae395`, `511a65b7`, `d30d5e52`, `349e6e44`, `f2c78102`, `2ae0a484`, and `0577972d` close the findings from the review at `4c3c7ca`.

### Durable turn lifecycle

- `Cancelled` and `Failed` are recoverable turn outcomes; the durable thread can accept a follow-up.
- Every terminal outcome is persisted under schema v5 with an explicit outcome.
- Unsuccessful attempts remain visible and idempotent but do not enter future model context.
- An oversized newest turn stops context selection instead of silently falling back to stale history.
- Exact prompt fitting can clip tool observations while preserving call/result pairing.
- Unclaimed event streams expire after 30 seconds, persist the attempt as aborted, and stop blocking model transitions.
- Frontend unmount requests cancellation; late events from a closed stream cannot overwrite terminal UI state.
- Choosing a saved conversation immediately hydrates its bounded transcript preview on the right before resume; durable turn counts disclose older saved history.
- Existing file paths explicitly named by the user are sandbox-resolved and must be successfully read before a final answer is accepted. Broad requests remain step-bounded and must disclose uninspected scope; this does not claim exhaustive recursive coverage.
- Stop remains in a non-sendable state until the backend clears the owning turn; SSE-disconnect fallback persists ^Gborted and finalizes cancelling before follow-up is enabled.

### Grounding and inventory correctness

- Deterministic replacement is limited to explicit single-directory list requests.
- Semantic questions such as "What does .env configure?" remain model-owned.
- Case-distinct filenames remain distinct on case-sensitive filesystems.
- Multiple unqualified directory listings are never merged into a misleading flat inventory.
- Literal `%` is preserved; backticks/control characters remain safely encoded.

### Security and bounded I/O

- Workspace requires an exact same-authority browser `Origin`; Vite development uses a same-origin proxy.
- File reads and searches use one opened regular-file handle and bounded readers, rejecting special files and preventing path-check/read races.
- Directory listing and folder browsing retain deterministic sorted subsets when capped.
- Read-only tool advertisement is name-based, not positional.
- SQLite and filesystem work from async HTTP handlers runs through `spawn_blocking`.
- Whole-database foreign-key validation runs only after creation/migration; structural schema checks remain fail-closed.

A Workspace root may be any directory the local user explicitly selects. That is an intentional product boundary, protected by exact-origin loopback authorization and canonical per-session confinement, not an ambient filesystem grant.

### Additive behavior restored

The existing Full CLI/TUI agent keeps its merge-base behavior:

- Full search supports up to 100 hits; Workspace remains capped at 20 hits, 5,000 files, and two seconds.
- `--allow-fs` search works outside the default root.
- Native Qwen/Qwen35/Ornith tool-history replay is enabled only by the Workspace bridge.
- Full-profile Ctrl-C handling and bounded approval summaries retain their prior behavior.
- Lenient Windows-path repair touches only path-shaped fields, preserving valid escapes elsewhere.

### Contract and evidence discipline

- `COMPATIBILITY.md`, `STATUS.md`, and architecture docs describe the actual three-tool read-only surface.
- The 2026-07-17 write-capable Qwen bundle is labeled as a historical prototype receipt, not proof of current code.
- Write-era generators, orphan model-run scripts, dead approval UI, duplicate aliases, and unreferenced PNGs were removed.
- Workspace logic and model-free desktop/mobile visual smokes now run in CI.

## Architecture

- `src/chat/workspace_memory.rs`: schema-v5 threads, terminal outcomes, FTS5 retrieval, evidence integrity, and reversible compaction.
- `src/api/workspace.rs`: exact-origin HTTP boundary, turn coordinator, claim deadline, blocking-I/O isolation, and thread APIs.
- `src/chat/agent.rs`: exact context fitting, Workspace-only native history, evidence checks, and narrowed inventories.
- `src/chat/tools.rs`: exact read-only profile, bounded reads/list/search, deterministic caps, and restored Full behavior.
- `src/chat/workspace_bridge.rs`: read-only runtime policy, bounded events, cancellation, and a model-step deadline shared by prompt fitting and generation.
- Frontend state keeps durable turns separate from a capped 240-event activity ring and renders at most 100 turns while older history remains saved.

## Validation

Validated across review closure `821ae395`, fresh-checkout visual fix `9730346c`, cross-platform read-only-profile test fix `9a5acd91`, and final review remediation `511a65b7`:

- `cargo fmt --all -- --check`;
- `cargo check --tests`;
- `cargo clippy --all-targets --all-features -- -D warnings`;
- focused Rust tests across Workspace state, memory, prompt budgeting, inventories, additive Qwen history, tools, parsing, bounded special-file I/O, and cancellable preflight deadlines;
- `cargo test --test workspace_api`: 8/8;
- frontend Workspace reducer/API smoke, including a 300-event bounded-state regression;
- frontend production build;
- deterministic browser smoke at 1280x800 and 390x844;
- no write UI and `allow_writes: false` request assertion;
- no overlap/overflow, compaction plus Undo, failed Stop remaining retryable, early SSE completion staying non-sendable, and follow-up appearing only after authoritative cancellation settlement;
- saved-thread selection hydrating prior questions, answers, and terminal outcomes before Resume & send, verified with both mocked browser acceptance and the live 11-turn thread;
- strict evidence privacy audit: zero findings;
- public evidence claims, ledger schema, ledger drift, and public scrub;
- historical bundle checksum: 8/8.

## Explicit Non-Claims

This PR does not claim:

- production-ready interactive latency;
- a current exact-model receipt for the final read-only implementation;
- context support beyond the fixed 4,096-token Workspace envelope;
- semantic/embedding retrieval or reusable GPU KV sessions;
- per-request proof of no CPU fallback;
- neighboring-model, cross-platform performance, or throughput promotion;
- recursive inventory unless explicitly requested and observed;
- writes, shell, network, GUI control, subagents, or unattended execution in Web Workspace.

The final exact-Qwen rerun remains blocked by local enterprise Code Integrity policy and is deliberately not represented as completed.

## Review Order

1. `src/chat/workspace_memory.rs`
2. `src/api/workspace.rs`
3. `src/chat/agent.rs` and `src/chat/tools.rs`
4. `src/chat/workspace_bridge.rs`
5. `frontend/src/lib/workspaceAgent.js` and `frontend/src/views/WorkspaceView.jsx`
6. focused tests and CI wiring


karan68:
Hey @timtoole02 , this is a big pr , can you please go through it and let me know if there is any thing to discuss on it

timtoole02:
> Hey @timtoole02 , this is a big pr , can you please go through it and let me know if there is any thing to discuss on it

Taking a look.

timtoole02:
# Review — PR #473: Web Workspace

Reviewed at `4c3c7ca` against merge base `b9f9a40`. Verified locally on macOS: `cargo check --all-targets`, `cargo clippy --all-targets`, `cargo fmt --check` all clean; `cargo test --test workspace_api` 8/8; `cargo test --lib workspace` 40/40.

## Overview

Adds a durable, read-only, folder-scoped conversation surface to the loopback Web UI: 12 new HTTP routes under `/api/agent/workspace`, a SQLite/FTS5 memory store, a `WorkspaceReadOnly` tool profile over `read_file`/`list_dir`/`search`, exact prompt budgeting with reversible compaction, and an 862-line React view. 63 files, +11,740/-112.

The engineering here is careful in the places that usually go wrong, and the design doc (`WEB_AGENT_WORKSPACE_PLAN.md`) is unusually honest — it even records that the final model-backed rerun was blocked and must not be claimed. The problems cluster in three areas: the durable-conversation state machine has several dead ends, the "purely additive" claim does not hold, and the documentation/evidence layer still describes the write-capable prototype this PR replaced.

## Verified sound (worth stating, since much of it was adversarially attacked and held)

- **The read-only claim is real and enforced at the executor.** `tools.rs:1024` `validate_for` rejects any tool outside the profile before arg parsing or I/O, called from `agent.rs:379` for every model-emitted call. A model emitting `write_file` or `run_shell` that was never advertised is rejected. The HTTP lane pins `ToolProfile::WorkspaceReadOnly` with `allow_net:false`, `allow_fs:false`, `ShellSandbox::Disabled` (`workspace_bridge.rs:449-471`), and `allow_writes` is inert, not merely 400'd.
- **Path confinement is correct.** `tools.rs:330-343` canonicalizes *after* symlink resolution, then does a component-wise root prefix check. `search` re-verifies at root, each popped dir, and each entry, with a `visited` set so symlink loops terminate.
- **SQL layer is clean.** Fully parameterized; the only `format!`ed statement takes a hardcoded table name. `fts_query` reduces input to `[alphanumeric_]` runs before quoting, so FTS5 operators cannot reach the parser. `TransactionBehavior::Immediate` on every multi-statement mutation. `busy_timeout` set before `journal_mode=WAL` — the ordering most code gets backwards.
- **No XSS.** No `dangerouslySetInnerHTML` anywhere; markdown goes through a scheme-allowlisted renderer; file contents render as text nodes. The PR *adds* hostile-input assertions to `markdown-smoke.mjs`.
- **Prompt injection is not newly introduced.** A dedicated adversarial pass produced zero surviving findings: the tokenizer/chat-template path that would let scanned file content forge a turn is byte-identical at the merge base, and the pre-existing CLI lane exposed the same channel to `write_file`/`run_shell`. The hardening gap is real but belongs to unchanged code.
- **Shared `Modal`/`SidebarRail` edits are additive** and regress no other view.
- **The 40 in-crate unit tests are substantive**, not smoke: schema-v1 migration, newer-schema fail-closed, compaction reversibility with FTS preservation, evidence-tampering fail-closed, transactional idempotency, approval-timeout and cancellation races.

---

## Blocking — durable conversation state machine

### 1. Cancel or error is session-fatal; the UI keeps offering a composer that can only 409

`install_turn` (`src/api/workspace.rs:176`) refuses any new turn unless state is exactly `Idle`. A Stop sets `Cancelled` and a driver error sets `Failed`, and nothing assigns `Idle` except `finish_turn_if_current`, which requires an owning in-flight turn that has already been cleared.

Meanwhile `stop()` (`WorkspaceView.jsx:460`) deliberately keeps `session` set — only `reset()` clears it — and the follow-up composer renders whenever `!running` (`WorkspaceView.jsx:627`). So: user clicks Stop, the composer reappears, they type their next question, and get `409 workspace_turn_already_active — "a turn is already active"` forever. Manual Compact/Undo are equally dead (`workspace.rs:586` requires `Idle`). Only Reset — a brand-new session — recovers.

This contradicts the plan doc's "cancellation is turn-scoped rather than process-global". It is session-fatal.

### 2. A turn is persisted only on `ModelAnswer`

`append_turn_with_evidence` is called inside `if let WorkspaceEvent::ModelAnswer` (`workspace.rs:1163`). `Aborted`, `StepCapped`, `Repeated`, `DriverError` and the 90s deadline all emit `Finished` without writing anything. The user's question vanishes from the "durable conversation", and duplicate-suppression finds nothing either, so a retry with the same `client_message_id` is treated as new.

Combined with #1: a timed-out turn loses the message *and* bricks the session.

### 3. `bound_context` drops the newest turn instead of stopping

```rust
// src/chat/workspace_memory.rs:775
for turn in recent.into_iter().rev() {          // newest first
    if used.saturating_add(size) <= max_chars {
        used += size;
        kept_recent.push(turn);
    }                                            // no else, no break
}
```

An oversized newest turn is skipped and older, smaller turns are kept. Production budget is `2 * 1024` **bytes** for all recent + retrieved turns combined (`workspace.rs:999`, `workspace.rs:1432`); a 512-token answer is already ~2000 chars and the max generation reserve is 1024 tokens. So on any substantial answer, "explain that further" is answered against turn N-1, silently. This is the feature's core promise failing quietly, and it is the first thing I would fix.

### 4. The trim loop cannot shed tool output, so a normal multi-read step hard-fails the turn

`remove_oldest_optional_context` (`agent.rs:890`) can only drop `Memory` blocks or an adjacent `User`+`Assistant` pair preceding the current user message. In the Workspace lane an `Assistant` message is pushed exactly once, immediately before returning `Answered` — so no `User`/`Assistant` pair ever exists mid-loop, and prior turns arrive only as `Memory`. The verbatim last tool group is untrimmable by construction (`compile_history_for_step` keeps it whole, `agent.rs:767`).

With `MAX_WORKSPACE_TOOL_CALLS_PER_STEP = 8` and a 2 KiB per-observation clip, one step can legally carry 16 KiB of tool text against an available `4096 - 512 = 3584`. The fitter deletes the ≤1 KiB memory blocks, finds no pair, returns `Err`, and the turn dies with `context budget error: …`. Nothing is persisted (#2) and the session is now `Failed` (#1).

Concretely: "summarise these files" → 6 `read_file` calls ≥2 KiB each → turn dies. The unit test that covers the fitter (`agent.rs:1779`) uses a driver scoring `ToolCalls | ToolResult => 0`, i.e. it explicitly assumes tool output is free.

*(No infinite loop and no panic here — the error exit is reachable and correct. The problem is that the exit is reachable on ordinary input.)*

### 5. The "deterministic file inventory" hijacks ordinary content questions

`agent.rs:314` discards the model's answer and substitutes a generated list. The trigger (`agent.rs:681`) is substring matching on the raw prompt: any of `list|show|which|what|find all|check all|all the`, minus a small content-verb blocklist. The extension detector (`agent.rs:694`) accepts any whitespace token shaped like `.ext`. Result:

- "**What** does the `.env` file configure?" → `Found 1 .env files in the selected folder: - .env`
- "**Which** `.rs` file implements the parser?" → flat dump of every `.rs` filename
- "**What** is the `.git` directory for?" → `No .git files were found in the selected folder.` — confidently wrong

Note this interception is the one sibling check *not* gated on `cfg.tool_profile.is_workspace()` (compare `agent.rs:318`); it is inert for the CLI agent only because the observations vector is populated behind a workspace check elsewhere. One edit to that push and it hijacks `camelid chat --agent`.

### 6. Case-folded dedup silently drops distinct files, and the test asserts the loss

```rust
// src/chat/agent.rs:618
let lower = entry.to_ascii_lowercase();
files.entry(lower).or_insert_with(|| entry.to_string());
```

On Linux (which CI runs) or case-sensitive APFS, a folder holding `README.md` and `readme.md` renders one bullet and "Found 1 Markdown file". `canonical_inventory_filters_sorts_and_deduplicates_case_insensitively` (`agent.rs:2120`) asserts this as intended. Because the output *overrides* the model, the user has no route to the correct answer. Invisible on NTFS — the platform the evidence bundle was captured on.

The same map merges every `list_dir` observation in the turn without directory qualification (`agent.rs:597`), so `docs/README.md` appears as a bare `README.md`, collides with the root one, and the "Nested folders were not searched" trailer is false.

---

## Blocking — contract discipline

### 7. COMPATIBILITY.md, STATUS.md and WORKSPACE_MEMORY_SPEC.md still describe the five-tool write surface

COMPATIBILITY.md declares itself the release contract — *"If another document or UI sounds broader, this file wins."* Its new rule says the tool set is *"exactly `read_file`, `list_dir`, `search`, `write_file`, and `edit_file`"* and that *"writes require approval showing the exact resolved target and complete proposed content."* The code refuses all of that.

- `COMPATIBILITY.md:36`
- `STATUS.md:5` — "complete on the bounded five-file-tool surface"
- `docs/architecture/WORKSPACE_MEMORY_SPEC.md:126` — same five tools, "writes/edits are approval-gated"

The ledger entry and `WEB_AGENT_WORKSPACE_PLAN.md` were updated correctly in the final commit (`4c3c7ca`) and are honest; these three were missed. `check-ledger-drift.mjs` compares ledger↔code and cannot see a contradiction inside a prose capability rule, so CI is green on a false contract. As merged, the repo's authoritative file advertises approval-gated local file writes over the loopback UI that cannot happen.

Also minor: `WEB_AGENT_WORKSPACE_PLAN.md` is stamped `Last verified: 2026-07-22` — tomorrow.

### 8. The evidence bundle certifies the pre-pivot implementation

`qa/evidence-bundles/workspace-qwen3-4b-q4km-…-head-8c2a2b74/manifest.json` pins `implementation_sha256: 910e33de…`. Recomputed with the exact algorithm from `scripts/assemble-workspace-evidence.mjs:64` over the 16 files the manifest names, at this PR's HEAD:

```
recomputed : b36b383c9c5a0a67551d95cb6e883b7d1fa68f1f92fdec7cb561024c7c206151
manifest   : 910e33de5ccfa377be06598fe6ba88959e574b661e2db6eb099d25cc3a4310ba
```

The manifest self-describes as a dirty-tree snapshot at `8c2a2b74`, so this is not a forged claim — but COMPATIBILITY.md cites it as proof of the *current* rule. `real-ui-approval.png` shows a "Review file change / Allow once / Always allow write_file / Deny" modal and a footer reading "every write asks first" — UI this PR deletes. `SHA256SUMS` verifies 8/8, so the bundle is internally consistent; it just certifies code that no longer exists.

Two of the nine new scripts can never pass against the shipped code:
- `scripts/workspace-stage5-eval.mjs:170` hard-fails unless it observes `approval.required` for `write_file`
- `frontend/scripts/workspace-stage5-ui.mjs:36` waits on `.workspace-approval`, which exists only in dead CSS (`workspace.css:217-254`)

…and these are the two scripts the manifest names as its evidence generators.

**None of the nine new `.mjs` scripts run in CI.** The `frontend` job names seven `smoke:*` steps; `smoke:workspace` and `smoke:workspace-visual` are not among them. `validation-scripts` globs `scripts/test-*.mjs`, matching none. Four (`workspace-memory-soak|-cancel|-sustained|-inventory-e2e`, ~730 lines) are referenced by no workflow, doc, or package script at all.

The highest-value single change here: `frontend/scripts/workspace-readonly-visual-smoke.mjs` is the strongest artifact in the PR — it stubs the API, drives the built UI in two viewports, and asserts falsifiable properties (no write UI leaked, `allow_writes === false` on the real request body, no overlap/overflow, cancel-failure not misreported). It needs Chrome and `npm run preview` but **no model**, so it is CI-runnable on `ubuntu-latest` today. Wire it in.

---

## Security

### 9. The Workspace origin guard accepts any loopback origin

`local_management_request_allowed` (`src/api/workspace.rs:1561`) checks only that the `Origin`'s **host** is loopback. The guard of the same name it was cloned from (`src/api/mod.rs:18343`, which protects model-delete) requires `origin.authority() != Some(&authority)` — exact host **and port**.

This is deliberate and documented (plan doc invariant 3 says "same-origin fetch metadata **or an allowed loopback Origin**"), and it has a real motivation: `useDashboardData.js:33` returns `http://127.0.0.1:8181` as the API base under Vite DEV, so strict authority equality would break the dev workflow. The test at `workspace.rs:1726` encodes the intent:

```rust
headers.insert("host", "127.0.0.1:8181".parse().unwrap());
headers.insert("origin", "http://localhost:4173".parse().unwrap());
assert!(local_management_request_allowed(&headers));
```

Still worth changing. All 12 routes sit under `CorsLayer::permissive()` (`mod.rs:1924`), so any page on any other localhost port can both drive *and read* them: `GET /browse` to enumerate the filesystem → `POST /sessions` rooted at `$HOME` → read file bytes off `tool.result` SSE events. Public websites are correctly blocked (non-loopback `Origin` fails), so this needs an attacker-influenced local origin — another project's dev server, a local app with a web UI, XSS in one. Conditional, but the dev-convenience reason is better served by a dev-only `cfg` than a permanent widening of a guard whose stricter twin already exists in this file's sibling.

### 10. Nothing bounds which directory may be a workspace root

`create_session` accepts any canonical directory (`workspace.rs:925`); `browse` defaults to `/` on Unix and lists any directory on the host, unconfined (`workspace.rs:745`). Read-only is enforced; *read-only over what* is not. Root = `$HOME` makes `read_file(".ssh/id_rsa")` an in-sandbox read. Egress is genuinely blocked (no `http_fetch`, shell disabled), so this is disclosure-to-the-viewer — which is exactly why #9 matters.

### 11. Unbounded whole-file read

`tools.rs:1395` `std::fs::read(path)` for the no-range branch; `MAX_RANGED_FILE_BYTES` (8 MiB) is checked only when `start_line`/`max_lines` are supplied, and the 64 KiB truncation happens *after* the whole file is in memory. A multi-GiB file in the root is a memory spike; `/proc/kcore` is an abort. Pre-existing in the CLI agent, newly reachable from an HTTP handler.

---

## "Purely additive" does not hold

`src/chat/tools.rs` and `src/chat/agent.rs` change `ToolProfile::Full` — the existing `camelid chat --agent` — in ways not gated on the new profile:

1. **`search` capped at 20 hits.** `MAX_SEARCH_HITS = 100` existed at the merge base and is gone; the schema now declares `"maximum": 20` (`tools.rs:428`), with new `MAX_SEARCH_FILES = 5_000` and a 2-second wall-clock cutoff (`tools.rs:257`). An existing run that got 100 hits now gets 20; on a cold cache a large repo hits the cutoff and returns a near-empty result the model reads as "no matches".
2. **`search` broken under `--allow-fs`.** `tools.rs:1468` re-checks the root with no `fs_unrestricted` exemption, unlike `Sandbox::resolve` (`tools.rs:343`) and unlike `read_file`/`list_dir`. The system prompt tells the model "you may read and write files anywhere on this computer"; `search` on an absolute path now always errors.
3. **Qwen3 tool history re-rendered.** `history_to_messages` gained a `family` param (`agent.rs:1276`); for `qwen3`, prior tool calls become `<tool_call>` assistant text and results become `role:"user"` `<tool_response>` blocks instead of `role:"tool"`. `LiveDriver` is shared with `agent_tui`, `agent_eval`, `agent_orchestration`, `subagent` — so every existing Qwen3 CLI agent run now sends a different prompt shape than the committed `camelid.agent_eval/v1` PASS receipts were produced against. Probably a fix; needs CLI-side re-validation, which isn't in the diff. It also doesn't cover `qwen35`/Ornith: rendering gates on exact `"qwen3"` (`agent.rs:1276`) while parsing gates on `contains("ornith") || contains("qwen35")` (`tool_parse.rs:17`), and Ornith rows are `tool_capable: true`.
4. **Ctrl-C now discards the partial answer.** A new post-step cancel check (`agent.rs:296`) — absent at the merge base — turns what used to render as `Answered` into `Aborted`.
5. **Write approvals dump full file content into the prompt.** `approval_detail` gained `"\n--- proposed content ---\n{content}"` (`tools.rs:886`); a 200 KB generated file floods the terminal/TUI approval before the y/n.
6. **`escape_lone_backslashes` corrupts valid escapes.** `tool_parse.rs:97` doubles every backslash except `\\` and `\"` — including `\n`, `\t`, `\uXXXX` — and applies to the whole arguments blob. A `write_file` call with an unescaped Windows path *and* a legitimate `\n` in `content` parses on retry with the newline turned into a literal backslash-n. Silent file corruption, no error. (The doc comment states this is intentional for paths; the bug is that it leaks across fields.)

---

## Robustness and performance

- **`WorkspaceMemoryStore::open()` runs a whole-database `PRAGMA foreign_key_check` behind a process-global mutex, per HTTP request** (`workspace_memory.rs:71-83` → `:747`). With no violations, SQLite must scan every FK edge to conclude there are none. `open()` is called on every `/threads`, `/sessions`, `/messages`. Cost grows with total history; all workspace requests serialize against each other. `context_for` additionally opens four connections per call.
- **No `spawn_blocking` anywhere.** Every rusqlite call and the `browse` `read_dir` run on tokio worker threads. (A fair counter-argument: WAL means readers don't block on the writer, so the 5s `busy_timeout` stall scenario mostly doesn't apply. The per-request FK scan is the real cost.)
- **`list_dir` caps before sorting** (`tools.rs:1417` vs `:1428`). `read_dir` order is filesystem-defined; NTFS returns roughly collation order, APFS/ext4 return hash order. So off Windows the retained 4096 are an arbitrary subset and paging by `offset` returns an arbitrary page. Commit `23df630` is the fingerprint: it replaced `assert!(output.text().contains("entry-4095.md"))` with a shape-only check rather than sorting before capping. Sort first.
- **An unclaimed session blocks model load/unload server-wide.** `WaitingForEvents` returns true from `blocks_model_transition()` (`workspace.rs:104`), and this PR gates both `load_model_from_path_with_activation` and `unload_model` on it. A *claimed* stream cancels correctly on disconnect via `CancelStreamOnDrop` (`workspace.rs:335`), but if the stream is never claimed there is no guard and no reaper — the only `Duration` in the file is the 10s SSE keep-alive. The frontend's unmount effect (`WorkspaceView.jsx:379`) closes the EventSource without cancelling. Recovery exists but only by hand: the session id equals the memory thread id (`workspace.rs:1012`), so `GET /threads` exposes it and `DELETE /sessions/:id` clears it — no UI affordance does. Add a claim deadline and cancel on unmount.
- **Unbounded transcript.** `state.events` is append-only for the life of the view and line 854 renders every event; the sibling `ChatWorkspace.jsx:564` caps at `slice(-60)`. `reduceWorkspaceEvent` copies the whole array on every `model.delta`, so a streaming turn is O(n²).
- **`turn_index` reported to clients resets to 0 after compaction** (`workspace.rs:1443` derives it from the compaction-filtered context rather than the store). The DB stays correct; the SSE event and POST response do not.
- **`output_tokens` is an SSE delta count**, not tokens (`agent.rs:1224`), surfaced verbatim as `model.timing.output_tokens`.
- **`resident_cuda` is always `null` on macOS/Metal** (`inference.rs:10934`), so the inspector's residency row is dead on this platform. No new engine-state mutation on any cancellation path — the known CUDA-gated-reset-without-Metal-twin class of bug is *not* reintroduced.
- `.expect()` on thread spawn in an HTTP handler (`workspace.rs:1134`, `:1243`) — matches pre-existing repo convention (`engine.rs:94`), so noting rather than pressing.

## Smaller cleanups

- Dead approval surface across three layers: `decideWorkspaceApproval()` exported and imported by nothing; reducer still tracks `pendingApproval` and renders "Waiting for your approval — review the pending change to continue" with no approve/deny control in 862 lines; `overlayClassName` added to the shared `Modal` for zero call sites; 11 orphan `.workspace-approval*` CSS rules.
- `className="workspace-folder-modal"` (`WorkspaceView.jsx:169`) has no matching CSS rule.
- `FolderPicker` fetch passes no `AbortSignal` (unlike every other fetch in the view); `answerCopied` `setTimeout` never cleared on unmount; `aria-valuemin/max` call `getBoundingClientRect()` **during render**.
- A failed `cancelWorkspaceSession` skips the `close()` block, leaving the EventSource open (`WorkspaceView.jsx:460`).
- `tools.rs:444` `tools.truncate(3)` trims the read-only advertisement positionally — safe today because `ToolProfile::allows` is the real gate, but fragile.
- `simplify_path` (`workspace.rs:1005`) duplicates `tools::display_path` byte-for-byte; `try_decide` is duplicated across `WorkspaceBridgeClient`/`WorkspaceBridgeControl`; `render_relevant_memory`/`render_recent_memory` differ by one string.
- `agent.rs:615` `entry.contains(['\r','\n'])` — the `\n` arm is unreachable inside `.lines()`.
- `agent.rs:570` percent-encodes `%`, so `100%-done.md` renders as `100%25-done.md`.
- `src/lib.rs:6` `pub mod chat;` moves the whole TUI stack (`ratatui`, `rustyline`, agent, subagent) into the library, so every consumer of the crate now compiles it.
- Two npm aliases for the same script (`smoke:workspace-visual`, `smoke:workspace-readonly-visual`).
- Three orphan doc PNGs (341 KB) referenced by no tracked file.
- Evidence PNGs legibly display `C:\camelid-fork\target\workspace-stage5-…` — the exact strings `scrub()` redacts in the JSON siblings. `check-public-scrub.sh` uses `git grep -I` (skips binaries) and `audit-evidence-bundle-privacy.mjs` only scans text extensions, so images bypass the redaction policy. No username or home directory exposed; hygiene hole, not PII.
- The 4096-token envelope is a hardcoded constant (`workspace_bridge.rs:27`) with no reference to the model's context length. The only admitted row, `qwen3_4b_q4_k_m`, declares `bounded_context_4096_pack: "not_promoted"` — so routine turns drive ~3.5k-token prompts on a row whose committed parity evidence stops at 1024.
- The 90s deadline is real on both headers and body, but it is *total elapsed* (a healthy long generation is killed mid-answer, losing the turn per #2) and it does not cover `generation_preflight`'s fixed 30s timeout, which checks no cancel flag and can be issued repeatedly inside the trim loop.

## Suggested sequencing

1. Fix the state machine: `Cancelled`/`Failed` → recoverable; persist turns on every terminal outcome; `bound_context` stops instead of skipping; make tool output trimmable.
2. Reconcile COMPATIBILITY.md / STATUS.md / WORKSPACE_MEMORY_SPEC.md with the ledger note this PR already got right; either rename the evidence bundle as historical in the contract, or capture a read-only one.
3. Either delete the inventory override or narrow it to explicit list requests and drop the case-folding — as written it answers questions it was not asked.
4. Restore authority equality on the origin guard behind a dev-only exception; add a `read_file` size cap before the read.
5. Un-regress `ToolProfile::Full`: restore the 100-hit search cap for non-workspace profiles, exempt `fs_unrestricted` in `search`, scope the backslash repair to path-shaped fields.
6. Wire `workspace-readonly-visual-smoke.mjs` and `smoke:workspace` into CI; delete or quarantine the write-era scripts.
7. Sort before capping in `list_dir`; add a session claim deadline; cancel on unmount.

karan68:
@timtoole02  updated the pr with your comments , let me know if it looks ok

timtoole02:
Will do!! Thanks!!

PR #487, opened 2026-07-21
fix(resident): stop a CPU fallback from silently zeroing a GPU KV history

timtoole02:
## The bug

The GPU-resident lanes advance `kv_cache.position` while writing K/V **only** into the GPU cache: `try_metal_resident_prefill` sets `position = n` outright, and every resident decode step (Metal and CUDA alike) appends on the GPU and ends with `set_filled(position + 1)`. The CPU buffers stay length 0. Nothing kept a sequence on that lane, and one step off it was enough to lose the prompt:

1. A resident step declines mid-sequence and the caller falls through to the CPU layer loop.
2. That loop attends over `kv_cache` positions `[0, position]` — all zeros for every position the GPU produced. Its `write_kv_cache` then grows the buffers via `ensure_position_capacity` and writes real K/V for the **current** position only, leaving `[0, position)` zero-filled.
3. The next resident step sees `filled != position` and reseeds. The guard there was a **bounds** probe, so it passed — the bytes were addressable, they were simply zero — and the seed copied that zeroed prompt onto the GPU.
4. Every later token attended over a zeroed prompt. Wrong output, no error.

Step 2 is the part that shaped the fix: **the fallback step is already wrong on its own, before any reseed.** Gating the reseed alone does not help, because declining the resident step just routes to a CPU step that is equally blind.

## The fix

The cache now carries a **materialized-through watermark** — the highest position any CPU forward, prefill, or GPU→host mirror actually *wrote*, as opposed to merely allocated — and the recovery mirrors the resident engine's KV back into the CPU cache before any CPU forward reads it.

CUDA has always done exactly this after its prefill (`copy_resident_cuda_kv_to_host`); it simply never did it per decoded token, and Metal had no readback at all. That asymmetry was the bug.

- `LlamaKvCache::materialized_through`, advanced in `store_kv_head_row` (the canonical store every writer routes through, so the mirror needs no special-casing) and **clamped down** in `rollback_to_position` — without that clamp it would vouch for rows a rejected speculative branch wrote, which are stale rather than zero and harder to spot than zeros.
- `f32_history_materialized` now guards the Metal reseed; `history_materialized` (dtype-neutral) guards the CUDA reseed, which previously had no validity check at all. `cpu_kv_authoritative` moves off `allocated_sequence_length` for the same reason.
- `metal::ResidentDecodeState::read_kv_layer`, the exact inverse of `seed_layer` and the twin of CUDA's existing readback. Metal buffers are shared storage, so recovery is a strided memcpy over unified memory.
- `ensure_cpu_kv_materialized` asks both lanes, in the order they can be trusted. The Metal engine hangs off the session, so it is unambiguously this sequence's. The CUDA engine is a process-global keyed by model identity, so `recover_cpu_kv_from_cuda_resident` must establish identity first: matching key, matching **absolute** layer range (two pipeline shards of one model collide on the key alone), weights still uploaded, and `filled == position` exactly — the same readiness triple `verify_drafts_gpu` already requires before letting the global engine speak for a sequence. `ResidentCudaSlot` records its range for that check.
- Recovery never returns `Err` on the CPU forward path, and a partial recovery restores the watermark rather than vouching for layers it never reached. When neither lane can supply the gap the history is genuinely gone; it warns instead of pretending.

No kernel math, reduction order, or default-on path is changed.

## Verification

On M4 / `Llama-3.2-1B-Instruct-Q8_0`:

- **`tests/resident_kv_cpu_fallback.rs` passes, and FAILS with only the two recovery call sites disabled** — `" capital of Spain is Madrid"` becomes `" best man is the man"`, derailing exactly one token after the forced fallback. Its binding assertion is resident-with-fallback == resident-without-fallback, so the forced fallback is the only variable. The pure-CPU run is compared only when it agrees with the clean resident run, since the CPU cache rounds K/V through f16 while the resident cache keeps f32 and the two may part company at depth for unrelated reasons.
- **`qa/speed/spec-verify-parity.sh` PASS** — lossless on every gate column, both lanes, `gpu_verify_rounds > 0` throughout (receipt included).
- **`tests/spec_draft_rollback.rs` passes**, `cpu_steps=0` on all three rounds.
- `cargo fmt --check`, `clippy --all-targets -D warnings`, and the full `--all-targets` suite.

## Caveats for review

- **The CUDA half is unverified.** `cudarc` does not build on macOS, so the `#[cfg(feature = "cuda")]` code here has not been compiled, let alone run. The regression test is backend-agnostic by design — it drives `generate_next_token_greedy_resident`, which dispatches to whichever resident lane the host has — so **running it on a CUDA box covers that half**, and that is the main thing this PR needs from review.
- One **pre-existing, unrelated** test failure is still red on this branch and reproduces identically at `main`: `tensor::nvfp4_tests::encode_vectors_reproduce_pin_wire_bytes_and_dequant`.


timtoole02:
## Verifying the CUDA half on Windows

Windows builds CUDA by default (`build.rs` sets the `cuda` cfg and the Windows `cudarc` is non-optional), so no `--features` flag is needed.

**1. Compile it.** This is the highest-value step: the `#[cfg(feature = "cuda")]` code in this PR has never been compiled on any machine.

```
git fetch origin
git checkout fix/resident-kv-cpu-fallback-history
cargo clippy --all-targets -- -D warnings
```

**2. Run the regression test.** It needs a model:

```
set CAMELID_RESIDENT_KV_FALLBACK_GGUF=C:\path\to\Llama-3.2-1B-Instruct-Q8_0.gguf
cargo test --release --test resident_kv_cpu_fallback -- --nocapture
```

It prints three token streams and asserts that resident-with-one-forced-fallback equals resident-with-no-fallback. Two failure modes worth telling apart:

- `resident lane declined unexpectedly at generated token N` — the CUDA lane never engaged, so the run proves nothing either way. Check that GPU acceleration is on and the model is eligible.
- The three streams diverge — a real defect in the CUDA recovery path.

The test sets `CAMELID_METAL_RESIDENT_DECODE=1`, which is inert here; CUDA resident decode is default-on wherever a device is present, and the dispatcher prefers it.

**3. Full suite.**

```
cargo test --release --all-targets --no-fail-fast
```

Expect exactly one red: `tensor::nvfp4_tests::encode_vectors_reproduce_pin_wire_bytes_and_dequant`. It is pre-existing and unrelated — it reproduces identically at clean `main`, and it is release-only (debug passes 8/8, release fails on the signaling-NaN row `path-snan-first`). Whether it also fails on Windows is useful signal in itself: that would rule out an Apple-Silicon-specific cause.

timtoole02:
## CUDA half verified on Windows — all three steps green

Ran the plan above on **Windows 11 / RTX 3060 Laptop GPU (compute 8.6, 6 GB)**, CUDA default-on build (`build.rs` sets the `cuda` cfg; Windows CI runners have no CUDA toolkit, so this really was the first compilation of the `#[cfg(feature = "cuda")]` code).

**1. Compile — clean.** `cargo clippy --all-targets -- -D warnings` → exit 0, no warnings. First-ever compile of `recover_cpu_kv_from_cuda_resident`, `copy_resident_cuda_kv_to_host`, `CudaResidentDecode::{n_layers, read_kv_layer}`, and `ResidentCudaSlot`.

**2. Regression test — the CUDA lane engaged and recovery is byte-exact.**
```
cargo test --release --test resident_kv_cpu_fallback -- --nocapture
[cuda] selected device 0 of 1: "NVIDIA GeForce RTX 3060 Laptop GPU" (compute capability 8.6) | VRAM 5122 MiB free / 6143 MiB total
pure-CPU              : [22463, 13, 578, 6864, 315, 18157, 374, 25048]
[gpu] all 16 layers resident in VRAM (5122 MiB free) — full GPU speed
resident, no fallback : [22463, 13, 578, 6864, 315, 18157, 374, 25048]
resident, one fallback: [22463, 13, 578, 6864, 315, 18157, 374, 25048]
test cpu_fallback_mid_sequence_preserves_the_resident_kv_history ... ok
test result: ok. 1 passed; 0 failed; ... finished in 10.83s
```
The resident lane did **not** decline — it engaged (16 layers resident), a mid-sequence CPU fallback was forced, and the recovered history produced the identical stream. All three streams agree; because the CUDA resident KV is f16-on-device, the readback rounds idempotently through the CPU cache's f16 store, so even the pure-CPU leg matches (the f32/f16 caveat didn't bite on CUDA).

**3. Full suite — 0 failed, and the expected nvfp4 red is Apple-Silicon-specific.**
`cargo test --release --all-targets --no-fail-fast` → every binary reports `0 failed`. Notably `tensor::nvfp4_tests::encode_vectors_reproduce_pin_wire_bytes_and_dequant` **passes** on Windows/x86 (it is red at `main` only on M4). That confirms the M4 red is the ARM signaling-NaN `path-snan-first` behavior, not a real defect — no cross-platform failure remains.

### Independent data-path check
Confirmed `read_kv_layer` emits `[kv_head][position][head_dim]` with stride = `n_positions`, which matches `store_mirrored_layer_kv`'s `(h*position_count+p)*head_dim` indexing exactly; the identity gate (`key`, absolute layer range, `weights_ready()`, `n_layers()==range.len()`, `filled()==position`) all check out.

### One follow-up (not a blocker for this PR)
An adversarial review flagged that the fix guards **2 of 3** CPU KV-history readers. The third — `forward_greedy_verify_chunk` (speculative-decode CPU verify, `src/inference.rs:3761`) — has no `ensure_cpu_kv_materialized()` call, so under `CAMELID_SPEC_GPU=1` **and** a declined GPU verify (offloaded engine, or a multi-session same-model collision) it can attend over the same zero-filled prefix and then advance `materialized_through` across the gap. This is **pre-existing** (the spec loop and `resident_paths_disabled` gating live in `src/api/mod.rs`, untouched here — it reproduces at `main`), **config-gated** (not the default path), and **not a regression** — so it doesn't hold up this PR. Filing separately to add the same guard plus a reproducing test for the speculative-offloaded path.

Merging on the strength of the three green steps above.

PR #489, opened 2026-07-22
MESA: compile CUDA into the default x86_64 Linux build + `serve --gpu` switch

timtoole02:
## What & why

Bring **x86_64 Linux** up to the Windows plateau: CUDA compiled into the **default** build, the GPU used automatically when a device is present, and an explicit on/off switch — without touching the CPU parity reference, the Windows wiring, or the aarch64 (Pi / NanoCamelid) opt-in path.

The detect → default-on → UI-toggle → `/api/runtime/gpu` machinery already existed; it was just **dark on Linux** because the `cuda` feature wasn't compiled into the default Linux build (`is_available()` returned false with reason "feature not built"). This is the build-wiring gap. **No kernel, dispatch, quant, or parity-logic changes.**

## The change (4 files, additive)

1. **`Cargo.toml`** — `cudarc` is now **non-optional** on `cfg(all(target_os="linux", target_arch="x86_64"))`, mirroring the Windows dep. A second block keeps it `optional` for every other non-macOS/non-Windows target, so `cuda = ["dep:cudarc"]` still resolves and **aarch64 (Pi) + BSD stay CPU-only by default**.
2. **`build.rs`** — injects `cargo:rustc-cfg=feature="cuda"` for x86_64 Linux (same as the Windows block), with **no** Optimus/`/STACK` link args (those stay Windows-only).
3. **`src/cuda.rs`** — sibling `compile_error!` guard (G4, **included** per the campaign's default recommendation) so a bare x86_64-Linux build can never silently drop CUDA.
4. **`src/main.rs`** — `serve --gpu auto|on|off` (env `CAMELID_GPU`) seeds the runtime GPU switches at startup for headless/agent runs; also honored on the double-click launch path via `GpuMode::from_env()`.

### Precedence contract
`--gpu` / `CAMELID_GPU` defaults to **`auto`** → lazy-seeds exactly as today (`gpu_accel` from `is_available()`, hybrid Q8 from `CAMELID_CUDA_Q8`). `on`/`off` are **authoritative at startup** and override the env seed; the UI `POST /api/runtime/gpu` still flips state live afterwards. Deterministic mode and `CAMELID_CUDA_RESIDENT_DECODE=0` still force GPU off at their own call sites. *(Campaign decision point: flag-wins-over-env chosen. Say the word to invert so env wins for CI reproducibility.)*

## §4 consequence to note
After this, **every** x86_64-Linux build compiles cudarc (Windows-identical). CI ubuntu already built `--all-features`, so it compiled this path before; now a bare ubuntu `cargo build`/`test` also pulls cudarc. Runtime CPU-only stays fully available via `--gpu off` or a GPU-less host (cudarc no-ops via `fallback-dynamic-loading`); only a *compile-time* exclusion is no longer a default option (that would be a dedicated future flag, not MESA).

## Verification (on this Windows / RTX 3060 host)
| Gate | Result |
|------|--------|
| G1 | `cargo tree --target x86_64-unknown-linux-gnu -i cudarc` flips from absent → present; aarch64 stays clean; manifest resolves |
| G2 | `cargo check --target x86_64-unknown-linux-gnu --bin camelid` (no `--features`) compiles the CUDA backend — `libcudarc.rmeta` built + the new guard didn't fire = proof the cfg is injected |
| G3 | `--gpu auto/on/off` verified **live** on the RTX 3060: `auto`→`enabled:true`, `on`→`enabled:true`, `off`→`enabled:false` (CPU pinned despite `available:true`) |
| G5 | `fmt` clean · `clippy --all-targets --all-features -D warnings` clean · **858 lib + 121 bin tests pass** · `doc` builds |
| G6 | Windows `Cargo.toml`/`build.rs`/`cuda.rs` regions **byte-identical**; Windows default build green + ran the G3 test |

Evidence bundle: `qa/evidence-bundles/mesa/`.

> ⚠️ **G5 note:** the `--bin camelid` test harness hits `STATUS_STACK_OVERFLOW` in a **BASALT** CLI-parse test (`gemma4_generate_parses_forced_decode_flags`) on Windows debug — a **pre-existing Windows-only artifact**, not a MESA regression (the `/STACK:8388608` link arg covers the `camelid` bin, not its test-harness exe). Passes with `RUST_MIN_STACK=16777216`. Linux CI does not hit this.

## Open hardware gate (post-merge, cannot close on Windows or in CI)
On the RTX **Linux** box: confirm a default `cargo build --release` (no `--features`) reports `available:true, enabled:true` out of the box; re-run `cuda_parity` for **token-identical** greedy output vs the CPU/llama.cpp oracle under the default build; spot-check `--gpu off` forces CPU. File under `qa/evidence-bundles/mesa/parity/`. **Status: landed + CI-pending; GPU-runtime parity on x86_64 Linux pending hardware.**

## Docs follow-up (flagged, NOT edited — kept scope tight)
Two lines now understate the Linux CUDA story and want a truthful touch-up in a separate change (avoided here so this stays a low-blast-radius build-wiring PR + the CAIRN drift-gate can't be reliably run on Windows):
- `README.md:185` — "CPU; CUDA **buildable**…" → CUDA is now default-compiled on Linux x86_64 (engages automatically on supported NVIDIA setups).
- `COMPATIBILITY.md:117` — "An **opt-in** CUDA backend (`cargo build --features cuda`)" → now default-on for Windows + x86_64 Linux; `--features cuda` is only needed for aarch64/BSD.


timtoole02:
## Rebased onto v0.4.2, and the deferred Linux GPU gate is now closed — it found two startup crashes

The branch was 105 commits behind and built against a v0.3.1-era base; its failing `rust (windows-latest)` job was a pre-existing `String`/`&str` error in `src/chat/tools.rs`, unrelated to MESA and already fixed on main. Rebasing cleared it with no conflicts.

The "Linux runtime cuda_parity" gate that was filed as post-merge has now been **run**, on WSL2 Ubuntu + RTX 3060 with GPU passthrough. It found two independent process-killing bugs, both made reachable by compiling CUDA into the default Linux build, and **neither catchable by the guards already in place**.

### 1. Missing NVRTC aborted the process
cudarc panics from inside its lazy NVRTC loader, so the `.map_err` on every `compile_ptx_with_opts` never runs. Installing the NVIDIA driver does **not** install NVRTC — it ships with the CUDA toolkit — so the ordinary driver-only Linux host could not start camelid at all:

| scenario (default build, no NVRTC) | before |
|---|---|
| GPU present, `--gpu auto` | panic at `cudarc-0.19.7/src/lib.rs:200`, process died |
| **`--gpu off`** (explicitly CPU!) | same panic — capability detection reaches the CUDA path regardless of the switch |

`cuda.rs` already guarded the *driver* load against this exact failure class with `catch_unwind` ("so `--all-features` builds fall back to the CPU path instead of aborting the process"). NVRTC just was not covered. Now guarded at all three entry points: `cuda.rs`, `cuda_resident.rs` (the resident decode engine), `diffusion_gemma/cuda.rs`.

### 2. `CUDA_VISIBLE_DEVICES=-1` aborted the process
Independent of the first: `free(): double free detected in tcache 2`, **with or without NVRTC**. That is a glibc SIGABRT, not a Rust panic, so `catch_unwind` cannot intercept it.

Nothing in the CUDA path ever read `CUDA_VISIBLE_DEVICES`, even though this repo documents it as the way to pin a run to the CPU (`alloc_gate.rs`, the decode receipts) and `main.rs:3949` sets it for a child process. The driver was initialized anyway. Now honored at all three driver entry points — correct on every platform, since masking all devices *is* a request to skip CUDA.

### Post-fix verification (default Linux build, no `--features` flag)

| scenario | result |
|---|---|
| A. no NVRTC, GPU present, `--gpu auto` | starts, falls back to CPU |
| B. no NVRTC, `--gpu off` | starts, CPU |
| C. `CUDA_VISIBLE_DEVICES=-1` | starts clean — `[hw] GPU: none detected` |
| D. NVRTC present, `--gpu on` | starts, `cuda_resident_active: true`, `cuda_resident_q8_runtime` |

Plus the original parity leg: same greedy prompt, same binary, `--gpu off` vs `--gpu on` → **token-identical output**, with `[gpu] all 16 layers resident in VRAM` on the GPU leg. Kernels confirmed compiled into the default binary (76 `__global__` kernels, `--gpu` switch present) with no feature flag.

### Docs
README claimed Linux was "CPU; CUDA buildable" and COMPATIBILITY described CUDA as opt-in behind `--features cuda` (already untrue on Windows). Both corrected, and both now say explicitly that compiling the path in is **build wiring, not a support claim** — the validated rows stay pinned to the Windows GPUs they were recorded on.

### Known limitations, stated not buried
- **Not verified off WSL2.** No bare-metal Linux GPU host was available, so whether the double free reproduces on a native driver stack is unproven. The guard is correct regardless.
- **The Linux tarball does not bundle NVRTC** the way the Windows ZIP does (`package-windows-cuda.ps1`). So driver-only Linux users now get a clean CPU fallback rather than a crash, but not GPU. Making MESA's premise fully land on Linux needs NVRTC packaged into the release artifact — worth a follow-up.
- **First GPU start on Linux takes ~124s** compiling the resident kernel blob through NVRTC (uncached). Not a blocker; worth knowing.
- Scenarios A/B still print the caught cudarc panic to stderr on startup. Functionally fine, cosmetically alarming for a default build — follow-up.

PR #491, opened 2026-07-23
feat(verify): seal and verify the agent_eval promotion receipt

karan68:
﻿## What

The `agent-eval` gate mints the receipt that justifies flipping `tool_capable` true (its own docs: "the receipt that justifies flipping `tool_capable` true"). Yet it was the **only** agent receipt emitted with **no digest** -- unlike syscap / orchestration / bench, nothing could prove it intact, and `camelid verify-receipt` could not check it. This seals it and teaches the verifier to verify it, so the one promotion-bearing receipt is now tamper-evident.

## Changes

- **Seal** ([src/chat/agent_eval.rs](../blob/main/src/chat/agent_eval.rs) `finish`): stamps `receipt_id` via the shared `receipt::receipt_id_over` primitive. Additive to `camelid.agent_eval/v1`; the 9 committed pre-seal receipts are untouched.
- **Verify** ([src/receipt/agent.rs](../blob/main/src/receipt/agent.rs)): adds `AgentReceiptClass::Eval` + `EVAL_RECEIPT_SCHEMA_V1`. Its honest-scope invariant is internal consistency -- `promotion_eligible` must equal `outcome == "PASS"`; a receipt claiming eligibility without a PASS is rejected.
- **Legacy tolerance**: an unsealed receipt (no `receipt_id`) is reported as an *unsealed legacy* receipt, not as a malformed body.
- **Accuracy**: the reported `host` is now `Option` (agent_eval carries no host block, so no more `host ?/?`), and the verdict lines no longer claim "promotes no capability" for the one gate that *is* promotion-bearing.

## Testing

- `cargo fmt --check`; `cargo clippy --all-targets --all-features -- -D warnings` -- clean
- `cargo test --lib receipt::agent` -- 19 pass (new: sealed-eval verifies, eligibility inconsistency rejected, unsealed-legacy reported)
- **End-to-end on Windows + Qwen3-4B-Q4_K_M** (real `agent-eval` battery, PASS): the freshly minted receipt verifies (`AGENT RECEIPT VERIFIED`), a committed legacy `qa/agent-eval/*.json` is reported unsealed, and a tampered copy fails (`self-digest`).

Follows up #490 (which sealed/verified the other three agent receipts); no schema/ledger consumer parses these receipts, so the added field is safe.

---

## Behavioral notes (per review)

[...]

PR #494, opened 2026-07-23
feat(verify): audit committed receipt integrity + a CI gate

karan68:
﻿## What

`camelid verify-receipt` proves ONE receipt intact, but nothing checked the hundreds of committed sealed receipts. A bad merge, a find-replace, or a hand-edit could silently corrupt a parity / agent / distributed receipt and no test would notice -- yet "local AI you can verify" leans entirely on that evidence. This adds the mechanical layer under "git history + review."

- **`receipt::audit`** -- walks a directory and verifies the self-digest of every sealed receipt (parity, distributed, agent) via the shared `receipt_id_over` primitive. Unsealed / non-receipt JSON is skipped, so it never false-fails on the ~100 ad-hoc evidence schemas.
- **`camelid verify-receipts <dir>`** -- the CLI surface (defaults to `qa`).
- **`tests/receipt_integrity.rs`** -- runs the audit over `qa/` inside the standard `cargo test --all-targets`, so a new corruption fails CI on **every OS leg** with no workflow-file wiring.

## Finding: this immediately surfaced 9 already-broken seals

Running the audit against `qa/` found 9 committed sealed receipts whose stored `receipt_id` no longer matches their body (the existing `verify-receipt` agrees -- not an audit artifact). Two root causes, both grounded in git history:

- **Privacy scrubs that edited receipts after sealing (4):** `syscap-1782688037` carries the `<home>` marker (commit `d2825069` "scrub Windows home path"); two `distributed/*` receipts (commit `cfaed035` "redact LAN IPs/SSH-key path"); one oracle receipt. Redacted correctly for privacy, but **never re-sealed**.
- **Sealed under an earlier serialization (5):** `bench-rung4-1782702313` (single commit, never edited, carries high-precision floats) and 4 `capability/mac_*/parity-receipt.json` (single commit, never edited).

These are grandfathered by an **in-code baseline** (`receipt::audit::BASELINE`) keyed on their exact `(stored, computed)` digest pair, each with a documented reason. So the gate enforces integrity **forward** without rewriting historical provenance -- and because the key is content-derived, any FURTHER edit to a baselined receipt changes its `computed` digest, drops the match, and fails.

The maintainer may separately decide to re-seal the scrub-redacted receipts (legitimately: seal the committed redacted bytes) and investigate the serialization-drift ones; the error message points that way. This PR does not touch any committed receipt.

## Testing

- `cargo fmt --check`; `cargo clippy --all-targets --all-features -- -D warnings` -- clean
- `cargo test --lib receipt::audit` -- 10 tests (verify, tamper, baselined-debt, further-edit-fails-again, stale-entry, unsealed, skip-unknown-schema, recursive walk)
- `cargo test --test receipt_integrity` -- the gate passes over the real `qa/`
- **End-to-end**: `verify-receipts qa` -> `3 verified, 9 allowlisted debt, 8 unsealed, 0 failed` (exit 0); a further-edited baselined receipt -> `FAIL digest` (exit 1).

Follows #490 / #491 (which sealed and verified the agent-receipt families); reuses their `receipt_id_over` primitive as the single source of truth.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1233,6 +1233,15 @@ enum Command {
         #[arg(long)]
         threads: Option<usize>,
     },
+    /// Audit the self-digest of every sealed receipt under a directory tree
+    /// (parity / distributed / agent families). A corrupted or hand-edited
+    /// committed receipt fails; unsealed and non-receipt JSON is skipped. This is
+    /// the mechanical companion to review for the receipt evidence base.
+    VerifyReceipts {
+        /// Directory walked recursively for `*.json` receipts.
+        #[arg(default_value = "qa")]
+        dir: PathBuf,
+    },
     /// Recompute and stamp `receipt_id` on a receipt body. Emitters (e.g. the
     /// chat-parity harness) delegate sealing here so canonical serialization
     /// and digesting live in exactly one implementation.
@@ -2571,6 +2580,9 @@ async fn main() -> anyhow::Result<()> {
             })
             .await;
             std::process::exit(outcome.exit_code());
+        }
+        Command::VerifyReceipts { dir } => {
+            std::process::exit(camelid::receipt::audit::run(&dir));
         }
         Command::SealReceipt { input, output } => {
             let raw = std::fs::read_to_string(&input)?;
--- a/src/receipt/audit.rs
+++ b/src/receipt/audit.rs
@@ -1,0 +1,517 @@
+//! Repo-wide integrity audit for the committed **sealed** receipts.
+//!
+//! [`super::verify`] and [`super::agent`] prove ONE receipt intact; this walks a
+//! directory tree and checks EVERY sealed receipt's self-digest, so a corrupted
+//! or hand-edited committed receipt cannot pass review unnoticed. It is the
+//! mechanical companion to the CI gate - the automatic layer under "git history
+//! + review" that the receipt scheme otherwise leans on entirely.
+//!
+//! Scope, deliberately narrow so the audit never false-fails:
+//! - Only receipts whose `schema` is one of the families sealed with the shared
+//!   [`receipt_id`](super::receipt_id_over) convention (parity, distributed, and
+//!   the agent family) are checked. Every other JSON file - the many unsealed
+//!   evidence / summary / prompt-pack schemas - is skipped.
+//! - A sealed-family receipt that carries no `receipt_id` (e.g. an `agent_eval`
+//!   receipt minted before sealing) is reported as *unsealed*, not failed.
+//! - Only a receipt whose stored `receipt_id` does not match its recomputed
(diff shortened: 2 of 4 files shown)


karan68:
﻿Thanks — this was a genuinely useful review; the fail-open point was correct and worth fixing. Addressed in `df7b64dc`. Gates green (fmt, `clippy --all-targets --all-features -D warnings`, `receipt::audit` 11 tests, integration gate over real `qa/`).

**#1 fail-open on unparseable/unreadable — FIXED (your "address before merge").** The walk no longer silently `continue`s. `AuditReport` now tracks `unreadable` and `unparseable`; both are failures (`ok()` requires them empty) and the integration test asserts they're zero. A merge-conflict-marker `.json` now fails — new `audit_dir` test proves it. Running it immediately found 2 pre-existing non-JSON `.json` under `qa/`: a key=value benchmark **plan** (`same-host-plan.json`, not JSON, misnamed) and a **0-byte** capture (`unload.json`). Those are documented in a tiny `UNPARSEABLE_ALLOWLIST` (path-keyed, EOL-robust); anything new fails.

**Q1 (was it deliberate?):** oversight, not deliberate — good catch. Surfaced + asserted now.

**#2 serde_json coupling — documented.** Added a note on `BASELINE` that the `computed` values are bound to `serde_json` serialization. Worth stressing: this isn't gate-specific — a float-formatting change would break **every** committed receipt carrying a float, since the whole receipt scheme relies on that serialization being reproducible. The gate just makes it visible. **Q2:** I didn't unilaterally pin `serde_json` (it's a crate-wide `1` dep — a maintainer call), but the doc says to treat a bump as a receipt-format event and re-derive. Happy to pin if you'd prefer.

**#3 unsealed not enforced — clarified, and enforcement would false-fail.** Grounded it: the 8 "unsealed" files are `camelid.parity-receipt/v1` **with no receipt_id** (the ornith `RECEIPT_ITEM*` evidence), so legitimately-unsealed parity receipts exist in-tree — failing on "sealed-family but unsealed" would break CI today. Clarified the module docstring that the gate checks **integrity of sealed receipts, not that sealing happened**. **Q3:** so, no — not without its own allowlist; out of scope here.

**#4 sealed_schemas completeness — `KEEP IN SYNC` comment added.**

**Nits:** walk test now uses `tempfile::tempdir()` (unique + panic-safe cleanup) and includes the conflict-marker case; commented the deliberate test-scoped `.leak()`. Acknowledged-wontfix: FAIL→stderr (the whole report is one cohesive stdout block; the gate reads the struct, not streams), symlink-cycle guard (qa/ is committed, no symlinks), `matched_baseline` positional indices (internal, always the same slice, tested), streaming read (fine at qa/ scale), and `verify-receipt`/`verify-receipts` naming.

PR #506, opened 2026-07-25
[DON'T MERGE] Remote agent control over encrypted mobile sessions

karan68:
## Summary

- add a session-owned agent runtime, strict `camelid.remote/v1` protocol, durable SQLite authority, and Noise IK end-to-end encryption
- add a blind bounded relay plus asynchronous local pairing, per-device revocation, replay, approvals, cancellation, and restart recovery
- add an Expo/React Native Android development client with Keystore-backed keys, per-history replay projections, new-session creation, and explicit session continuation
- add the loopback-only Remote management view, capability disclosure, emergency disable, and no-store QR workflow
- keep many durable histories but exactly one active execution session; dormant histories remain replay-only until an atomic activation

## Safety properties

- private host/device keys and relay capabilities do not enter application logs or plaintext SQLite
- commands and approvals are identity-, session-, turn-, call-, and digest-bound; store commit precedes broadcast
- relay sees only routing metadata and bounded encrypted binary frames; it has no inference, tool, or plaintext protocol dependency
- remote capability scope is fixed locally; shell/network default off and remote yolo, persistent grants, MCP, subagents, and GUI control are unavailable
- legacy CLI saved sessions are replay-only because they lack exact model artifact identity, and historical grants are never restored

## Validation

- protocol: 19 tests
- durable store: 25 tests
- blind relay: 11 tests plus Windows persistence gate
- root remote slice: 42 tests
- mobile: 66 tests, TypeScript, Expo lint, and native artifact seal
- mobile production dependency audit: 0 vulnerabilities after a scoped patched `brace-expansion` override
- strict Rust Clippy and rustfmt
- physical Android RMX5003: pair/revoke/re-pair, restart/replay, read-only Qwen task, multi-history create/switch/continue, contextual continuation, and real-time catalog refresh from event 15 to 20

The governing architecture and receipts contain exact model, device, artifact hashes, screenshot hashes, and bounded non-claims.

## Manual qualification

The repository now includes `docs/architecture/REMOTE_AGENT_CONTROL_MANUAL_TEST_PLAN.md` with 48 normal, abnormal, adversarial, concurrency, lifecycle, network, storage, accessibility, and recovery scenarios. Each scenario defines independent evidence and pass/fail criteria; priority-0 failures stop promotion.

The branch was rebased onto upstream `cc24dfe7`. Conflict resolution preserves normal `serve` Workspace CLI token issuance while `agent host` remains credential-free; `remote_host_state_never_enables_workspace_cli_authority` and the existing exact-token Workspace CLI test both pass.

Physical RMX5003 follow-up is recorded in `docs/architecture/REMOTE_AGENT_CONTROL_PHYSICAL_097FA7E0_RECEIPT.md`. Fresh camera pairing, Noise authentication, read tool execution, exact write approval, strong-biometric Allow once, explicit denial, approval expiry/abort safety, new/dormant/continued histories, real-time catalog refresh, and restart replay passed on the disposable workspace. Physical testing found and fixed Android navigation overlap on approval Deny/Abort (`699fc2b9`).

One model-context quality finding remains: after a denied edit, a later context-only prompt incorrectly claimed the denied replacement had been written, although durable events and filesystem content correctly showed it had not. This is not hidden as a transport pass and must be resolved or explicitly accepted before merge.

## Do not merge

This is development evidence only. It does not claim a production relay, complete Android/OEM qualification, iOS readiness, push delivery, retention/delete UX, packaging, or independent cryptographic/security review. Keep this PR open for architecture and security review; do not merge until the remaining Phase 5 release gates are explicitly resolved.

The full mobile development dependency audit still reports 32 high-severity advisories through test/lint tooling. npm's automatic remediation would make breaking Expo/React Native changes, so those findings are disclosed rather than force-fixed; the production-only audit is clean.


karan68:
Hey @timtoole02 , can you please put this to an end to end review , this is one of the major feature i have been working on

timtoole02:
@karan68 Ok, this is insane. I didn't expect the protocol/store layer to be this tight. `settle_approval_in` binding settlement to approval+session+turn+call+digest in one row update and erroring if `changed != 1`, command idempotency by `(device_id, command_id)` with a request_digest conflict check, canonical JSON rejecting floats so a digest can't drift on number rendering, chunk framing carrying a whole-message SHA-256, per-byte tamper tests across both handshake and transport. That's the right shape. Real work.

Full disclosure: I ran AI over this. 51k insertions across 156 files is not something I can hold in my head in one sitting. Felt weird for about four seconds and then I remembered what we're building here.

Not merging yet. Here's the list.

**Blockers**

1. The loopback guard is CSRF protection, not authorization. `local_management_request_allowed` (src/api/mod.rs:18615) checks Host/Origin, which stops a browser page and nothing else. Any local process sets those headers freely, so it can POST /api/agent/remote/pairing, read route_id + host_noise_public + pairing_secret straight out of the response, attach to the relay, poll status for the confirmation_id, and confirm itself. Headless. No human sees the fingerprint. And `confirm_remote_pairing` doesn't take the fingerprint at all, only `(confirmation_id, accepted)`, so the coordinator ends up comparing its stored fingerprint against itself. The displayed fingerprint is decorative. Meanwhile `remote_host_state_never_enables_workspace_cli_authority` deliberately asserts there's no token on this surface, so the strongest local credential we have is switched off on the one surface that grants persistent remote agent control. That inversion needs an answer before anything else moves.

2. CI doesn't cover the two riskiest crates. `default-members = ["."]`, so root clippy is root-only, and the added `-p` lists skip `camelid-remote-store` and `camelid-relay`. The relay is never compiled in CI at all. The store's 25 tests never run. Nothing in ci.yml mentions `mobile`, so the 66 mobile tests, tsc, Expo lint, and verify-native-artifacts.mjs are all one-time prose claims with no gate. Wrong way round for this repo.

3. The model-context finding. You're right to surface it and right that it blocks. The phone screen being trustworthy about what happened on my machine is the entire product. A model that misreports write state in the denied direction breaks exactly the boundary approvals exist to defend. Durable events being correct is necessary, not sufficient. The user reads the transcript.

4. The committed .so blobs. verify-native-artifacts.mjs hashes them against a manifest in the same commit, which is an integrity check on itself, not provenance. Nothing ties the blob to crates/camelid-remote-crypto-ffi. Manifest also says rust 1.95.0 while CI pins 1.89.0. Either build it in CI from source or make the receipt a reproducible-build comparison.

**Should fix**

- route_id is the only device credential and it's in the URL path. Revocation doesn't rotate it, so a revoked phone keeps relay attachment forever, can hold all 4 MAX_DEVICES_PER_ROUTE slots and lock out my real device. Path secrets also land in every proxy access log, which cuts against the "capabilities do not enter logs" claim.
- `revoke_push` has no auth at all. Capability in the path, no bearer, unlike every sibling endpoint.
- No pong deadline. bridge_host pings but never checks a pong came back, so a black-holed host socket holds the route and the real host gets 409 until TCP gives up.
- Relay has no rate limiting, no global connection cap, no TLS, and main.rs wires UnavailablePush, so /v1/push/* is dead surface that always 503s. Feature-gate it off.
- DeviceKeyVault.createKey doesn't set setUnlockedDeviceRequired or setUserAuthenticationRequired, no StrongBox. secureHostStore.ts uses WHEN_UNLOCKED_THIS_DEVICE_ONLY. Same PR, two policies, Android is the weaker one.
- StaticKeypair::drop uses fill(0), which the optimizer can elide. Use zeroize. The intermediate snow keypair Vec isn't wiped either.
- Hand-rolled constant_time_equal in the relay, and `presented_route != route` in register_push is plain PartialEq on a secret. Pull in subtle.
- camelid_remote_crypto_ffi.kt is duplicated verbatim in two paths. Add a CI byte-equality check.

**One question**

Is there an anti-replay cache for the Noise IK first handshake message? One-time pairing_secret plus a single active offer covers it in practice but I don't see an explicit test for a replayed initiator message against a consumed offer.

**On scope**

The boundaries are already in the code, so let's use them. Land camelid-remote-protocol + crypto + store first, behind no user-facing surface. Host slice + CLI + loopback UI second, which is where the pairing authority question gets settled. services/camelid-relay has zero dependency on Camelid and a totally different deployment and threat model, and mobile/ brings a 15k-line lockfile and 32 unresolved dev advisories into an inference engine's dependency graph. Both of those want their own repos.

Separate from the code: I need to think about whether I want to own a hosted relay and an Expo app. Uptime, TLS certs, push credentials, Play Store, OEM fragmentation. That's a permanent commitment bolted onto a project whose whole pitch is one local Rust binary with no dependencies. Not a knock on the work. Just a thing I have to answer before any of it merges, and that answer is on me, not you.

Give me the four blockers and we'll go from there.

PR #509, opened 2026-07-26
fix(desktop): clarify sidecar startup failures

karan68:
## Summary

Makes Camelid Desktop's sidecar-startup failure surface reliable rather than event-timing-dependent. The native shell remains additive: it owns only sidecar lifecycle and splash feedback; the engine remains the authority for health, models, inference, and the embedded UI after health succeeds.

## Why

The original splash could remain on an indefinite spinner when a healthy sidecar never appeared, because the startup worker tried to synchronously drain an alive child's stderr pipe. It could also lose a fast early error emitted before splash JavaScript registered its listener. Finally, a later TCP connection to the selected port was correlation, not proof that a bind conflict caused the engine exit.

### Terminate and reap before draining stderr

- All unsuccessful health-gate paths now converge through `finish_startup_failure`.
- It kills and waits for the sidecar before `read_to_string` drains stderr, guaranteeing EOF even when the 40-second health gate times out while the child remains alive.
- The original startup error and captured diagnostics remain available to the user-facing technical-details pane.

### Durable, replayable startup state

- The native shell maintains a serializable `StartupSnapshot` containing either current status or structured error information.
- Native state is updated before every `engine-status` event.
- The splash first registers its event listener, then invokes the `startup_snapshot` command and renders the returned authoritative state. A missing sidecar or other fast failure can no longer be lost before listener registration.
- Error detail continues to be rendered via `textContent`, not HTML injection.

### Confirmed bind diagnostics only

- Removed the post-exit loopback TCP probe.
- A generic early engine exit remains `Engine startup failed` even if some other process later listens on the chosen port.
- `Sidecar port unavailable` is now assigned only for loopback-reservation failures or explicit, captured OS bind diagnostics from the sidecar (`address already in use`, Windows 10048, or Unix 98).
- User documentation now describes an engine-reported bind failure, not an inferred bind race.

## User-facing Behavior

| Condition | Splash title | Guidance |
| --- | --- | --- |
| Sidecar binary missing | Camelid engine is missing | Reinstall or restore `camelid.exe`, then retry. |
| Confirmed sidecar bind failure | Sidecar port unavailable | Close the conflicting local process and retry. |
| Sidecar never becomes healthy | Engine startup timed out | Retry and use technical details for a persistent failure. |
| Other early sidecar exit | Engine startup failed | Review technical details and retry. |

The splash remains fail-closed until `/v1/health` returns `200`. Passing health only unlocks navigation to the existing engine UI; it does not claim that a model is ready.

## Tests

- `unhealthy_live_child_is_reaped_before_stderr_is_drained` covers a live child holding stderr open, verifies timeout classification, captured stderr, and reaping without a multi-second block.
- `only_captured_bind_diagnostics_change_the_final_error_kind` verifies that captured bind output maps to the port error while generic stderr stays generic.
- `startup_error_is_replayable_after_the_listener_would_register` verifies native state retains an early structured error for later splash retrieval.

## Validation

- `cargo test -p camelid-desktop --all-targets` - 10 passed
- `cargo clippy -p camelid-desktop --all-targets -- -D warnings`
- `scripts/check-public-scrub.sh`
- `git diff --check`
- Editor diagnostics clean; touched files retain LF EOLs

## Scope

- No changes to engine inference, model support claims, readiness semantics, API behavior, or Web UI behavior after successful health gating.
- No fixed-port assumptions: Desktop continues using an ephemeral loopback address.

## Merge Checklist

- [x] Local focused tests and static gates pass
- [ ] Fresh PR CI matrix is green
- [ ] Review comments are resolved

### DIFF
--- a/camelid-desktop/README.md
+++ b/camelid-desktop/README.md
@@ -46,6 +46,26 @@ camelid-desktop.exe ──spawns──▶ camelid.exe serve --addr 127.0.0.1:<ep
 On window close the sidecar is terminated cleanly; a Windows **job object** with
 `KILL_ON_JOB_CLOSE` is the backstop so a desktop crash cannot orphan a `camelid` process.
 
+## Startup failures
+
+The splash is fail-closed: it stays visible until the sidecar returns `200` from
+`/v1/health`. It polls every 350 ms for up to 40 seconds. Native startup state is retained and
+replayed after the splash listener registers, so a fast failure cannot be lost before the page
+loads. A failure shows an actionable title and next step first, followed by the engine's actual
+error and captured stderr under **Technical details**; it never navigates to a fake-ready UI.
+
+| Splash error | Meaning | Next step |
+| --- | --- | --- |
+| **Camelid engine is missing** | `camelid.exe` was not found beside the desktop executable, in the bundled sidecar resources, or on `PATH`. | Reinstall Camelid Desktop or restore `camelid.exe` beside `camelid-desktop.exe`, then retry. |
+| **Sidecar port unavailable** | The engine reported that it could not bind Camelid's selected ephemeral loopback port. This is not a fixed `8181` port conflict. | Close the conflicting local process and retry. |
+| **Engine startup timed out** | The sidecar did not pass the 40-second `/v1/health` gate. | Retry, then use the visible technical details to diagnose a persistent failure. |
+| **Engine startup failed** | The sidecar exited before it became healthy for another reason. | Review the visible technical details and retry. |
+
+Model readiness is separate from sidecar startup. Once `/v1/health` passes, Desktop navigates to
+the engine's existing UI; if no eligible model is loaded, that UI remains the authority and shows
+its normal model-required state. Desktop does not claim a ready model or manufacture a model error
+on the splash.
+
 ## Requirements
 
 - Windows 10/11 with the **WebView2 runtime** (preinstalled on current Windows 10/11; the
--- a/camelid-desktop/splash/index.html
+++ b/camelid-desktop/splash/index.html
@@ -26,8 +26,12 @@
       }
       .title { font-size: 20px; font-weight: 600; letter-spacing: 0.2px; }
       .status { color: #9aa6b2; min-height: 1.4em; }
-      .err { color: #f0a3a3; white-space: pre-wrap; text-align: left; max-width: 90%;
-             font-family: ui-monospace, "Cascadia Code", Consolas, monospace; font-size: 12.5px; }
+      .err { color: #f0a3a3; text-align: left; max-width: min(90%, 680px); }
+      .err-title { font-size: 16px; font-weight: 600; }
+      .err-guidance { color: #d8dee6; margin-top: 6px; }
+      .err-details { margin-top: 12px; }
+      .err-detail { margin: 8px 0 0; max-height: 15rem; overflow: auto; white-space: pre-wrap;
+            font-family: ui-monospace, "Cascadia Code", Consolas, monospace; font-size: 12.5px; }
       .spinner {
         width: 26px; height: 26px; border-radius: 50%;
         border: 3px solid #2a323c; border-top-color: #6ea8fe;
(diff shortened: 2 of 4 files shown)


karan68:
Implemented the requested lifecycle fixes in `3795c930`.

- **Timeout deadlock:** all failed health-gate paths now kill and reap the child before draining stderr. `unhealthy_live_child_is_reaped_before_stderr_is_drained` exercises a live child holding the pipe open.
- **Missed early splash error:** native `StartupState` now persists the latest structured status/error; the splash subscribes first, then invokes `startup_snapshot`. `startup_error_is_replayable_after_the_listener_would_register` covers the durable state primitive.
- **Speculative port conflict:** removed the post-exit TCP probe. `PortUnavailable` now requires reservation failure or captured engine bind diagnostics; generic stderr remains `StartupFailed`. The finalizer regression covers both cases.

Local `cargo test -p camelid-desktop --all-targets` passes (10 tests), along with strict Clippy, formatting, public scrub, diff, diagnostics, and LF checks. Fresh CI is running on this head; keeping `[DON'T MERGE]` until it completes.

karan68:
I checked all three findings against the PR head (`3795c930c`) rather than the diff-at-open, and **all three are already fixed on this branch** — each with a regression test whose name states the finding. I think this review was written against an earlier revision.

Evidence, all at `3795c930c`:

## 1. `drain_stderr()` before kill/reap — already fixed

The failure path does not call `drain_stderr()` directly. `spawn_on_port` routes every health failure through `finish_startup_failure`, which reaps first:

```rust
match wait_for_health(port, &mut child) {
    Ok(()) => Ok(Engine { .. }),
    Err(err) => Err(finish_startup_failure(&mut child, err)),
}

fn finish_startup_failure(child: &mut Child, error: EngineError) -> EngineError {
    // A timed-out child still owns the stderr pipe. Reap it before a blocking read so the
    // startup worker cannot get stuck and leave the splash spinner visible indefinitely.
    let stderr = terminate_and_collect_stderr(child);
    ...
}

fn terminate_and_collect_stderr(child: &mut Child) -> Option<String> {
    let _ = child.kill();
    let _ = child.wait();
    drain_stderr(child)
}
```

The regression you asked for — a child that stays alive, keeps stderr open, and never returns health — exists as `unhealthy_live_child_is_reaped_before_stderr_is_drained()`. It asserts the child is alive before the call (`try_wait().is_none()`), that the error is `StartupTimeout`, that stderr is still captured afterwards, that the whole thing completes in under 3s, and that the child is reaped (`try_wait().is_some()`).

## 2. Fast failures lost before the splash listener registers — already fixed

There is durable native state, not just events. `StartupState(Mutex<StartupSnapshot>)` is `.manage()`d; `emit_status`/`emit_error` write to it **before** emitting; and `startup_snapshot` is exposed as a Tauri command:

```rust
/// Events improve responsiveness; this native state prevents early failures being lost before
/// the splash listener has registered.
#[derive(Default)]
struct StartupState(Mutex<StartupSnapshot>);

#[tauri::command]
fn startup_snapshot(state: State<'_, StartupState>) -> StartupSnapshot { state.snapshot() }
```

The splash queries it **after** registration completes, which closes the race in both directions:

```js
window.__TAURI__.event.listen("engine-status", (e) => {
  renderStartupState(e.payload);
}).then(() => {
  window.__TAURI__.core.invoke("startup_snapshot").then(renderStartupState)
```

Covered by `startup_error_is_replayable_after_the_listener_would_register()`.

## 3. Port classification from an unproven TCP connect — already fixed

Post-exit TCP inference was removed. Classification now requires a captured OS bind diagnostic from the sidecar's own stderr:

```rust
/// Recognize only explicit OS bind diagnostics captured from the sidecar.
///
/// The ephemeral reservation must be released before the sidecar binds. A listener observed
/// *after* a generic sidecar exit is not proof that it caused the exit, so it must not control
/// the user-facing diagnosis. The engine's own bind failure is the authority instead.
fn stderr_is_bind_failure(stderr: Option<&str>) -> bool { ... }
```

Two tests cover it: `captured_bind_diagnostic_is_required_for_port_classification()` (accepts the Linux `os error 98` and Windows `os error 10048` diagnostics, rejects `"model metadata is invalid; sidecar stopped before binding"` and `None`) and `only_captured_bind_diagnostics_change_the_final_error_kind()`, which drives `finish_startup_failure` with two live children — one emitting a bind diagnostic, one generic — and asserts only the first becomes `PortUnavailable`.

## On the CI note

> those tests do not exercise the live unhealthy-child path or the Tauri listener-registration race

Both are exercised. `unhealthy_live_child_is_reaped_before_stderr_is_drained` spawns a real long-lived child holding a real stderr pipe with no health endpoint; `startup_error_is_replayable_after_the_listener_would_register` drives the durable-snapshot replay path. The `desktop` job runs `cargo test -p camelid-desktop --all-targets`, so both ran in the green CI on this head.

If you were looking at an editor-resolved copy rather than the branch, that would explain it — worth double-checking the revision, since the three fixes and the three findings line up one-to-one. Happy to look again at anything I've misread.

PR #510, opened 2026-07-26
ci(release): verify shipped artifacts before publish

karan68:
## Problem

`release.yml` builds, checksums and signs three archives, then publishes them without ever opening or running one. Every existing gate inspects either the source tree (`ci.yml`) or the staging directory (`scripts/package-linux-cuda.sh`, `scripts/package-windows-cuda.ps1`).

That leaves a class of defect that only appears once the bytes are packaged. v0.4.1 → v0.4.4 were all that class, tagged within about a day:

| Tag | PR | Fix |
|---|---|---|
| v0.4.1 | #500 | resolve GGUF loads under a Windows `\\?\` verbatim models-dir |
| v0.4.2 | #501 | swap models on load instead of dead-ending on the fit guard |
| v0.4.3 | #503 | ship NVRTC in the x86_64 Linux tarball |
| v0.4.4 | #505 | stop shipping NVRTC twice in the Linux tarball |

## What this does NOT claim

**The existing staging guards already work and are not duplicated here.** `scripts/package-linux-cuda.sh` fails on a missing `libnvrtc.so.<major>` soname, a dangling symlink, and a duplicated payload. This PR does not re-litigate #503/#505.

The gaps are what comes *after* those guards:

1. They run before `tar -czf` / `Compress-Archive`, so the archive itself was never inspected.
2. They live inside a conditional packaging step (`if [ "$label" = "linux-x86_64" ]`). A renamed matrix label silently skips them and still ships. A check on the extracted archive cannot be skipped by a packaging branch that never ran.
3. The `.sha256` sidecars are written once on the build runner and never re-verified after the artifacts round-trip through `upload-artifact` → `download-artifact` into the publish job.
4. No artifact was ever executed, on any platform.
5. Nothing asserted that the binary's self-reported version matches the release tag.

## What ships

Two dependency-free Node scripts, wired into `release.yml` between packaging and upload.

**`scripts/check-release-artifact.mjs`** — execution-free, verifies the extracted archive:

- checksum round-trip, parsing both sidecar conventions (`shasum` writes a trailing newline; PowerShell `Out-File -NoNewline` does not)
- per-label archive layout — `tar -czf "$dist.tar.gz" "$dist"` nests under a wrapper directory, `Compress-Archive -Path "$dist/*"` is flat at the zip root
- required payload, NVRTC pair, forbidden `.alt` backend
- byte-identical duplicate payloads, dangling or escaping symlinks
- build junk (`.pdb`, `target/`, credential-shaped files)
- POSIX executable bit on the binary
- that the NSIS installer was moved out of the desktop portable zip

**`scripts/smoke-release-artifact.mjs`** — boots the shipped binary out of the extracted archive, model-less:

- `/v1/health` (asserting `loaded_now`/`generation_ready` are false, so the run is provably model-less), `/api/capabilities`, and `GET /` requiring both the app-shell title and a hashed `/assets/*.js` bundle — that last one catches a web UI build that never made it into the binary
- asserts binary version == release tag
- tears the process tree down and checks stderr for crash markers

On a runner with no NVIDIA driver, this is a standing regression guard for `02f6a8b04` — *"a missing NVRTC must fall back to CPU, not abort the process"*.

**Publish job** re-hashes every artifact against its sidecar before uploading, and requires the archive/sidecar sets to match in both directions.

## Design notes

- **Verify the extracted archive rather than parsing archives.** Round-tripping through the same extraction tools users run is a stronger signal than a hand-rolled tar/zip reader, and it leaves the contract a pure function of a directory tree — which is what makes it exhaustively unit-testable.
- **Duplicate detection compares content hashes, scoped to the NVRTC family.** Stronger than a size comparison (cannot false-positive on same-size different libraries) and narrow enough that two benign identical files can never block a release.
- **The version assertion is applied on tag pushes only.** `workflow_dispatch` builds whatever ref it was started from, so pinning the binary to the input tag there would fail by design.
- **The desktop job boots the sidecar `camelid.exe`, not the Tauri GUI.** Launching a GUI headless on a runner proves nothing; the engine it bundles is the thing that must serve. A failure there still cannot block the server release — that job remains in no other job's `needs`.
- **Self-tests are named `test-*.mjs`** so the existing `ci.yml` validation-scripts glob picks them up. **No `ci.yml` change** (deliberate — #507 touches that file).
- **The publish-job re-verify is plain shell** (`shasum` + `awk`), so it needs no `checkout`/`setup-node` added to that job. `awk` rather than `read` so a sidecar with no trailing newline still yields its hash.
- Receipts are uploaded as CI run artifacts only — never attached to the release, never committed (`qa/evidence-bundles` is a scrubbed, checksum-audited tree and these are not one of its bundles).

## Scope

No Rust, no source, no `ci.yml`, no support-claim and no ledger changes.

The smoke is **model-less by design** — CI has no GGUF. It proves boot, health, capability reporting, UI embedding and clean shutdown. It makes **no inference or parity claim**; those stay with the existing exact-row evidence lanes.

## Validation

- Both self-tests pass, covering every failure branch, both sidecar conventions, the layout asymmetry in both directions, and a real loopback HTTP server standing in for the engine.
- Full `ci.yml` validation-scripts glob run locally: 22 pass, 3 fail. All three failures are pre-existing Windows-only path assumptions in untouched scripts (`test-audit-evidence-bundle-privacy.mjs` asserts a POSIX `/SHA256SUMS` separator; `test-bench-llama3-same-host.mjs` expects a `/tmp` path; `test-check-public-evidence-claims.mjs` trips its own private-path scanner on a `C:\Users` temp dir). That job runs on ubuntu in CI, which is why they are green upstream.
- Ran both gates against a **real release build**, staged, zipped, checksummed and extracted exactly as `release.yml` does it: `sha256 ok`, `version: v0.4.4 (tag v0.4.4)`, health + capabilities + embedded UI all served, clean teardown, no orphan processes. That host has an NVIDIA driver but the artifact shipped no NVRTC, so the CPU-fallback path was genuinely exercised.
- Real `tar` round-trip for the nested layout, including symlink preservation.
- Negative cases all exit 1 with precise messages: absent NVRTC under `--require-nvrtc`, corrupted sidecar, wrong label, version mismatch.
- Publish-job shell loop harness-tested: both sidecar conventions, tampered archive, sidecar without archive, archive without sidecar, empty `dist/`.
- Workflow YAML parses and every verify step lands after packaging/signing and before upload/publish.
- `check-public-scrub.sh` exits 0. All new files are LF (`i/lf w/lf`), `git diff --check` clean.

## Known limitations

- **The Windows legs are not dry-runnable on a fork.** `build-windows` and `desktop-windows` use `environment: release` with Azure OIDC signing; a fork lacks those secrets, so only the linux/macos `build` legs can be exercised via `workflow_dispatch`. The Windows path was validated locally instead, against a real staged artifact.
- **Verification runs on the runner that built each artifact**, which is the platform it ships to. Extracting the Linux tarball somewhere symlinks collapse into copies (Windows without developer mode) makes the duplicate-payload guard fire on a correct packaging job; the failure message says so explicitly.
- The smoke asserts nothing about inference, throughput or model behaviour.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -529,11 +529,48 @@ jobs:
     needs: [build, build-windows]
     runs-on: ubuntu-latest
     steps:
-      - name: Download build artifacts
+      # Download BY NAME, one step per artifact this job is responsible for.
+      # `download-artifact` with neither `name` nor `pattern` fetches every
+      # artifact in the run, and `merge-multiple: true` flattens them all into
+      # one directory -- so a bare download would sweep the diagnostic
+      # verification receipts (and whatever the independent desktop job had
+      # finished uploading) into `dist`, and `files: dist/*` below would attach
+      # them to the public release. Naming each artifact keeps the published
+      # set exact and fails closed if one is missing.
+      - name: Download the macOS artifact
         uses: actions/download-artifact@v4
         with:
+          name: camelid-macos-arm64
           path: dist
-          merge-multiple: true
+
+      - name: Download the Linux artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: camelid-linux-x86_64
+          path: dist
+
+      - name: Download the Windows artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: camelid-windows-x64
+          path: dist
+
+      # Nothing but the three server archives and their sidecars may reach the
+      # release. This is a cheap standing assertion that the downloads above
+      # stayed scoped.
+      - name: Confirm only publishable files are staged
+        working-directory: dist
+        run: |
+          set -euo pipefail
+          unexpected=$(find . -maxdepth 1 -type f \
+            ! -name '*.tar.gz' ! -name '*.tar.gz.sha256' \
+            ! -name '*.zip' ! -name '*.zip.sha256' -print)
+          if [ -n "$unexpected" ]; then
+            echo "refusing to publish unexpected files:" >&2
+            echo "$unexpected" >&2
+            exit 1
+          fi
+          ls -1
 
       # Each .sha256 was computed on its own build runner. The bytes then made
       # two more hops (upload-artifact, download-artifact) before this job
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,11 @@ permissions:
   contents: write
 
 env:
+  # The single release identity for this run. A tag push resolves to the pushed
+  # tag; a workflow_dispatch resolves to the tag the operator asked to publish.
+  # Every job checks THIS out and asserts the binary reports it, so a dispatch
+  # cannot build one ref and attach those bytes to a different tag.
+  RELEASE_TAG: ${{ github.event.inputs.tag || github.ref_name }}
   # Any CUDA 12.x works: the NVRTC soname (nvrtc64_120_0.dll) is stable across the
   # 12.x series and cudarc is pinned to the 12.x ABI (`cuda-12090`). The matching
   # nvrtc-builtins ship in the same component, so the pair is always self-consistent.
@@ -37,8 +42,13 @@ jobs:
             label: linux-x86_64
     runs-on: ${{ matrix.os }}
     steps:
+      # Pinned to RELEASE_TAG, not the triggering ref: a workflow_dispatch must
+      # build the tag it is going to publish under, or the attached bytes are
+      # not the tag's bytes.
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          ref: ${{ env.RELEASE_TAG }}
 
       # The web UI is embedded into the binary at compile time, so it must be
       # built before cargo build.
@@ -125,20 +135,14 @@ jobs:
             $nvrtc_flag \
             --out "$dist-check-receipt.json"
 
-          # Only a tag push guarantees the checked-out commit IS the released
-          # commit. workflow_dispatch builds whatever ref it was started from,
-          # so pinning the binary to the input tag there would fail by design.
-          version_flag=""
-          if [ "${{ github.ref_type }}" = "tag" ]; then
-            version_flag="--expect-version ${{ github.ref_name }}"
-          else
-            echo "note: not a tag push; skipping the binary/tag version assertion"
-          fi
-
+          # Asserted unconditionally. The checkout above pinned this job to
+          # RELEASE_TAG, so the binary must report it whether this run came
+          # from a tag push or a manual dispatch -- the release is published
+          # under that same tag either way.
           node scripts/smoke-release-artifact.mjs \
             --dir "verify/$dist" \
             --label "${{ matrix.label }}" \
-            $version_flag \
+            --expect-version "${{ env.RELEASE_TAG }}" \
             --out "$dist-smoke-receipt.json"
 
       - name: Upload artifacts
@@ -178,6 +182,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          ref: ${{ env.RELEASE_TAG }}
 
       - name: Setup Node
         uses: actions/setup-node@v6
@@ -277,11 +283,7 @@ jobs:
           if ($LASTEXITCODE -ne 0) { throw "check-release-artifact failed ($LASTEXITCODE)" }
 
           # Compress-Archive packs the CONTENTS, so the payload is at the root.
-          $versionArgs = @()
-          if ('${{ github.ref_type }}' -eq 'tag') { $versionArgs = @('--expect-version', '${{ github.ref_name }}') }
-          else { Write-Host 'note: not a tag push; skipping the binary/tag version assertion' }
-
-          node scripts/smoke-release-artifact.mjs --dir verify --label windows-x64 @versionArgs --out "$dist-smoke-receipt.json"
+          node scripts/smoke-release-artifact.mjs --dir verify --label windows-x64 --expect-version '${{ env.RELEASE_TAG }}' --out "$dist-smoke-receipt.json"
           if ($LASTEXITCODE -ne 0) { throw "smoke-release-artifact failed ($LASTEXITCODE)" }
 
       - name: Upload artifacts
@@ -331,6 +333,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          ref: ${{ env.RELEASE_TAG }}
 
       - name: Setup Node
         uses: actions/setup-node@v6
@@ -480,11 +484,7 @@ jobs:
             --out "$dist-check-receipt.json"
           if ($LASTEXITCODE -ne 0) { throw "check-release-artifact failed ($LASTEXITCODE)" }
 
-          $versionArgs = @()
-          if ('${{ github.ref_type }}' -eq 'tag') { $versionArgs = @('--expect-version', '${{ github.ref_name }}') }
-          else { Write-Host 'note: not a tag push; skipping the binary/tag version assertion' }
-
-          node scripts/smoke-release-artifact.mjs --dir verify-desktop --label desktop-windows-x64 @versionArgs --out "$dist-smoke-receipt.json"
+          node scripts/smoke-release-artifact.mjs --dir verify-desktop --label desktop-windows-x64 --expect-version '${{ env.RELEASE_TAG }}' --out "$dist-smoke-receipt.json"
           if ($LASTEXITCODE -ne 0) { throw "smoke-release-artifact failed ($LASTEXITCODE)" }
 
           # The installer is published separately and must exist as its own file.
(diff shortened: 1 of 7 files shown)


karan68:
I reproduced each against the committed code before changing anything, and one of them was a regression this branch introduced. Fixed in `f5e687d1`.

## P1 — NVRTC soname could certify a broken link

Confirmed. I ran the exact entry sets through `checkManifest`; **all four returned `[]`**:

| Archive | Before |
|---|---|
| `libnvrtc.so.12 -> libnvrtc-builtins.so.12.9` | passed |
| `libnvrtc.so.12 -> libnvrtc.so.12` (self-referential) | passed |
| `libnvrtc.so.12 -> libnvrtc.so.13 -> libnvrtc.so.12` (cycle) | passed |
| `libnvrtc.so.12 -> LICENSE` | passed |

Your point about the smoke is the one that makes this a P1 rather than a nit: the boot smoke *deliberately* runs driverless to exercise CPU fallback, so it can never catch a wrong-family soname. The gate would have certified precisely the silent CPU-fallback packaging failure it exists to prevent.

Added `resolveSymlinkChain()`, which walks a chain to the regular file it ultimately names and reports `dangling` / `escape` / `cycle` / `too-deep`. Both NVRTC roles now require a candidate to *resolve* to a regular file whose basename still matches that role's pattern, and the failure names what it resolved to:

```
no usable NVRTC compiler library — the GPU path would silently fall back to CPU
on a driver-only host. Checked: libnvrtc.so.12 resolves to libnvrtc-builtins.so.12.9,
which is not an NVRTC compiler library
```

I also applied it to the builtins role and to Windows, since the same substitution is possible in both directions. The general symlink check now detects cycles and over-deep chains, which the old one-hop existence test read as satisfied.

## P2 — receipts were being published

Confirmed, and this one was **my regression**, introduced by the receipt uploads. The step was:

```yaml
- uses: actions/download-artifact@v4
  with:
    path: dist
    merge-multiple: true
```

No `name`, no `pattern` — so every artifact in the run, flattened into `dist`, then `files: dist/*`. My PR text claimed receipts were CI-run-only, which was simply false as written.

The release job now downloads the three server artifacts **by name**, plus a guard step that refuses to publish anything that is not a `.tar.gz`/`.zip` or its `.sha256` sidecar — so the claim is now enforced rather than asserted.

Per your note about the desktop artifact: naming the downloads also means the published set no longer depends on whether the independent `desktop-windows` job happened to finish before `release` ran. That job continues to attach its own signed artifacts exactly as before, so the net published set is unchanged apart from the receipts.

## P3 — smoke could pass after the artifact died

Confirmed on both counts. Nothing rechecked liveness after the probes, so a binary that served the three requests and exited immediately left `terminate()` looking at a populated `exitCode` and returning success. And `terminate()` awaited `exit` rather than `close`, and did not wait at all after escalating to `SIGKILL`, so `assertNoCrashMarkers` could read a still-draining buffer — exactly where a panic or abort lands.

Extracted `bootAndProbe()` so this is testable without a release binary. It now re-probes `/v1/health` after the probe sequence *and* asserts the child is still alive; `terminate()` awaits stdio `close` via a promise created at spawn time (so a fast exit cannot race it), including after `SIGKILL`; stderr is read only afterwards. A mid-probe transport error is now a reported failure rather than an unhandled rejection.

The fixture you asked for found a further bug in my own first attempt: `events.once(child, 'close')` **rejects** on the `error` event a failed spawn emits, so an unspawnable binary crashed the runner instead of failing cleanly. Replaced with a manual close promise, and there is now a regression test for it.

## Tests added

- `resolveSymlinkChain` units: self-reference, two-hop cycle, absolute and relative escape, multi-hop resolution, hop bound.
- Wrong-family soname cases for both roles on Linux and Windows, alongside the valid `soname -> versioned real file` shape so the fix cannot pass by simply rejecting all symlinks.
- A child-process fixture that serves valid health/capabilities/UI responses then exits immediately after the final probe, writing a panic marker with `writeSync` on the way down — asserting both that the death is caught and that the marker still reaches the crash-marker check.
- A never-listening process and an unspawnable binary.
- Harness coverage for the new publish guard, including the exact receipt-leak case.

Re-validated after the refactor: the real release binary still passes the static check and the boot smoke with no orphan processes; workflow YAML parses; public scrub passes; all files LF.

karan68:
All four are real. I reproduced each against the previous head before changing anything, and one of them was a design call I got wrong rather than an oversight. Fixed in `b70a9f4b`.

## 1. Manual dispatch could publish unasserted bytes under any tag

Confirmed, and you're right that this shouldn't exist. I had documented skipping `--expect-version` on dispatch as intentional — because `workflow_dispatch` builds whatever ref it was started from, so asserting the input tag "would fail by design". That reasoning was backwards: the correct fix is to make the build match the tag, not to drop the assertion. As written, a dispatch from any branch with input `vX.Y.Z` built that branch, skipped the assertion, and attached those bytes to `vX.Y.Z`.

There is now a single release identity for the run:

```yaml
env:
  RELEASE_TAG: ${{ github.event.inputs.tag || github.ref_name }}
```

All three build jobs check that ref out (`ref: ${{ env.RELEASE_TAG }}`), both publish steps use it, and `--expect-version` is now asserted **unconditionally**. A dispatch builds the tag it publishes under, and fails closed at checkout if that tag doesn't exist. No conditional version gating remains anywhere in the file.

## 2. The WebUI smoke passed while browsers got a blank app

Confirmed. `assertWebUi(200, htmlHeaders, '<html><head><title>Camelid</title><script src="/assets/missing.js"></script></head></html>')` returned `[]`.

`probeServer` now extracts the hashed module URL and requests it, requiring 200, a JavaScript content type, and a non-empty body. The moment I wired it up, the existing fixture test failed with:

```
GET /assets/index-BwWMwyQo.js: expected 200, got 404 — the app shell references
a bundle the binary does not serve
```

which is the proof the fetch actually happens. Verified against the real release binary as well — its bundle is served with a JavaScript content type and passes.

Note the fixture's death trigger moved with it: the bundle, not `/`, is now the final probe, so the "serves everything then falls over" test still covers the real last request.

## 3. Publish-time checksum verification accepted a mislabeled sidecar

Confirmed. `parseChecksumFile` happily returned `{hash, name: "some-other-file.zip"}`, and separately accepted a valid first record followed by garbage — while the publish gate's `awk '{print $1; exit}'` ignored the embedded name entirely.

Two changes. `parseChecksumFile` now requires **exactly one record** (a `shasum -c` run consumes every record, so honouring only the first verifies something different from what a user's own check does). And I took the "reuse the Node parser" option: `scripts/verify-release-checksums.mjs` runs at publication time, so the build-time and publish-time gates cannot drift and the tolerant-format handling is written once. It subsumes the previous staged-file check too.

Run against a realistic `dist/`:

```
FAIL: refusing to publish unexpected file: camelid-linux-x86_64-check-receipt.json
FAIL: camelid-windows-x64.zip.sha256 names "some-other-file.zip", not
      "camelid-windows-x64.zip" — a user running `shasum -c` would check a different file
FAIL: camelid-windows-x64.zip.sha256: checksum file must contain exactly one record, found 2
```

## 4. The artifact contract was an open set

Confirmed — a valid Windows payload plus `credentials.json`, plus an unexpected `.exe`, and plus a nested directory all returned `[]`.

Each label now carries an allowlist, with constrained NVRTC patterns rather than "contains nvrtc":

```js
const NVRTC_UNIX_PATTERNS = [/^libnvrtc\.so(\.\d+)+$/, /^libnvrtc-builtins\.so(\.\d+)+$/]
const NVRTC_WINDOWS_PATTERNS = [/^nvrtc64_\d+_\d+\.dll$/i, /^nvrtc-builtins64_\d+\.dll$/i]
```

I checked the allowlists against what the packaging steps actually stage (`package-linux-cuda.sh`, `package-windows-cuda.ps1`, and the portable-layout assembly) so this fails closed on a future change rather than on today's artifacts — and there's an explicit test asserting the real staged NVRTC filenames satisfy it. `nvrtc64_evil.dll` and an unversioned `libnvrtc.so` now fail.

## Tests

`extractUiAssetUrl`/`assertUiAsset` units; end-to-end cases for a referenced-but-missing bundle and an HTML body served at the bundle URL; allowlist cases for a credential, a stray executable, a nested directory, an extra document, and nvrtc-shaped names missing the versioned pattern; strict single-record parsing; and a full self-test for the new publish gate (named `test-*.mjs`, so the existing `validation-scripts` glob picks it up with no `ci.yml` change — that job now runs 23 scripts).

Re-validated after the refactor: the real release binary still passes the static check and the boot smoke, workflow YAML parses, public scrub passes, all files LF.

## Remaining known limitation

Unchanged and still worth stating: `build-windows` and `desktop-windows` use `environment: release` with Azure OIDC signing, so a fork cannot dry-run them via `workflow_dispatch` — only the linux/macos `build` legs. The Windows paths were validated locally against a real staged artifact instead. The first real tagged release is where those two legs execute end to end.

PR #515, opened 2026-07-27
fix(phi4): hold exact artifact and match GPT4O template

karan68:
## Summary
- Match the executed llama.cpp `LLAMA_VOCAB_PRE_TYPE_GPT4O` collapsed-regex tokenizer semantics, including ASCII case discrimination, Unicode General Category `L`/`N`, separated marks, newline-slash punctuation, and non-ASCII-uppercase backtracking.
- Pin 19 direct `llama-tokenize --ids --no-bos` oracle captures for the exact Phi-4 GGUF. The pack covers NFD marks, Indic marks, Cyrillic, punctuation/newline, and the non-ASCII-capital backtracking cases (`ÜBER`, `AÉB`, `ÉTAT`, `ÅNGSTROM`).
- Gate `gpt-4o` admission to the exact Phi-4 artifact’s filename, metadata, and SHA-256; renamed/substituted GGUFs fail closed.
- Route the **actual embedded GGUF template** (`<|role|>content<|end|><|assistant|>`) through the compact renderer. This fixes the prior role-colon fallback that made Phi generate synthetic `system:`/`user:` transcript turns.
- Tool-bearing Phi-4 requests fail closed until that template path is independently validated.
- Keep Phi-4 out of `curated_catalog()` and represent it only as `active_validation_blocked_parity`, with a structured HOLD receipt and derived ledger entry.

## HOLD Boundary
This is not a catalog or supported-row promotion. The exact artifact remains held because:

- CUDA raw greedy decode diverged at generated index 3 and is not attributed as a near-tie.
- The CPU parity matrix is inconclusive.
- The forward-path cause remains unisolated.

## Validation
- `cargo check --lib`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `node scripts/check-ledger-schema.mjs`
- `node scripts/check-ledger-drift.mjs`
- `scripts/check-public-scrub.sh`
- reviewer harness: **19/19** GPT4O segmentation cases match the emulated pinned oracle
- exact 2.49 GB GGUF SHA-gated tokenizer vector/EOG regression: passed
- real embedded-template routing pack: passed
- Phi-4 tools fail-closed regression: passed
- live visual check at `127.0.0.1:8200`: prompt begins with `<|user|>` token `200021`, no synthetic role transcript, concise prompt stopped with `finish_reason: stop`


karan68:
Addressed in `4eba404d2`.

1. GPT4O now ports the executed `llama-vocab.cpp:433` collapsed-regex semantics: General Category `L`/`N`, ASCII-only case classes, marks outside word classes, and `\r`/`\n`/`/` punctuation suffix.
2. The old incorrect NFD expectation is replaced by a 12-case pinned oracle pack. The pure tokenizer test consumes its segment expectations.
3. The same pack stores direct `llama-tokenize --ids --no-bos` captures (commit `acd79d603`, exact artifact SHA/size, UTF-8 bytes); the real GGUF regression consumes all vector rows and checks Phi-4 EOG registration.
4. `gpt-4o` is fail-closed unless the GGUF has the known Phi-4 filename/model/architecture tuple; an admission-boundary unit test covers rejection of a renamed artifact.
5. Phi-4 was removed from `curated_catalog()`. It is now an `active_validation_blocked_parity` capability HOLD with `qa/muster/HOLD-phi4-mini-instruct-q4_k_m.json`, no support claim, and a regression asserting the filename cannot re-enter the catalog.
6. The orphaned raw fixture was deleted. The tokenizer oracle and full real-template shape pack are both test-consumed. The full template includes the tool-system and EOS branches; detector direction is tested both ways.

Validation locally: strict fmt/Clippy, ledger schema/drift, public scrub, direct exact-GGUF vector+EOG regression, 12-case tokenizer corpus, full template pack, HOLD contract, plus tokenizer/template regression groups. CI is running on the amended head.

karan68:
Addressed in `3f496c077`.

- Restored true `last_lower_like` backtracking for `UPPER* LOWER+`; the reviewer harness now reports **19 match, 0 diverge** after a clean rebuild.
- Added all seven non-ASCII-capital cases to the committed direct `llama-tokenize` oracle pack, including token IDs and UTF-8 byte pins.
- Corrected the template fixture and detector to the **actual embedded GGUF template**. The tokenizer-config EOS branch was not present in the exact GGUF, which had caused a role-colon fallback. Ordinary Phi-4 requests now route to `<|user|>...<|end|><|assistant|>`; tools fail closed with a dedicated regression.
- Strengthened GPT4O admission to require filename, metadata, and the pinned SHA-256. `unicode-general-category` is exact-pinned at 1.1.0 and documented as Unicode 16.0 data.
- Rewrote the PR title/body to the actual held scope.

Live visual verification against the rebuilt release server: `prompt_token_ids` begins `200021,24912,5485` (`<|user|>` plus prompt), no synthetic role transcript is generated, and `Reply with exactly one word: hello` returns `hello` with `finish_reason: stop`.

Final local gates: strict all-target/all-feature Clippy, fmt, ledger schema/drift, public scrub, exact GGUF vector/EOG fixture, template routing/tool fail-closed tests, and the 19-case harness.

timtoole02:
Thanks for this one — and sorry for the branch rewrite. Here's exactly what happened and why.

We were working the same corner of the tokenizer at the same time. While this PR was open, #518 landed a rewrite of the SPM encode path (reference-order bigram merging, `<|…|>` rstrip semantics) and #519 corrected the Phi-3 evidence once prompt-token parity started passing. Your branch was cut before either, so it went `CONFLICTING` through no fault of its own — I just needed to make the two lines of work fit together.

I've rebased onto `a4555c97` and pushed the result. **You'll want `git fetch && git reset --hard origin/feat/phi4-mini-certified` before touching it again.**

What I resolved, so none of it is a surprise:

**`ledger/camelid-ledger.json`** — conflicted on three of your five commits. It's a derived file, so instead of hand-merging I re-ran `scripts/extract-capabilities-to-ledger.mjs` at each step and let `check-ledger-drift.mjs` prove it matches the code. Your Phi-4 row comes through intact (36 rows).

**`src/api/mod.rs`** — the `phi3` assertions. Worth being clear that these were **not** your code: your branch inherited `all_match=false` / `all_pass=false` from its base, and #519 rewrote the evidence string and its assertions together to `all_match=true`. I took main's versions and kept your new `phi4` block and the `curated_catalog` guard. I diffed the added lines old-vs-new to be sure nothing of yours was dropped — the only delta is two comment lines I added.

The nice part: your GPT4O work and the concurrent SPM rewrite never actually collided in substance. You gated on `TokenizerModel::Gpt2Bpe`, #518 gated on `model_name == "llama"` — orthogonal paths, so the merge was clean where it counted. That's a direct consequence of how tightly you scoped the change.

**And the oracle test now passes against the real artifact.** I pulled `Phi-4-mini-instruct-Q4_K_M.gguf` from `unsloth/Phi-4-mini-instruct-GGUF` at the revision your HOLD receipt names, confirmed sha256 `88c00229…1730a`, and ran it on an M4:

```
test api::gemma4_template_tests::phi4_compact_prompt_preserves_roles_without_marker_adjacent_newlines ... ok
test api::gemma4_template_tests::phi4_tool_requests_fail_closed_instead_of_dropping_tools ... ok
test tokenizer::tests::phi4_mini_gpt4o_tokenizer_matches_pinned_oracle ... ok

test result: ok. 3 passed; 0 failed; 0 ignored
```

That was the one claim CI structurally couldn't reach, and it holds on every case in the pinned pack. The rest of the local run is green too: `fmt`, `clippy --all-targets -D warnings`, `cargo test --all-targets`, ledger drift + schema, public scrub, and 23/23 validation harness self-tests.

The discipline here is what makes this easy to trust — admission gated on an exact SHA-256, the row deliberately kept out of `curated_catalog`, tool requests failing closed, and a HOLD receipt that names its own blockers rather than glossing them. That's the standard, and it made this a pleasure to work through.

Two non-blocking follow-ups, whenever it suits:

- `is_exact_phi4_mini_q4km` hashes the full 2.5 GB on every load of that file (twice along the test path). Fine for a gate, but worth a cached or cheaper-short-circuit form if the row is ever promoted.
- `is_mark` still uses the vendored `mark_ranges.rs.inc` table, while `unicode-general-category` is now in the tree and could supply the `*Mark` categories — the two could converge and drop the vendored table.

Thanks again for the care on this one.

Issue #469, opened 2026-07-17
Can't pull qwen3_4b_q4_k_m


russbroomell:
Fresh install on a MacBook and I see this model in the model browser but can't pull it - tried both in the TUI list and in the CLI

PR #526, opened 2026-07-28
feat(webui): one-click first-run activation

karan68:
## The problem

A fresh install cannot reach a single token.

The desktop spawns `camelid serve --addr 127.0.0.1:<port> --no-open --models-dir <dir beside the exe>` into an empty `models/` folder, so the chat surface the user lands on has nothing to answer with. The documented next step is `camelid pull llama32_3b` -- a 3.42 GB decision a new user has no basis to make. There is prior evidence of people stalling exactly here (upstream #469; there is a regression test for it in `src/catalog.rs`).

## What this does

One card, on the surface the user already lands on: one click, one 610 MB download, then Chat.

**`lib/firstRunActivation.js`** derives both answers from live server state.

* "Fresh install" = engine online, no model loaded, no GGUF on disk. Never a stored "has onboarded" flag -- that is wrong in both directions (it survives a wiped models folder; a new browser profile resurrects the card for an established install), and this is exactly the surface where a stale profile has hidden a first-load bug in this repo before.
* The offer is the **smallest `supported_*` contract row whose fit verdict this host does not refuse**. Deriving it means promoting or demoting a row cannot leave a hard-coded model id behind in onboarding, and an unverified row can never be offered however cheap its download. A test pins that.

**`components/onboarding/FirstRunCard.jsx`** runs download -> land -> check -> load -> warm -> Chat on the existing single-flight settlement coordinator.

It is mounted in the app shell, not inside the chat view, for two reasons that are both real failure modes: an in-flight install must keep its watcher when the user navigates away, and it must stay mounted for the whole activation, because the moment the file lands the host stops looking like a fresh install and unmounting there would abandon the load. It also rejoins a download already running for its row, so a reload cannot orphan one. Cancelling leaves nothing installed and a working retry.

**`lib/modelActivation.js`** lifts the inspect-first load protocol out of `ModelsView`, because it now has two callers and the ORDER is the contract: header-only inspect before anything reads a multi-GB file, authoritative load with `replace: true`, then identity + readiness confirmation. A second hand-written copy is how one of them quietly loses a step. Typed 422s come back as codes, so a permanent refusal (`model_too_large_for_host`) is never offered a "Try again" button and no raw HTTP error reaches the user.

It also **warms the generation path** after the load. `serve` already does exactly this when it boots *with* a model; a model loaded later never got that treatment, so the one-time engine build landed on the user's first message and read as "the model is slow".

## Two defects found on the way

**1. The catalog's exact-row join was missing `catalog_id`.**

`catalog_id` is the key the *server* joins on (`filename_is_supported_exact_row`). The frontend instead reconstructed a row id from the display name, so a row id carrying a token the name does not (`qwen3_0_6b_INSTRUCT_q8_0` vs "Qwen3 0.6B Q8_0") never matched. Eight pinned supported rows -- every Qwen3 Q8_0 row and all four Gemma 4 rows -- rendered as merely experimental in the catalog while the server classified the same artifact as supported.

Measured against the shipped ledger: **9 frontend/server disagreements before, 1 after.** This makes no new claim; it makes the UI agree with the contract it already publishes.

**2. The chat gate contradicted the card -- caught by checking what the user actually sees after the click.**

The card installs a validated row, and the chat surface then rendered *"Experimental -- output is unverified and has no parity guarantee"* over it, status line "Runtime ready, support gated", while the evidence chip beside it said `qwen3_0_6b_instruct_q8_0: supported exact row smoke`.

Observed, not theorised. Same file, same machine:

| how it was loaded | what Chat claimed |
|---|---|
| by the app (`POST /api/models/load`) | "Runtime ready, support gated" + experimental banner |
| auto-loaded at startup | "Local chat ready", no banner |

Nothing about the file differs. The engine's startup auto-load names a model from GGUF metadata ("Qwen3 0.6B Instruct" -- matches the row id); `POST /api/models/load` names it whatever id the caller sent ("Qwen3-0.6B-Q8_0.gguf" -- does not). The support verdict depended on which code path issued the load.

`lib/modelLanes.js` already hit this and already solved it, with the diagnosis written out in a comment: the backend's `lane_class` is "strictly stronger evidence than the frontend's subject-string matching", because `classify_model_lane()` requires BOTH an implemented architecture parsed from the real header AND `filename_is_supported_exact_row()`. The Models page has trusted it ever since. The chat gate was simply never given it. Now it is -- and it cannot unlock chat on its own, because `chatUnlocked` still requires runtime readiness (pinned by tests).

## Rebased onto `main` (a2b70ee)

This branch originally also carried a Rust fix: `auto_select_model` ignored `--models-dir` / `CAMELID_MODELS_DIR`, so a model installed into a non-default models directory was invisible at the next launch. **That fix is dropped** -- #529 landed the same behaviour on `main` independently, in `model_default::effective_default_model`, which is a superset of mine (it also honours an explicitly saved default and filters non-files). Taking `main`'s version rather than merging two implementations of one rule.

**This PR now changes no Rust at all.** The relaunch row in the table below depends on that behaviour, and it still holds -- via `main`'s implementation rather than this branch's.

Note for anyone working from the original brief: the premise that `serve` never auto-loads without `--model` is **wrong**. `src/main.rs` has always done `model.or_else(auto_select_model)`; it just looked only in exe-adjacent and CWD locations, which is why the desktop default worked by coincidence.

## Verification

Live on a real box: release binary, empty models dir, **fresh browser profile**.

| step | result |
|---|---|
| fresh boot | model-less; `/api/models/current` -> 404 |
| card | offers `Qwen3-0.6B-Q8_0.gguf` (`qwen3_0_6b_instruct_q8_0`, contract `supported_exact_row_smoke`, fit `fits_resident`), **0 page errors** |
| one click | `loaded_now` + `generation_ready` in **77.2 s**, **639,446,688 bytes**, sha256 `9465e63a22add5354d9bb4b99e90117043c7124007664907259bd16d043bb031` |
| first message | streams `READY`, warm (not a cold build) |
| cancel mid-download | 0 files on disk, 0 loadable models, 0 active downloads, working "Try again" |
| relaunch | loaded and generation-ready with **0 user actions** |
| after fix #2 | app-loaded file reports `experimental_banner: false`, "Local chat ready", 0 page errors |

### Re-measured on the rebased tree (after the compaction and the retention fix)

Three fresh runs -- empty models dir, fresh Chrome profile each time, the real engine, the real built frontend:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| fresh boot | 0 models, `/current` 404 | same | same |
| activation (click -> `generation_ready`) | 108 s | 77.1 s | 96 s |
| artifact bytes | 639,446,688 | 639,446,688 | 639,446,688 |
| sha256 | `9465e63a...bb031` | identical | identical |
| `lane_class` from the disk scan | `supported` | `supported` | `supported` |
| page errors | 0 | 0 | 0 |
| card height, idle | 214 px | 214 px | 214 px |

Activation spread (77-108 s) is download-bound; the artifact is byte-identical to the original receipt every time, and `lane_class: supported` is the server agreeing with the chip the card shows.

**The warm step, measured properly.** Run 3 timed generation at the wire (in the proxy: request out -> first response byte) rather than by matching rendered text:

| generation request | TTFB |
|---|---|
| the card's warm-up call | **22,594 ms** |
| the user's first message | **261 ms** |

That is what the warm step is worth: the one-time engine build costs ~22.6 s, and the card pays it instead of the user. Without it that cost lands on the first message.

**Cancel and relaunch, re-verified live on the merged tree** (release binary rebuilt from this branch, so the relaunch row exercises `main`'s `model_default::effective_default_model` rather than the fix this branch dropped):

| cancel mid-download | observed |
|---|---|
| bytes on disk when cancel was pressed | 33,898,496 |
| files left in the models dir | **none** (no `.gguf`, no `.part`) |
| engine local scan | 0 models |
| active downloads / download records | 0 / 0 |
| `loaded_now` | false |
| card | still present, offers "Try again", says "Download canceled. Nothing was installed" |
| page errors | 0 |

| relaunch (models dir already holds the activated GGUF) | observed |
|---|---|
| user actions required | **0** |
| time to `generation_ready` | **7.3 s** |
| health | `loaded_now: true`, `generation_ready: true` |
| first-run card | **absent** (slot not even mounted) |
| composer | enabled |
| experimental banner | false, status reads "Local chat ready" |
| page errors | 0 |

The relaunch run also re-confirms finding 2 from the other direction: the startup auto-load names the model from GGUF metadata (`Qwen3 0.6B Instruct`) while the app-load path names it by filename (`Qwen3-0.6B-Q8_0.gguf`) -- the two ids that used to produce two different support verdicts for one file. Both now report `experimental_banner: false` and "Local chat ready".

**A measurement bug I hit and am disclosing rather than reporting:** my first two runs "measured" TTFT by waiting for `READY` to appear on the page -- but the prompt I typed *was* "Reply with exactly: READY", so the match fired on the echo of my own typing and produced an absurd 42-45 ms. The DOM-match number in run 3 (83 ms) is retained in the receipt purely to show it is still wrong. Only the wire timings above are load-bearing. The old **926 ms** first-token figure came from that same flawed method and should not be trusted either; it is struck from the table above.

**Gates:** frontend build; all 13 CI frontend steps -- 12 npm smokes plus the browser component smoke -- on the rebased tree; ledger schema/drift; the public-scrub node gates including the fit-verdict vocabulary check. No Rust is changed, so the Rust gates are `main`'s.

## Review round 2

Both P1s were real and are fixed, with browser-level coverage as requested (`frontend/scripts/first-run-card-smoke.mjs`, 6 scenarios, wired into CI):

* **Retry must not re-download.** `firstRunRetryAction({ artifactInstalled })` decides the retry target by whether the artifact landed, not by which phase failed; a landed file re-runs check/load only.
* **Cancel must be confirmed.** Only `res.ok` counts as a stop. Otherwise the card re-reads `/catalog/downloads` **and** the local scan and routes on what it observes -- and those observations are **tri-state**, so a probe that could not be read stays `null` and never collapses into "nothing was installed". A 409 that finished first routes into activation.

Two further defects I found reviewing my own fix:

* **A cancel racing settlement could activate twice.** The cancel path cleared the single-flight flag before calling `activate()`, so the settlement poll and the cancel could each start one -- two loads and two acks for one file. `activate()` is already single-flight; the reset was the bug.
* **The card deleted its own retry button.** It was mounted on `firstRun || firstRunBusy`, and a *failure* is neither: once the artifact landed, both went false and the card -- error, "Retry setup" and all -- disappeared on the next 2.5 s dashboard refresh, stranding a 610 MB download with no way back. The original scenario clicked retry immediately, so it raced the refresh and passed either way; making it dwell 7 s fails on the old code and passes now.

Both were ablated, not argued.

The inspect-first structural guards in the 3B-closure and UI-regression smokes moved to the module that now owns the protocol, and both surfaces are pinned to route through it rather than re-fork it.

## Scope notes / what a reviewer should push on

* Findings 1 and 2 both widen where the "supported" chip appears -- but only onto rows the **server already classifies as supported** and the Models page already lists under Supported. I believe that is contradiction-removal, not claim-widening. Worth a second opinion, since it is the most consequential judgement here.
* One pre-existing disagreement is deliberately **left alone**: curated `catalog_id: "gemma3_1b_it_q8_0"` vs ledger row id `gemma_3_1b_it_q8_0`. Fixing that typo would move a support classification, which is out of scope for this change.
* The desktop splash was **not** given acquisition progress. It cannot have it: the splash is torn down as soon as `/v1/health` passes and the WebView navigates, and acquisition happens after that, inside the WebUI. The card is the continuity.
* Verification exercised the engine + embedded WebUI (which is exactly what the desktop shell loads), not the Tauri shell itself.
* No engine stderr was added: the desktop keeps a piped, never-drained stderr, so new output there would block the engine.
* No ledger, `COMPATIBILITY.md`, or support wording was touched.

image


frontend/src/components/onboarding/FirstRunCard.jsx
```diff
@@ -0,0 +1,408 @@
...
+              {canceling ? 'Canceling…' : 'Cancel download'}
+            </button>
+          ) : null}
+        </div>
+      ) : (
+        <>
+          {failure ? <p className="cxfirstrun__error" role="alert">{failure.message}</p> : null}
+          <div className="cxfirstrun__actions">
+            {retryable ? (
+              <button type="button" className="cxfirstrun__primary" onClick={startDownload}>
```
timtoole02: [P1] Retry activation instead of reinstalling the catalog artifact

`activate()` runs only after settlement has observed the file on disk. If inspect, load, or readiness then fails, the card clears `settlementInFlightRef` and exposes “Try again”, but this handler always calls `startDownload()`. The backend removes the completed download record and starts curl again; on Windows the final rename onto the existing GGUF can also fail. Preserve that the download landed and retry `activate()` (like `CatalogLaneBrowse`’s `retryAcquisition`) instead of downloading another 610 MB.

frontend/src/components/onboarding/FirstRunCard.jsx
```diff
@@ -0,0 +1,408 @@
...
+    } catch (error) {
+      fail(String(error?.message || error))
+    }
+  }
+
+  const cancelDownload = async () => {
+    if (!itemRef.current || canceling) return
+    setCanceling(true)
+    try {
+      await fetch(`${base}/api/models/catalog/cancel`, {
```
timtoole02: [P1] Do not report cancellation until the backend confirms it

The response is never checked, and the catch path still falls through to `fail('Download canceled…')`. A cancel during `starting` can return 404 before install registers; a cancel just after completion can return 409. In both cases the download or file may still exist, polling stops because the phase becomes `failed`, and the UI says nothing was installed. Treat only a successful cancellation as canceled; otherwise keep polling/reconcile downloads plus the local scan, and route an already-completed artifact into activation.

timtoole02 (changes requested):
Thanks for the thorough implementation and verification. The happy path and new smoke coverage look good, and CI is green, but the first-run card currently loses track of whether the artifact is already installed and whether cancellation was acknowledged. These transitions can produce a duplicate 610 MB download or a false “Nothing was installed” result. Please address the inline comments and add component-level coverage for post-download retry and cancel races before merging.

karan68:
Both confirmed and fixed in 301aca9. Thank you -- the second one in particular could have told a user their 610 MB was gone while it was sitting on disk.

## P1: retry re-downloading the artifact

Correct, and the consequence is worse than the wasted bandwidth. `install` drops the completed record and starts a fresh curl; that download's promote is `std::fs::rename(part_path, dest_path)`, and once the engine holds the existing GGUF the rename fails -- `finalize_download_artifact` then deletes the freshly downloaded `.part` and records `failed`. The retry ends with strictly less than it started with.

The retry target is now decided by whether the artifact landed rather than by which phase failed:

```js
const retryTarget = firstRunRetryAction({ artifactInstalled })
const retryHandler = retryTarget === 'activate' ? retryActivation : startDownload
```

`retryActivation` clears `settlementInFlightRef` and re-enters `activate()`, matching `CatalogLaneBrowse`'s `retryAcquisition`. `artifactInstalled` is state, not just the ref, precisely so the button can branch on it. The label becomes "Retry setup" and the card says the file is already downloaded, so the user knows no second download is coming.

## P1: reporting cancellation the backend never confirmed

Also correct, and I had both halves wrong: the response was unread and the `finally` claimed cancellation even from the catch path. Reading `cancel_catalog_download` gives three outcomes and only one is a stop:

| response | meaning |
|---|---|
| `200` | a running download was removed |
| `409 download_already_completed` | it finished first and **kept its file** -- cancel is not delete |
| `404` | no such download, which during `starting` just means install has not registered yet |

Three changes:

1. **The install request is awaited before cancelling.** That removes the `starting` race at the source rather than compensating for it -- previously the 404 was for a download that was about to exist.
2. **Only `res.ok` counts as a confirmed stop.**
3. **The outcome is decided by re-reading downloads plus the local scan**, and the observations are deliberately tri-state. A probe that could not be read stays `null` and never collapses into "no", so a failed probe cannot produce a false all-clear:

```
artifactInstalled === true -> activate  (the file is real; finish the job)
stillDownloading  === true -> resume    (cancel did not take; keep watching)
either unknown             -> resume    (we did not look, so we do not claim)
both observed false        -> canceled
```

The wording distinguishes a confirmed stop from an inferred one, and the polling keeps running on `resume` instead of freezing on `failed`.

## Component-level coverage

`frontend/scripts/first-run-card-smoke.mjs` mounts the real card in a browser against a scripted backend, so each branch is deterministic -- the test decides exactly what `/catalog/cancel` answers and exactly when the file appears in the scan. Five scenarios:

- a failed activation after the file landed retries the load and issues **no** second install
- a 409 cancel over a completed file activates it instead of reporting a false cancel, and still never re-downloads
- a 404 cancel on a download that is in fact still running resumes, and the download can still complete afterwards
- a cancel whose probes are unreadable refuses to claim anything
- a confirmed cancel reports nothing installed, and its retry *does* install

The completion race is driven from inside the cancel request handler, because mutating the stub before the click let the card notice the file and unmount the button first -- the click has to land while Cancel still exists.

Both fixes were ablated:

- reverting the retry branch -> `retrying a landed artifact must NOT issue a second install (that would refetch 610 MB)`
- restoring the unconditional cancel -> `a 409 cancel over a completed file must not report a failure`

I also moved the install-count assertion to fire at click time rather than after waiting for completion: with the fix reverted the original version only ever failed as a 30 s timeout, which hides which of the two paths ran.

The pure decisions (`firstRunRetryAction`, `firstRunCancelOutcome`) live in `lib/firstRunActivation.js` with unit coverage, and `ui-regression-smoke.mjs` pins both behaviours structurally -- including `assert.doesNotMatch(..., /finally \{[\s\S]{0,200}fail\('Download canceled/)` so the exact shape you flagged cannot come back.

## Validation

14 frontend smokes plus the build. The full live receipt was re-run against a rebuilt release binary: fresh profile, empty models dir, one click, 74.1 s, 639,446,688 bytes, sha256 `9465e63a22add5354d9bb4b99e90117043c7124007664907259bd16d043bb031`, first message streams `READY` at 720 ms, relaunch loaded with zero clicks, zero page errors. No Rust changed.

karan68:
Follow-up in d626541 -- one bug I introduced in the previous commit, and a UX correction.

## Activate exactly once (self-review)

My cancel fix did this before calling `activate()`:

```js
settlementInFlightRef.current = false
activate()
```

which defeats the single-flight guard I was relying on everywhere else. The settlement poll can claim the activation in the same window the cancel is being answered, so forcing the flag starts a **second concurrent activation** -- two loads and two acknowledgements for one file. `activate()` is already single-flight, so the correct code is just to call it: if settlement owns it the call is a no-op, which is right, because the activation is happening either way.

Reproduced deterministically rather than argued: the stub lands the artifact the instant `/catalog/cancel` arrives but holds the response open for 2.5 s, so the poll claims the activation first and the cancel's reconciliation arrives second. The scenario asserts exactly one load, one ack, one install. Ablated -- restoring the flag reset fails with `the model must be loaded exactly once`.

That is 6 component scenarios now.

## The card was too big

It was ~430 px of a 768 px viewport and pushed the hero and composer off screen. Worse, the same instruction appeared **four times** on one screen: the card, the chat hero, the composer detail line, and the composer hint.

Restructured into a two-column row -- what this is on the left, what to press on the right -- with provenance and the download caveats moved into a collapsed "What gets downloaded" disclosure. Present for anyone who stops to ask; one line when they do not. The progress block still expands in place while working.

Measured at 1069x768:

| state | before | after |
|---|---|---|
| idle | ~430 px | **214 px** |
| working | ~430 px | 257 px |
| failed | ~430 px | 205 px |
| failed, disclosure open | -- | 298 px |

Hero and composer are on screen in every state, and it stacks at 420 px. The duplicate instructions are gone: the composer detail line defers to the card during first run, the placeholder reads "Set up the model above, then chat here", and the redundant hint line is suppressed (the send button keeps its own tooltip).

No behaviour changed in this part -- it is render + CSS, and all six component scenarios still pass against it, including the cancel button which stays inside the progress block.

14 frontend smokes plus the build, all green.

timtoole02 (approved):
Reviewed the updated head and the current-main integration. The original retry/cancel findings are resolved, the conflict with the new embedding sidecar path is preserved correctly, encoder-only rows are excluded from first-run Chat recommendations, and the merged frontend smoke suite passes locally.

PR #540, opened 2026-07-29
feat(diagnostics): record panics and session lifecycle to a bounded on-disk journal

karan68:
## Why

A packaged install that panics leaves **no trace on the machine it happened on**. The console window is the only sink, so closing it — or losing it with the process — destroys the evidence. A user whose engine stopped has no way to answer "why" on their own hardware, which is the wrong answer for a local-first product.

## What

`src/diagnostics.rs` appends one JSON object per line to `%LOCALAPPDATA%\Camelid\logs\camelid.log` (XDG-ish elsewhere), mirroring the directory convention already used by the gait store.

- **The panic hook is installed AFTER `quiet_cudarc_loader_panics`**, so it is the *outer* hook: every panic is recorded first, and the existing filter still decides what reaches the console. cudarc loader panics become suppressed-but-recorded rather than silenced — they fire on exactly the GPU boxes nobody can inspect.
- `serve` brackets the run with `session_start` / `session_exit`, so a panic or an exit failure carries run context (version, address, platform).
- The path is printed at startup, and README documents how to read the file.

The module is self-contained: the full contract is stated in its own docs, with no external document to keep in sync.

## A `panic` record does not mean the process died

`src/cuda.rs` probes the GPU with **five** `catch_unwind` calls around `CudaContext::new` and PTX compilation, and `api/engine.rs` wraps a decode job's body so a panic becomes a failed request rather than a dead process. The shipped Windows build has CUDA compiled in, so on a machine with no CUDA runtime those probes panic on **every startup**, are handled, and the engine goes on to run happily on the CPU.

Unmarked, they would be the *majority* of records on exactly the population this journal exists for, burying the real crash under routine noise. Records therefore carry `"expected": true` when the panic came from the cudarc loader, detected by the same origin test `quiet_cudarc_loader_panics` already uses to keep them off the console. `"expected": false` is the interesting kind.

## What the missing-exit signal does and does not mean

`serve` installs **no signal handler** — there is no `ctrl_c` or `with_graceful_shutdown` anywhere in the engine — so an ordinary Ctrl-C ends the process exactly as abruptly as an out-of-memory kill, and neither produces a `session_exit`.

A missing exit record therefore means **"did not fail on the way out"**, *not* "was killed".

Telling a deliberate stop apart from a kill needs a graceful-shutdown path in `serve`, deliberately **not** built here: owning Ctrl-C would put a diagnostics feature on the critical path of stopping the server, and diagnostics must never be able to break what they diagnose.

## Contract

- **The engine owns the file.** Nothing new goes to stderr for the desktop to scrape: `camelid-desktop` retains a piped stderr it drains only on startup failure, so a chatty stderr would fill the OS pipe buffer and block the engine. The one deliberate exception is a single line at startup naming the journal path, next to the existing ready banner — without it the file is undiscoverable, and one bounded line per run cannot fill a pipe buffer. Documented as an exception in the module docs rather than left as a contradiction.
- **Fails open.** Every path degrades to "no journal", never to a failed startup or generation.
- **Bounded three ways**: a 1 MiB cap with exactly one retained predecessor, a clamp on messages (4 KiB) and backtraces (16 KiB), and a single write call per record.
- **No prompt or generated text.** Process facts and panic payloads only.
- **No new dependencies. No phone-home.**

## Defects found and fixed while reviewing my own first draft

| Defect | Consequence |
|---|---|
| Record written as two calls (line, then newline) | Camelid can run as several processes at once and an append is atomic only *per call*, so two writers could interleave around the newline and corrupt the JSONL |
| A failed rotation aborted the append | Dropped the record at exactly the moment worth recording |
| Empty `LOCALAPPDATA` accepted | Resolved to the **relative** path `Camelid/logs`, scattering journals through whatever directory the app started in |
| Panic messages and backtraces unbounded | One pathological record could consume the whole budget |
| Comment claimed `flush()` provided durability | `File` has no userspace buffer, so it is a no-op. What actually makes a record survive is `write_all` reaching the OS before returning; the page cache outlives the process. This is **not** protection against power loss, and the comment now says so |
| Hook test parsed the whole file as one JSON value | The hook is process-global once installed, so any concurrently panicking test (`should_panic` cases, the `catch_unwind` probes in catalog/cuda/diffusion, which the ubuntu `--all-features` leg exercises) also appends there. The test now selects its own record by message and asserts there is exactly one |

## Validation

- **10 unit tests**: directory fallback chain including empty values, UTF-8-safe clamping, hostile text that must not escape its line, a failed rotation, a write failure, concurrent writers, and an installed-hook test that panics for real and reads the record back off disk while asserting the previous hook still runs and the record is classified `expected: false`.
- **Full library suite 1265 passed / 0 failed** on this branch; bin target 23 passed / 0 failed.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.
- **Ablation**: with the single-write fix removed, the concurrency test fails **8 of 8 runs**, with two records glued together (`...}{"ts":...`).
- **Live receipts on a built binary** for all three signatures:

```
{"ts":"...","event":"session_start","pid":29536,"version":"v0.4.5-dirty","addr":"127.0.0.1:8199","os":"windows","arch":"x86_64"}
{"ts":"...","event":"panic","pid":29536,"message":"...","thread":"main","location":"src\\main.rs:1758:17"}
```
```
{"ts":"...","event":"session_exit","pid":31252,"outcome":"error","detail":"Only one usage of each socket address ... (os error 10048)"}
```
plus a hard kill, which leaves `session_start` with no exit record.

### This feature immediately earned its keep

While testing the sibling WebUI change (#541), the banner offered a restart command that failed with `os error 10048`. The journal explained it without any guesswork: a `session_start` on `127.0.0.1:8181` while the live engine was on `127.0.0.1:8188` — the command had silently fallen back to the default port. That is exactly the class of question this file exists to answer, and it produced the fix in #541.

## Known-flaky neighbour, not from this change

`gait::tests::memory_measurement_is_sane` asserts plausible memory bandwidth and can fail under the load of the full parallel suite on a memory-constrained box. Verified as pre-existing rather than assumed: the same suite on a clean tree with these changes stashed fails identically, and the test passes in isolation. Flagging it so a reviewer who hits it does not attribute it here.

## Scope boundaries

- Not a tracing file sink, not a `doctor` command, not a support bundle.
- Records are written for `serve` only; one-shot CLI invocations stay silent, but panics anywhere are still recorded.
- Nothing inside the process can record an OS-level kill (out-of-memory, task-kill). The journal shows the gap; the Windows Event Log is the source for that case.


src/main.rs
```diff
@@ -1744,7 +1749,16 @@ async fn main() -> anyhow::Result<()> {
             let model = model.or_else(|| auto_select_model(models_dir.as_deref()));
             // Open the browser only when run interactively and not opted out.
             let open_ui = !no_open && std::io::IsTerminal::is_terminal(&std::io::stdout());
-            api::serve(
+            // Journal the run. A `session_start` with no matching `session_exit`
+            // is the signature of an external kill — out of memory, a task-kill,
+            // or the console window being closed — which is otherwise the one
+            // failure a user has no way at all to explain after the fact.
```
timtoole02: This is the surviving copy of the overclaim the PR description says was corrected in the first draft. The module docs and the README both say the opposite — a missing `session_exit` must be read as "did not fail on the way out", never as evidence of a kill, because an ordinary Ctrl-C is exactly as unrecorded as an OOM kill. As written, this comment teaches the next reader of `main.rs` the wrong semantics.

```suggestion
            // Journal the run. A `session_start` with no matching `session_exit`
            // means the process did not leave through the failure path. That
            // includes an ordinary Ctrl-C as well as an external kill — `serve`
            // installs no signal handler — so read it as "did not fail on the
            // way out", never as proof of a kill. See src/diagnostics.rs.
```

timtoole02 (changes requested):
Strong module — bounded, fail-open, and precise about what a record does and does not mean. I verified the load-bearing claims against the tree rather than taking the description's word:

- The `from_cudarc` origin test is byte-identical to the one `quiet_cudarc_loader_panics` uses, and the install order does make the recorder the outer hook — suppressed loader panics become recorded-but-quiet exactly as described.
- The bounded/fail-open mechanics hold up: one write call per record (with the concurrency test that would catch an interleave), a failed rotation never drops the record, empty env vars can't resolve to a relative journal path, and rename-over-existing is handled for Windows.
- The 10-test count is accurate, and select-by-message in the installed-hook test is the right defense for a process-global hook under a parallel suite.
- The `gait::memory_measurement_is_sane` note matches the known state of that test; correctly not attributed here.

One change requested, inline on `src/main.rs`: a comment that still carries the overclaim the rest of this PR was careful to correct.

Two non-blocking notes:

- **DIAGNOSTICS_AND_DOCTOR_PLAN.md** — well-researched, and the claims I spot-checked hold up. But the strategy half (verdicts, slice ranking, kill criteria) speaks in the project's voice, while the code only depends on §6–§7. Fine to land as proposed; I may trim or relocate the strategy sections afterwards as a maintainer edit.
- The startup `eprintln!` is the one exception to §7.1's "nothing new goes to stderr." A single bounded line per run can't fill the pipe buffer, so no change needed — but worth a sentence in the module docs acknowledging the exception, or moving it next to the existing startup banner.

Trivial: the installed-hook test's fixed directory under `temp_dir()` could collide if two test processes ever ran concurrently on one machine; a pid suffix would make it airtight.

karan68:
Thanks — the requested change is in, and all three notes are addressed at `2391902f0`.

**The overclaim in `src/main.rs`.** You're right, and it's the one that mattered. I corrected that claim in the module docs, the README and the PR description, then left it standing in the one place a developer reading `main.rs` would actually meet it. Applied your wording verbatim. The lesson I'm taking: the fix was textual, so I should have grepped for the assertion instead of correcting it only where I remembered writing it.

**`DIAGNOSTICS_AND_DOCTOR_PLAN.md` — removed from the PR entirely.** You were willing to land it and trim the strategy half afterwards; I'd rather not hand you that edit. Working notes that speak in the project's voice shouldn't arrive through a code PR, so it's now kept local and out of the repo.

That had a consequence worth flagging: the module cited it eight times (`per the plan`, `§7.1`, `§7.2`, …), and deleting the file would have left every one of those dangling. The contract is now written out in the module's own docs, so `src/diagnostics.rs` is self-contained and `git grep DIAGNOSTICS_AND_DOCTOR_PLAN` returns nothing. Net effect on the PR: 5 files down to 4, −338 lines.

**The startup `eprintln!`.** You said no change needed, but you were pointing at a real contradiction — the module claimed nothing extra goes to stderr while `serve` prints a line. Documented as a deliberate, bounded exception with its justification (without it the journal is undiscoverable; one line per run cannot fill the desktop's pipe buffer) rather than leaving doc and code disagreeing.

**Temp directory collision.** Suffixed with the pid.

No behaviour change in this commit — comments, module docs, a test directory name, and one file deletion. `fmt`, `clippy --all-targets -- -D warnings` and `check-public-scrub` all exit 0; diagnostics tests 10 passed / 0 failed.

PR #542, opened 2026-07-29
fix(desktop): allow the models-directory commands through the ACL for the loopback UI

timtoole02:
## Bug

On the **Downloaded models** page, **Choose folder…** (and **Use default folder**) failed with:

> Command choose_models_directory not allowed by ACL

Reported on macOS with the shipped v0.4.5 DMG; the same defect exists on Windows — the commands were never invokable from any bundled build.

## Root cause

The embedded web UI runs at the loopback sidecar origin, which is a **remote** origin to Tauri v2's ACL. A remote origin may only invoke commands that a capability explicitly grants. `capabilities/sidecar.json` granted only `core:default`, and the app's own commands were never declared in the build manifest — so no `allow-*` permissions existed to grant at all. (Local webviews may call app commands implicitly, which is why the splash's `startup_snapshot` worked and this went unnoticed.)

## Fix

- `build.rs`: declare the three app commands via `tauri_build::AppManifest`, so tauri-build generates their `allow-*`/`deny-*` ACL permissions.
- `capabilities/sidecar.json`: grant `allow-choose-models-directory` + `allow-reset-models-directory` to the loopback origin — nothing broader; the folder picker stays a Rust-side scoped dialog.
- `capabilities/default.json`: pin `allow-startup-snapshot` explicitly for the local splash.
- Gitignore the autogenerated `permissions/` output (regenerated every build, like `gen/`).

tauri-build validates that capability permission identifiers resolve, so a future rename fails the build rather than shipping another silent ACL denial.

## Verified

- Generated ACL manifest now contains all three command permissions; both capabilities resolve them (checked `gen/schemas`).
- Built and installed the fixed bundle on an Apple Silicon Mac: **Choose folder…** now opens the native scoped folder picker; the splash and startup flow unchanged.


timtoole02:
Follow-up commit: the ACL fix exposed a second latent bug — Tauri runs sync commands on the main thread, and `blocking_pick_folder` must not block the main thread (the panel's modal session runs there). Result was a frozen picker + beachball. Both models-directory commands are now `async` (worker thread), and `build.rs` is reformatted for the CI Format check. Verified on macOS: with the picker open the main thread samples inside the normal AppKit event loop, and Cancel dismisses cleanly.

PR #558, opened 2026-07-30
Size the decode pool to the performance cores on hybrid Windows CPUs

karan68:
## What changed

On Windows the dedicated decode pool is sized from `GetLogicalProcessorInformation`, which reports how many physical cores exist but not what kind they are. On an Intel hybrid part that count mixes performance and efficiency cores, so the pool is sized for cores that decode far slower. This reads `EfficiencyClass` from `GetLogicalProcessorInformationEx` and, on a host that reports more than one class, sizes the decode pool to the performance-core count instead. Hosts reporting a single class keep their existing width exactly.

Two files, no new dependencies (`windows-sys 0.59` already enables the needed features), no new docs. It changes a thread count and nothing else — no placement, no pinning, no kernel changes.

> **This PR was rewritten after review.** The first revision also *pinned* the decode workers to the performance cores and claimed the pin was most of the win. Measured properly, that was wrong: the entire benefit is the width, and the hard affinity mask is significantly **worse** than not pinning. All the pinning code is gone and the numbers below are re-measured. Details in "What the first revision got wrong".

## The defect

The current policy comes from `docs/perf-deep-dive/PERF_RECEIPTS/same-host/p1-cpu-thread-sweep-llama3b-20260621T2150Z.json`, whose `host` field reads **"i7-11800H 8C/16T"** — a non-hybrid part where every core really is identical, so "physical core count" was the right width. On every Intel client CPU since 12th gen it is not. On an i9-14900HX `windows_physical_core_count()` returns **24**, which is 8 performance cores plus 16 efficiency cores, and the decode pool is built 24 workers wide.

Batch-1 decode is a memory-bound matvec that fork-joins across the pool every token, so the step waits on its slowest worker. Sixteen of those 24 workers are on efficiency cores.

Both APIs enumerate the same 24 cores in the same order — verified, **0 mismatches, identical masks**:

```
idx  0-7   EfficiencyClass 1  (performance, SMT)   masks 0x3, 0xC, 0x30, ..., 0xC000
idx  8-23  EfficiencyClass 0  (efficiency)         masks 0x10000 ... 0x80000000
```

## Results

Intel i9-14900HX (8P+16E), Windows, AC power, `Llama-3.2-1B-Instruct-Q8_0`, requests over `/v1/chat/completions` at `temperature=0`.

**The sampling unit is one server process**, because the pool is built once per process. Arms are randomized within each rep and compared **paired**, so machine drift cannot favour one arm. `CAMELID_CUDA_RESIDENT_DECODE=0` throughout: resident CUDA decode is on by default when a device is present and would otherwise decide the result instead of the thread policy.

n=6 paired starts per arm:

| arm | decode tok/s | sd | wall clock | TTFT | CPU seconds |
|---|---:|---:|---:|---:|---:|
| `CAMELID_WIN_HYBRID_POLICY=off` (= main, 24 workers) | 31.754 | 0.378 | 2.096 s | 0.081 s | 52.6 |
| **this PR (8 workers)** | **37.050** | 0.245 | **1.805 s** | 0.078 s | **23.5** |

| metric | this PR − main | 95% CI | verdict |
|---|---:|---|---|
| decode tok/s | +5.296 (+16.7%) | [+4.992, +5.600] | **significant, won 6/6** |
| CPU seconds for the same answer | −29.110 (−55.3%) | [−31.093, −27.127] | **significant, 0/6** |
| time to first token | −0.003 s | — | no detectable difference |

Prefill is untouched and keeps its own wider pool, which is why TTFT does not move.

### The width wants to be the performance-core count, not just "less than physical"

The obvious objection is that any width below 24 would do and `EfficiencyClass` is over-engineering. Sweeping the same lever says otherwise — the curve peaks at 8 and falls off on **both** sides. Same harness, n=6 paired starts per arm:

| decode workers | tok/s | vs main (24) | 95% CI | wins |
|---:|---:|---:|---|---:|
| 4 | 29.645 | **−2.976** | [−3.794, −2.158] | 0/6 |
| **8 (= performance cores)** | **37.536** | **+4.915** | [+4.006, +5.823] | 6/6 |
| 12 | 36.285 | +3.664 | [+2.842, +4.486] | 6/6 |
| 16 | 34.579 | +1.958 | [+0.706, +3.210] | 6/6 |
| 24 (main) | 32.621 | — | — | — |

Against 8 workers directly: 4 scores −7.891, 12 scores −1.251 (CI [−1.488, −1.013]) and 16 scores −2.957 (CI [−4.650, −1.263]) — **every one significant, every one 0/6**. Four workers is worse than shipping nothing. The optimum sits exactly at the boundary the `EfficiencyClass` query reports.

### Pinning does not help, and a hard mask hurts

Measured on the same binary and the same starts, against the identically-sized pool left unpinned:

| placement on top of the 8-wide pool | tok/s | vs unpinned | 95% CI | verdict |
|---|---:|---:|---|---|
| none (this PR) | 37.050 | — | — | — |
| `SetThreadIdealProcessor` hint | 36.504 | −0.546 | [−1.149, +0.057] | not significant |
| `SetThreadAffinityMask` | 35.364 | −1.686 | [−2.267, −1.105] | **significant, 0/6** |

Once the pool stops oversubscribing, the scheduler keeps the busy threads on performance cores by itself; a mask only stops it recovering when something else wants the core. So this PR ships no placement, and `CAMELID_WIN_PIN` is left byte-identical to `main`.

### Output is unchanged, measured not argued

**1 distinct SHA-256 across all 54 measured requests** spanning every arm and every width. Decode parallelises over independent output rows with serial per-row reductions, so pool width cannot change the arithmetic.

### GPU hosts: no change

With the resident CUDA engine driving decode, the decode pool width is irrelevant. An earlier n=8 arm at this width measured 139.843 tok/s against a 141.201 default — a −1.0% difference whose CI crosses zero.

## Safety

- **Single-class hosts are untouched.** `performance_core_masks` returns `None` when every core reports the same `EfficiencyClass`, so pre-hybrid Intel, all AMD, VMs and servers fall through to the existing `windows_physical_core_count()`. It is a pure function with unit tests, so the guarantee is checked on Linux and macOS CI without hybrid hardware.
- **Kill switch**: `CAMELID_WIN_HYBRID_POLICY=0|false|off|no` restores the previous width. Verified on the built binary — `decode_threads` returns to 24 and the sizing log disappears. An unrecognised value leaves a default-on policy on rather than silently disabling it; the parser is a pure function with a unit test.
- **Process affinity is intersected in**, so a job-object or `start /affinity` restricted process is never sized for cores it cannot run on, and a restriction leaving no performance core disables the policy instead of sizing to nothing.
- **Fails open**: detection failure or an empty selection leaves the pre-change width.
- `CAMELID_DECODE_THREADS` and `CAMELID_THREADS` keep their existing precedence — an explicit width still wins, and an operator-pinned global still disables the default policy.
- **`CAMELID_WIN_PIN` is not touched.** Its `GetLogicalProcessorInformation` call is left exactly as-is on purpose: that API reports the *current* processor group, while the new `...Ex` walk filters to group 0, so reusing one for the other would change behaviour on a multi-group host for no measured benefit.
- The Win32 record walk uses locally mirrored `repr(C)` structs with compile-time size assertions (`GROUP_AFFINITY` 16, `PROCESSOR_RELATIONSHIP` 40, record 48) and validates `GroupCount` before indexing `GroupMask[0]`.

### Startup log on the built binary

```
default                               hybrid CPU detected: sizing the decode pool to the
                                      performance cores performance_cores=8 physical_cores=24
                                      decode scheduling lane: ... decode_threads=8 global_threads=24

CAMELID_WIN_HYBRID_POLICY=off         decode scheduling lane: ... decode_threads=24 global_threads=24
                                      (no sizing log)
```

## Why this is default-on

The repo's convention is that new levers ship off until they have receipts, and this is deliberately not a new lever: the decode pool is **already** sized by a promoted default. That default was derived on a non-hybrid part and is wrong on hybrid silicon, so leaving the correction opt-in means the shipped default stays wrong for every Intel client CPU since 12th gen. Non-hybrid hosts provably take the `None` path, output is byte-identical, and one environment variable restores the old behaviour exactly.

## What the first revision got wrong

Worth stating plainly, because it changes how much weight the mechanism story deserves:

1. **It pinned, and claimed the pin was the win.** The re-measurement above shows the width is the whole effect and a hard mask is significantly negative. All placement code is removed.
2. **It cited a test-suite slowdown as evidence against hard affinity.** That comparison was time-confounded — a back-to-back control with the policy disabled (i.e. `main`'s behaviour) reproduced the same slowdown. The claim is withdrawn.
3. **Its confidence intervals used one degree of freedom too few** (an off-by-one in the analysis script's t lookup). Every interval here is recomputed.

## Validation

- `cargo fmt --all -- --check` — exit 0
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --lib` — **1329 passed, 2 failed, 70 ignored**
- `cargo build --release --bin camelid`, startup log verified both ways
- 54 measured server starts across two independent experiments, output hashed on every one

The 2 failures are **pre-existing on `main` and unrelated to this branch**, verified by checking out `bc74fcaf3` (the current main this branch is rebased on) clean and re-running just those two:

```
test api::tests::cooperative_stream_job_seeds_from_the_prompt_prefix_cache ... FAILED
test api::engine::tests::slot_occupancy_tracks_cooperative_and_exclusive_work ... FAILED
test result: FAILED. 0 passed; 2 failed; 1392 filtered out
```

Both are in the cooperative-streaming / prompt-prefix-cache code from #557; this branch touches neither file.

New tests: a single efficiency class disables the policy; an i9-14900HX-shaped topology selects exactly the 8 performance masks; more than two classes keeps only the fastest; process affinity narrows the selection; an affinity with no performance core disables the policy; malformed input is rejected; the kill switch is off only on an explicit value.

## Limitations

- One host, one CPU vendor. The non-hybrid no-op is proven as a code property by unit test, not measured on non-hybrid hardware.
- **Free memory dominates every number on this host.** The identical 16-worker arm measured 11.73 tok/s at 2.6 GB free and 33.13 tok/s at 7.25 GB free. Every figure above was taken at 7.25–7.84 GB free with a 1.77 GB server; an earlier pass under ~3 GB free showed the same ordering and the same conclusions but inflated the effect to +40%. Paired randomized arms are what make the comparisons survive this; the absolute values do not travel.
- One further test, `gait::tests::memory_measurement_is_sane`, is a single-threaded STREAM-triad probe with a 0.5 GB/s floor that fails under load in a debug build. It fails identically with this change disabled, so it is not attributable here, but it will flake on a busy machine.
- 8 is the peak among {4, 8, 12, 16, 24} on this host. Whether the performance-core count is the optimum on a differently shaped hybrid part (say 2P+8E) is not measured; the per-host optimum stays deferred to GAIT, and this PR only replaces a wrong default with a better-grounded one.
- Clippy was run `--all-targets` but not `--all-features`; the change is `#[cfg]`-gated to `windows/x86_64`.
- Unrelated, left alone: `--threads` is declared `#[arg(long, env = "CAMELID_THREADS")]`, so passing the flag on the command line does not set the environment variable, and the pools' "operator pinned the global" guards therefore do not fire for the documented flag.


timtoole02:
This is good stuff!!

timtoole02 (changes requested):
**Verdict: request changes — one substantive ask, plus two convention items. The core change is right and I want it in.**

This is a well-built PR. The mechanism is sound, the scope is tight (one input to one pool — prefill, the global rayon pool and `CAMELID_WIN_PIN` are all provably untouched), and the evidence discipline is better than most perf work that lands here: paired randomized arms with one process per sample, output hashed on every one of the 54 runs, and a width sweep that actually justifies `EfficiencyClass` over "just pick something smaller." The self-retraction of the first revision's pinning claim — including the time-confounded test-suite comparison and the t-lookup off-by-one — is exactly the right instinct, and it's the reason I'm giving the remaining numbers weight.

### What I checked and cleared

- **Struct mirrors.** `GROUP_AFFINITY` 16, `PROCESSOR_RELATIONSHIP` 24 + 16 = 40, `SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX` 8 + 40 = 48. The `const {}` assertions are correct and the `rust (windows-latest)` leg compiles them, so the layout is machine-checked rather than argued.
- **`max()` is the right end of the scale.** Higher `EfficiencyClass` is documented as intrinsically greater performance, and nonzero only on heterogeneous hosts — so `all == top → None` is exactly the non-hybrid no-op you claim.
- **Record walk is bounded.** `offset + record_size <= total`, the `size < record_size` break, and the `group_count >= 1` check before `group_mask[0]` are all correct. The `Vec<u64>` backing for alignment is a nice touch.
- **The fallback can only narrow.** The selection is a subset of group-0 cores, so `hybrid_decode_core_count()` is always ≤ `windows_physical_core_count()`. No path widens the pool.
- **Precedence is preserved.** `CAMELID_DECODE_THREADS` short-circuits, `CAMELID_DECODE_POOL_DEDICATED` returns `global` before this is consulted, `CAMELID_THREADS` forces `None`, and the `physical <= global` guard still holds.
- Ran the seven new unit tests on macOS: 7 passed, 0 failed. CI is green on all three legs.

---

### 1. The AMD no-op is asserted, not true — and it's the one that can regress a shipping machine

`src/inference/win_pin.rs:380`

```rust
// Pre-hybrid Intel, every AMD part, VMs and servers report one class.
```

This is the load-bearing safety claim of the PR — it's what makes default-on defensible — and I don't think it holds. Windows reports a heterogeneous `EfficiencyClass` on AMD in at least two shipping cases:

- **Zen 5c compact cores.** A Ryzen AI 9 HX 370 is 4× Zen 5 + 8× Zen 5c. Zen 5c is the same microarchitecture at a lower clock with less L3 — not a slow E core. This policy sizes decode to **4 workers out of 12 physical cores** on that part.
- **Multi-CCD desktop Ryzen.** Windows treats the differing CPPC preferred-performance values between CCDs as differing efficiency classes. On a 7950X/7950X3D that's 8 of 16.

What makes this more than theoretical is your own sweep: **4 workers scored −2.976 tok/s against the shipped default** — worse than shipping nothing. So under-selection isn't a neutral direction, it's a measured-negative one, and there is currently no floor or sanity clamp on the returned count. The entire evidence base is one Intel hybrid part, and the change is default-on for every Windows x86-64 host.

The cheapest fix that keeps the whole win and bounds the risk is to gate the policy on `GenuineIntel` via `__cpuid` leaf 0 — no new dependencies, a few lines, and it's honest about what was actually measured. AMD then keeps its existing width until someone with the hardware measures it, and the gate can come off with a receipt. A numeric floor would also work, but any threshold is arbitrary and I'd rather not invent one.

Either way, the test comment should stop making the AMD claim. `a_single_efficiency_class_disables_the_policy` is a good test of the code property; it just shouldn't carry a hardware assertion it doesn't check.

### 2. No receipt, no ledger row

The convention here is that a perf lever lands with a JSON under `docs/perf-deep-dive/PERF_RECEIPTS/same-host/` and a row in `STAMPEDE_CONDUCTOR.md` — that's how `CAMELID_WIN_PIN` and the #362 physical-core promotion are recorded. This PR cites `p1-cpu-thread-sweep-llama3b-20260621T2150Z.json` as the *source of the defect* but commits no receipt of its own, so the intervals now live only in a PR description and a module doc comment. Meanwhile `STAMPEDE_CONDUCTOR.md:162` still describes the decode pool as physical-core sized, which this PR makes stale.

Given that the argument for default-on is precisely "this replaces a promoted default derived from a committed receipt," the replacement should be a committed receipt too.

### 3. `CAMELID_WIN_HYBRID_POLICY` is documented nowhere outside the module

An operator hitting a regression on unusual silicon has no way to discover the kill switch. Folding this into the same doc commit as #2 covers it.

### 4. Minor — no action required, noting for the record

- `src/inference/win_pin.rs:231` — `GetProcessAffinityMask` fails on a process spanning processor groups and falls back to `usize::MAX`, so a >64-logical-CPU hybrid host sizes off group 0 regardless of where it's actually running. You document the group-0 restriction and no such client part exists today, so this is fine; I just want it written down somewhere other than a PR thread.
- `src/inference/win_pin.rs:295` — the selected masks are retained in the `OnceLock` but only ever `.len()`'d. Fine if it's a deliberate hook for a future placement lever; otherwise the count alone would do.
- The `physical_cores` field in the sizing log is the group-0 count from the `...Ex` walk, which can differ from `windows_physical_core_count()` on a multi-group host. Cosmetic, but the two numbers in that log line come from different denominators.

### 5. The two "pre-existing failures" don't reproduce in CI

`cargo test --all-targets` is green on ubuntu, macOS and windows for this branch. `cooperative_stream_job_seeds_from_the_prompt_prefix_cache` and `slot_occupancy_tracks_cooperative_and_exclusive_work` are both timing-sensitive, and your own Limitations section says the host was memory-starved — so these look like flakes under load rather than a broken `main`. That's consistent with your conclusion that they aren't attributable to this branch, and I agree it doesn't block. It does mean there's a separate flaky-test problem in the #557 cooperative-streaming code worth its own issue.

---

**To land:** the Intel gate (or an equivalent bound) from #1, plus the doc/receipt commit covering #2 and #3. The measured win, the byte-identical output and the honest re-measurement are all solid — I just don't want the one unverified hardware assumption to be the thing carrying default-on.

karan68:
Fixed all three. New commit `ae0a0d2d6`.

**The AMD point — you're right, and I should have caught it myself.** I wrote "every AMD part reports one class" from the P/E mental model, not from anything I actually checked. Zen 5c and the multi-CCD CPPC case both break it, and because the sweep puts 4 workers at −2.976 against the default, the failure mode isn't neutral — it's a measured regression on hardware I've never touched.

I went with the CPUID gate rather than a floor, for the reason you gave: any threshold I picked would be invented, whereas a vendor check states exactly what's true — this was measured on one Intel hybrid part, so it only runs there. `vendor_is_intel(ebx, edx, ecx)` is split out from the CPUID call so the decode is a unit test instead of something that only runs on my desk. AMD keeps its current width until someone with the hardware measures one, and the gate comes off with a receipt attached.

Also dropped the hardware claim from the single-class test. It now states the property it actually proves, which is a fact about the selection, not about anyone's silicon.

**Receipt and ledger** — fair, and once you said it the inconsistency was the part that bothered me: the entire argument for default-on is "a promoted default derived from a committed receipt went stale," which only holds if the replacement is committed the same way. Added `stampede-p1-hybrid-decode-width-14900hx-20260731.json` with the A/B, the width sweep, the placement sweep and the parity hashes, plus a `supersedes` pointer to the 11800H receipt. New ledger row 1.1, and line 162 now mentions the hybrid case so it stops being wrong.

**Kill switch** — documented in the Phase 1 section as a table of all four thread knobs and which one beats which, since the precedence was previously only discoverable by reading `resolve_decode_thread_count_from`.

**Your section 4** — did all three, they were small. Group-0 caveat written into the module docs. The retained masks weren't a hook for anything, so the `OnceLock` holds the count now, which collapsed two functions into one. Log denominator is `group0_physical_cores`.

**Section 5** — agreed. I'll open a separate issue for the two cooperative-streaming tests rather than leave it buried in this thread.

One thing I hit that's probably worth its own look: `__cpuid` is a safe fn on the toolchain `rust-toolchain.toml` pins (1.95), so the `unsafe` block trips `unused_unsafe` and dies under `-D warnings`. But `ci.yml` installs 1.89 and `release.yml` installs 1.87. I've scoped an `#[allow(unused_unsafe)]` so it builds clean on all three, but three files disagreeing about the toolchain seems like something you'd want to know about.

Gates on `ae0a0d2d6`: `fmt --check` clean, `clippy --all-targets -- -D warnings` clean, win_pin tests 8/8, release build fine. Re-verified the shipped behaviour on the box after the gate went in — default still sizes to 8, `CAMELID_WIN_HYBRID_POLICY=off` returns it to 24, `CAMELID_DECODE_THREADS=3` still wins, and `CAMELID_THREADS=6` still disables the dedicated pool entirely.

timtoole02 (approved):
Follow-up addresses all three findings. Verified: the CPUID leaf-0 constants decode to exactly `GenuineIntel` and `AuthenticAMD` (EBX/EDX/ECX order correct), the receipt is valid JSON and names the superseded i7-11800H receipt, ledger row 1.1 lands with row 1 corrected, and the operator-knob table documents the kill switch and the thread-knob precedence. 8/8 win_pin tests pass locally; all nine CI checks green including `rust (windows-latest)`.

Noting for the record: the receipt records `camelid_head: ee3e537df`, the pre-gate commit that was actually measured. That's the right call — the Intel gate can only narrow which hosts engage, never change the measured host's behaviour, so the numbers still describe the shipped path on that machine.

Thanks for the re-measurement discipline on this one, and for retracting the pinning claim rather than defending it.

PR #584, opened 2026-08-01
chore(release): bump version to 0.4.9

timtoole02:
Version bump for v0.4.9: `0.4.8` -> `0.4.9` across the four files that carry it. Same shape as
every prior bump — 4 files, 5 lines, no other content. `Cargo.lock` is edited in place so the
release build's `cargo build --release --locked` stays satisfied.

## Why this release exists

**v0.4.8 shipped with no Windows desktop installer.** `desktop-windows` failed, the server
artifacts published anyway, the release became `latest`, and the documented one-command install
broke for every user. #583 fixed the cause and added two guards; this release is the first one
where the signing path completes end to end.

| PR | change |
| --- | --- |
| #583 | absolute `signCommand` paths (the v0.4.8 cause), `verify-release-assets` demotion guard, and installer-script fallback |

## What this release run is the first real test of

The Azure signing call itself, and both new guards. The *wiring* is already proven by a real
local bundle — all seven `signCommand` invocations spawn and the installer builds — but the
signing service call only happens on a tag push.

Failure can no longer reach users, which is the point:

- `Prove the INSTALLED binary is signed` installs the built installer on the runner and asserts
  the unpacked binary verifies. A bad signature fails the job instead of publishing.
- `verify-release-assets` demotes any release with a missing asset to a prerelease, so
  `releases/latest` keeps pointing at a release that actually installs.
- `scripts/get-desktop-windows.ps1` walks back to the newest release that publishes an installer
  regardless, so the one-command install works even if a future publish gap slips past CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,7 +461,7 @@ dependencies = [
 
 [[package]]
 name = "camelid"
-version = "0.4.8"
+version = "0.4.9"
 dependencies = [
  "anyhow",
  "async-stream",
@@ -501,7 +501,7 @@ dependencies = [
 
 [[package]]
 name = "camelid-desktop"
-version = "0.4.8"
+version = "0.4.9"
 dependencies = [
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ resolver = "2"
 
 [package]
 name = "camelid"
-version = "0.4.8"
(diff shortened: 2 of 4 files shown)


timtoole02:
Superseded: the next release is v0.5.0, not v0.4.9. A fresh bump PR will be opened on top of #585 once that lands, so the version bump is the last commit before the tag.

PR #602, opened 2026-08-03
Add Gemma 4 Ghost-MoE Metal acceleration

timtoole02:
## What changed

- Adds the Gemma 4 26B Ghost-MoE runtime and hardened `.cghost` artifact handling.
- Keeps persistent routed Q4_0 expert slots on Metal while paging expert records from the local SSD.
- Adds exact file-backed Q6_K Metal tied-head execution, SIMDgroup Q4_0 kernels, asynchronous expert/tail chaining, and hybrid CPU-prefill-to-Metal-decode handoff.
- Adds truthful Ghost execution health, response-limit safeguards, exact prompt-token telemetry, cancellation/serialization hardening, and the empty-response fix.
- Updates the WebUI, runtime documentation, repacker, and focused smoke tests.

## Why

The original runtime could not practically keep all of Gemma 4 26B resident on an Apple Silicon machine. Ghost-MoE retains only the reusable and currently routed pieces in unified memory, with inactive expert data remaining on the internal SSD.

## User and developer impact

The selected M4 profile uses the CPU common core, persistent Metal Q4_0 expert slots, and the exact Metal Q6_K tied head. The 32-token acceptance benchmark completes in 8.42–9.04 seconds (3.54–3.80 end-to-end output tokens/second). The optimized expert SIMD kernel measured 572 µs versus 773 µs for the scalar-fused path.

The full-common Metal path is also available, but measured slower at 10.85–11.08 seconds under the tested memory pressure, so the runtime selects the faster hybrid lane.

## Validation

- Rust library suite: 1,589 passed, 0 failed, 24 ignored
- Rust binary tests: 31 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`
- Frontend UI, integration, streaming, execution-plan, and chat-gate smokes
- Production frontend build
- Live WebUI generation returned a non-empty exact `LIVE` response with backend token accounting
- Health endpoint confirmed `generation_ready=true`, `hybrid_metal`, Metal experts active, and Metal Q6_K head active

## Notes

This remains an explicitly experimental Gemma 4 Ghost-MoE lane without a parity guarantee. No model weights, generated binaries, caches, USB paths, credentials, or machine-specific paths are included in this branch.


timtoole02:
Perf update pushed in 02b71185. Root cause of the 3.54-3.80 tok/s: decode was dominated by synchronous expert page-in (61.7% slot hit rate at the old 24-slot ceiling = 290 MiB of critical-path SSD reads per token), not by the bandwidth wall (2.38 GB touched/token allows ~50 tok/s fully resident).

Changes:
- Persistent Metal expert slots now range 8..=128/layer (was 32). Measured routing skew: top-64 experts/layer cover 95.7% of routings, so residency converts directly into hit rate (61.7% -> 97.5% at 80 slots).
- Opt-in `CAMELID_GEMMA4_GHOST_METAL_TURBO=1`: reassociated-summation kernels (per-lane f32 + simd_sum, four rows per SIMD group so the shared activation is read once per four weight rows) — expert-stage bandwidth roughly doubles; branch-hoisted Q6_K decode for the tied head.
- Opt-in `CAMELID_GEMMA4_GHOST_METAL_HEAD_RESIDENT=1`: pins the ~600 MB tied table (default file-backed pages were refaulting under expert-paging pressure, 17 ms/token head sweeps -> 7 ms).
- With residency fixed, the full-common Metal lane now beats the hybrid; recommended profile updated (80 slots, ctx 1024, 64 MiB host cache).

Measured (16 GiB M4, internal SSD, acceptance prompt, greedy): 12.7 tok/s end-to-end over 256 visible tokens; steady-state 50-58 ms/token = 17-20 tok/s at 97.5% hit rate. A persistent serve keeps slots warm across requests, so follow-up requests decode at steady state from token 1. Turbo greedy IDs matched ordered for the full run; note the docs now record that the CPU-storage/hybrid/full-common lanes do not currently agree with each other on greedy IDs for this prompt (pre-existing; observed with stock kernels after regenerating the fingerprinted artifact pair with this branch's repacker).

Library suite 1589/0/24, fmt + clippy --all-targets clean.

PR #606, opened 2026-08-04
Promote the K-quant prefill owner to default-on for win-x86_64

karan68:
## What

Flips `kquant_prefill_owner` from opt-in to default-on **for `win-x86_64` only**. One resolver function, plus a kill switch and a test.

The owner itself is not new. It landed 2026-07-08, is bit-identical by construction, and has carried twin tests since. **What's new is that nobody re-checked its gate.**

## Why now

The promotion rule is written down in `STAMPEDE_CONDUCTOR.md`:

> **Gate:** prefill ≥ 28 tok/s → **GO**, promote owner to default. Between 24–28 → **PIVOT**: keep owner opt-in.

Lane B (Q4_K) measured into the PIVOT band at v2 and was parked. It then got faster **four more times** — the Q6_K sibling, the AVX-512 VNNI inner, the 8-row repack, and the 256-bit AVX-VNNI inner merged as #593 — and the flag stayed off through all of it.

A default that was correct when it was set is not automatically still correct.

## Measured

`bench-owner-sweep --lane kquant --max-tokens 1`, arms interleaved inside one process per round, 2 warmup rounds discarded, **AC power** (`BatteryStatus=2` per run), `CUDA_VISIBLE_DEVICES=-1`. Baseline is `off` — the question is owner vs *the shipped default*, not owner vs a different inner.

Ratios are `owner_avxvnni256` (what a promoted default actually selects on this host) vs `off`:

| model | 124 tok | 606 tok | 2421 tok |
|---|---|---|---|
| Llama-3.2-1B Q4_K_M | **1.8768** [1.8706, 1.8881] 8/8 | **1.7488** [1.7327, 1.7590] 8/8 | **1.5619** [1.5510, 1.5713] 6/6 |
| Llama-3.2-3B Q4_K_M | **1.8288** [1.8166, 1.8933] 8/8 | **1.7990** [1.7778, 1.8370] 8/8 | **1.6169** [1.6074, 1.6213] 5/5 |
| Qwen3-4B Q4_K_M | **1.8176** [1.7983, 1.8278] 8/8 | **1.7634** [1.7404, 1.7891] 8/8 | **1.5700** [1.5465, 1.5821] 5/5 |

**9 of 9 Q4_K_M cells significant. 66 of 66 paired rounds won. 1.56×–1.88×.**

**This does not depend on AVX-VNNI.** The `owner_avx2` arm — what an AVX2-only host selects — is **1.41×–1.61×** across the same nine cells, all significant. I own an AVX-VNNI part, so I measured the arm that doesn't use it.

## What the flag actually gates (self-review correction)

An earlier version of this PR said Q5_K_M / Q3_K_M / Q2_K / IQ4_XS shared this gate and were unswept. **That was wrong.** Grepping `kquant_prefill_owner_enabled` finds exactly three call sites:

| site | quant |
|---|---|
| `q4_k_block_dot_core_with_repack` | Q4_K |
| `matmul_rhs_transposed_q4_k_block_dot` | Q4_K |
| `q6_k_block_dot_core` | **Q6_K** |

All three are additionally guarded by `n_rows >= 4`, which is why decode is untouched. Q5_K, Q3_K, Q2_K and IQ4_XS have **no owner at all** and cannot change. (A Q5_K_M *model* is still partly affected, because those mixes carry Q6_K tensors.)

The real gap was the opposite of the one I'd written down: **Q6_K is gated, and my Q4_K_M sweeps only exercise it as the per-layer Q6_K tensor — one matmul in seven.** A model that is Q6_K throughout was the case nobody had measured, and the only receipts on file for it show the owner **slower** (23.76 vs 24.21 tok/s).

So I measured it, on a native Llama-3.2-1B Q6_K (974.5 MB, sha256 `0f4c510d…`), 606 tokens, 8 rounds:

| config | tok/s | vs off | wins | engaged |
|---|---|---|---|---|
| off | 72.693 | — | — | 0 |
| owner_avx2 | 108.307 | 1.4929× [1.4488, 1.5016] | 8/8 | 336 |
| owner_avxvnni256 | 108.387 | **1.4937×** [1.4835, 1.5004] | 8/8 | 336 |

**A 1.49× win, not a regression.** Those 2026-07-08 receipts don't describe the shipping path — they predate the fix that made the Q6_K wrapper actually delegate to the owner, and they used a local requant, which routes to the EXPERIMENTAL lane rather than native kernels. This receipt supersedes them.

One detail worth keeping because it argues the number is real rather than lucky: the AVX2 and AVX-VNNI arms land within **0.1%** of each other here. That is exactly what should happen — Q6_K has no VNNI inner at any width, so the kernel merged as #593 has nothing to do on this model.

**Combined: 10 of 10 cells significant, 74 of 74 paired rounds, 1.49×–1.88×, both gated paths covered.**

## The deciding test

The Q8 owner is the same shape of lever, promoted the same way — and it **regresses on a 1B model** (0.9730 vs off at 2421 tokens, 0/8 rounds; receipt `stampede-p3-q8-owner-avxvnni256-14900hx-20260803.json`). The K-quant owner had only ever been measured on 3B/4B, so small models were the specific risk to this promotion.

I pulled Llama-3.2-1B Q4_K_M specifically to look for that failure. It shows **the largest wins of anything measured**. The failure mode does not reproduce on this lane.

One cell (`qwen4b` @ 2421) first came back contaminated — its `off` arm spanned 10.18–24.12 tok/s and the CI was `[0.6876, 1.9466]`, not significant. It was **re-run clean** (off 23.58–24.11, owner 37.29–37.56 → 1.5700, 5/5) and the contaminated cell was **discarded, not averaged in**.

## Cost

| model | peak RSS off | on | Δ |
|---|---|---|---|
| Llama-1B Q4_K_M | 985.7 MB | 987.9 MB | +0.22% |
| Llama-3B Q4_K_M | 2314.2 MB | 2310.5 MB | −0.16% |
| Qwen3-4B Q4_K_M | 2814.0 MB | 2816.6 MB | +0.09% |

Within noise, in both directions. The owner reuses a thread-local packed-input scratch — this is **not** the 8-row repack, which has a real memory cost, is a separate flag, and stays off.

### P1 — the parity tests were vacuous. Fixed, and the fix is ablated.

Correct, and **wider than the two tests you linked**. Five tests build their reference leg by *removing* the owner variable, which meant "off" right up until this branch made an unset variable mean "on":

`q4_k_owner_prefill_bitwise_matches_block_dot_core`, `q6_k_owner_prefill_bitwise_matches_block_dot_core`, the repack8 bitwise test, `q4_k_repack8_budget_zero_degrades_bit_identically`, and `q4_k_owner_avxvnni_inner_is_bit_identical` (the one I added in #593).

The existing engaged-check cannot catch this: it only asks whether the *owner* leg dispatched, never whether the *base* leg abstained. All five now set `"0"` explicitly and assert `kquant_owner_prefill_taken == 0` before comparing.

**Ablated** — restoring the old line makes it fire:

```
assertion `left == right` failed: baseline leg dispatched the owner (n_rows=4) — vacuous comparison
```

So the guard is real, not decorative.

### P2 — "decode untouched" was false. Corrected, and measured.

Also correct, and the doc comment on `x86_kquant_matmul_owner_enabled` had been saying so all along. Batch-1 decode is genuinely unaffected, but the CPU spec-decode chunk verify submits 4–8 row batches and clears the `n_rows >= 4` gate.

Disclosure alone didn't feel like enough: a verify chunk is far smaller than the 124–2421 token prefills this promotion is argued on, and **the equivalent Q8 owner regresses on exactly that kind of small work**. So I measured it.

`bench-speculative --drafter ngram --draft-tokens 4`, 3B Q4_K_M, CPU forced, interleaved, 4 rounds:

| | owner off | owner on |
|---|---|---|
| spec tok/s (steady, r1–r3 median) | 19.06 | **31.84** |
| verify_ms | 3,305.3 | **1,978.5** |
| lossless | true | true |

**1.67× on the verify path, `lossless=true` and `first_divergent=-1` in all 8 runs**, `accept_rate` 1.0 throughout. `cpu_verify_rounds=13` with 4.85 accepted tokens/round confirms the chunks really do clear the gate. Round 0 was a cold start in both arms and is excluded as warmup rather than averaged in.

So speculative decode *is* affected — it's **upside I failed to disclose**, not a hidden cost.

The stale `Default OFF` doc comment is updated too.

## Parity, and proof the flip actually works

Generated text is byte-identical off↔on: `e07ef0c0bea3e982` (1B), `788682831fb9ada1` (3B), `f5f70dae91dd9706` (Qwen).

Three arms on the **built binary** — no env var (the promoted default), the kill switch, and explicit opt-in:

| model | default | kill switch off | explicit on | SHA |
|---|---|---|---|---|
| Llama-3B Q4_K_M | **7,511 ms** | 13,424 ms | 7,629 ms | `0776215f49fa7325` |
| Qwen3-4B Q4_K_M | **10,525 ms** | 20,188 ms | 10,932 ms | `f7bcca771fbefcdc` |

The default is now fast (1.79× / 1.92×), the kill switch genuinely restores the pre-promotion timings, and all three arms produce one SHA per model.

## Blast radius

- **Scope is `win-x86_64` only**, matching what was measured and mirroring how the Q8 owner's default was scoped. Every other target keeps `env_flag_default_off` — nothing changes for them.
- **Kill switch:** `CAMELID_X86_KQUANT_MATMUL_OWNER=0|false|off|no|disabled`.
- **Typo-safe:** any *other* value leaves the owner on, so a mistyped kill switch cannot silently disable a measured default. Same convention as `CAMELID_WIN_HYBRID_POLICY` (#558).
- `repack8` unchanged and still opt-in.

New `env_flag_default_on` helper; the new test pins the scoped default and asserts every off-word disables and every on-word enables, on every target.

## Gates

`cargo fmt --all --check` clean · `cargo clippy --all-targets -D warnings` clean · `cargo test --lib` **1463 passed / 2 failed**.
Those 2 (`api::engine::tests::slot_occupancy_tracks_cooperative_and_exclusive_work`, `api::tests::cooperative_stream_job_seeds_from_the_prompt_prefix_cache`) were **re-proved pre-existing at `77b054dd2`** by stashing this branch's `src` changes and rerunning on the clean tree — not assumed from an earlier run, since main has moved and #600 touched streaming. Neither is in the inference path.

## Open questions

- **One host.** Scoped to `win-x86_64` but measured on a single hybrid Raptor Lake part. An AVX2-only or AMD x86 host would strengthen or narrow the scope.
- **Pure Q6_K got one model at one prompt length**, not the full three-length sweep the Q4_K_M models received.
- **The Q8 owner's 1B regression stays open** and is not addressed here; separate flag, separate default.
- The written gate is an absolute tok/s against a llama.cpp pin on an i7-11800H. I argue this on **ratio-vs-shipped-default plus zero regression** instead, because I did not re-run that pinned reference on this host. If you'd rather have the gate satisfied literally, that's a llama.cpp comparison I can run.

Receipt: `docs/perf-deep-dive/PERF_RECEIPTS/same-host/stampede-p3-kquant-owner-default-on-14900hx-20260804.json`, ledger row 3 updated.


timtoole02:
**[P1] Windows parity tests become vacuous.** The [new default](https://github.com/timtoole02/Camelid/blob/7d6286dba83923380260ba085f6938d1eb180225/src/runtime_config.rs#L74-L82) makes an unset flag mean “on” for Windows x86_64, but the [Q4_K](https://github.com/timtoole02/Camelid/blob/7d6286dba83923380260ba085f6938d1eb180225/src/inference/tests.rs#L3819-L3827) and [Q6_K](https://github.com/timtoole02/Camelid/blob/7d6286dba83923380260ba085f6938d1eb180225/src/inference/tests.rs#L4118-L4125) tests remove the variable to create their “base” leg. On Windows, both legs now run the owner, defeating the tests’ engaged-check. Set the baseline explicitly to `"0"` and assert the owner counter remains zero before testing `"1"`. The repack/VNNI tests contain the same assumption.

**[P2] “Decode untouched” understates the blast radius.** The existing implementation explicitly says batches of four or more include [CPU speculative-decode verification](https://github.com/timtoole02/Camelid/blob/7d6286dba83923380260ba085f6938d1eb180225/src/inference.rs#L20013-L20017), while the [receipt calls decode untouched](https://github.com/timtoole02/Camelid/blob/7d6286dba83923380260ba085f6938d1eb180225/docs/perf-deep-dive/PERF_RECEIPTS/same-host/stampede-p3-kquant-owner-default-on-14900hx-20260804.json#L76-L80). Ordinary batch-1 decode is unaffected, but speculative verification changes paths. The PR should disclose that and ideally include a speculative-decode parity/performance check. The stale “Default OFF” comment in `inference.rs` should also be updated.

PR #616, opened 2026-08-05
feat(fabric): place whole requests across independent Camelid nodes

karan68:
## What

Adds `src/fabric`: a scheduler that places **whole requests** on Camelid nodes that each own a complete model and a complete session, plus a `camelid fabric` CLI to observe, dry-run and send.

Nothing crosses the network inside the token loop, so throughput scales with nodes instead of being taxed by them.

## Why this is not `serve-distributed`

The existing distributed lane shards one model's layers across machines. It is a **capacity** mechanism — it runs a model that fits on neither machine — and it is latency-additive by construction: every generated token crosses the wire twice and each machine idles while the other computes.

The fabric is the complementary arrangement, for **throughput and heterogeneity**. The two compose: a fabric node may itself be a distributed coordinator.

```
serve-distributed :  one model,  N machines,  N hops per token
fabric            :  N models,   N machines,  0 hops per token
```

## Shape

| File | Role |
| --- | --- |
| `fabric/node.rs` | Identity, observed state, `label=host[:port]` parsing |
| `fabric/policy.rs` | **Pure** placement decisions — no I/O |
| `fabric/probe.rs` | Turning `/v1/health` into a routing fact |
| `fabric/http.rs` | Minimal blocking HTTP/1.1, shared by probe and forward |
| `fabric/forward.rs` | Sending a placed request to the node that serves it |
| `fabric/mod.rs` | `Fabric`, `dispatch`, summary, rendering |

The correctness of the fabric lives in `policy.rs`, which is pure and therefore testable exhaustively with no server, no model and no network.

Routing reads `/v1/health` alone — it already carries `active_model_id` and load — so placement costs one GET per node, and nodes are probed concurrently so a single dead node cannot stall the view of live ones.

## Two load-bearing properties

Each has a test that fails without it.

**Placement is deterministic.** Ties break on label, so `camelid fabric route` predicts what the fabric will actually do.

**Affinity degrades, never fails.** A warm prefix is an optimisation; when the node holding it dies the request is still served, and the decision reports `affinity_lost` so a caller can tell a warm hit from a cold re-prefill.

## Deliberate limits, stated rather than half-supported

- **Streaming is refused.** An SSE body is a different response shape and a different cancellation story. `reject_streaming` runs before a socket is opened, and a test proves the refusal beats the dial.
- **A node answering 503 is an answer, not a failure.** It arrives as a `Forwarded` with its status; only a request that never completed is a `ForwardError`, and that error names the node so an operator can act on it.
- **Node capacity is not modelled.** `/v1/health` publishes current load but never the queue bound — `engine_queue_depth` is a gauge of jobs in flight, and the real limit (`CAMELID_QUEUE_DEPTH`) is read on the node and not serialised. An earlier revision read that field as a bound and refused to route to a **fully idle two-node fabric**. The fabric now ranks on observed load and lets a genuinely full node answer its own 503.

`chat::client` is deliberately not reused: it is private to `chat`, is keyed on a resolved `SocketAddr` where fabric members are named hosts, and carries SSE, bearer auth and tool-call handling that a health probe must not depend on.

## Verification

**Two real machines** — Windows RTX 4060 and a Mac M5 Pro over Wi-Fi, running engines **0.5.4 and 0.5.0** respectively, so version skew is covered by the lenient health payload. Status, throughput placement on an idle fabric, affinity honoured, affinity degrading to a live node, typed refusal for an unserved model, and malformed and duplicate node specs refused before any I/O.

**Two different models** — `Llama 3.2 1B Instruct` on GPU and `Qwen3 4B Instruct Awq` on CPU. Model-scoped placement chose the right node in both directions and real completions came back from both (743 ms and 42934 ms, the latter CPU-bound). A request with no `--model` picked the least-loaded node and used that node's model. An unserved model was refused naming both models the fabric does serve.

**Reproducibly, in CI** — `tests/fabric_end_to_end.rs` stands in two HTTP nodes on loopback, so all of the above runs with no model, no GPU and no second machine. Those tests also assert the **request** side (the fabric sends `POST /v1/chat/completions` with valid JSON carrying model, prompt and `stream:false`) and that placement is exclusive (after a model-scoped dispatch, the node that does not hold the model shows only a health probe).

## Gates

- 11 integration tests, 63 fabric unit tests
- Full lib suite **1632 passed / 0 failed** / 84 ignored
- `clippy --all-targets -- -D warnings` clean
- `rustfmt --check` clean
- `scripts/check-public-scrub.sh` clean
- All new files LF; `git diff --check` clean

## Not in this PR

No streaming. No probe caching or background refresh — every command re-probes. No operator-declared node capacity. No HTTP front door: `fabric run` sends one request from the CLI, and a resident proxy that clients point at is the natural next slice, built on the same `dispatch`.

### DIFF
--- a/src/fabric/forward.rs
+++ b/src/fabric/forward.rs
@@ -1,0 +1,264 @@
+//! Forwarding a placed request to the node that will serve it.
+//!
+//! Placement ([`super::policy`]) chooses a node; this sends the request there and
+//! brings the answer back. The whole point of the fabric is that this happens
+//! once per request rather than once per token, so a node's own generation loop
+//! never waits on the network.
+//!
+//! Streaming is deliberately out of scope for this path. An SSE response is a
+//! different response shape and a different cancellation story; rather than
+//! half-support it, a streaming request is refused with a reason.
+
+use std::time::{Duration, Instant};
+
+use serde_json::{json, Value};
+
+use super::http::{self, HttpError};
+use super::node::NodeSpec;
+
+/// Generation can legitimately take minutes, so this is far longer than a probe
+/// budget. It exists to bound a wedged node, not to bound a slow model.
+pub const DEFAULT_FORWARD_TIMEOUT: Duration = Duration::from_secs(300);
+
+/// A completion body is far larger than a health body but still bounded.
+const MAX_RESPONSE_BYTES: usize = 8 * 1024 * 1024;
+
+/// The node's answer, tagged with which node produced it.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Forwarded {
+    pub label: String,
+    /// The node's HTTP status. A node answering 503 is a real answer, not a
+    /// forwarding failure, so it arrives here rather than as an error.
+    pub status: u16,
+    pub body: Value,
+    pub elapsed: Duration,
+}
+
+impl Forwarded {
+    pub fn is_success(&self) -> bool {
+        (200..300).contains(&self.status)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ForwardError {
+    /// The request never completed against that node.
+    Transport { label: String, detail: String },
+    /// The node answered, but not with JSON.
+    Json { label: String, detail: String },
+    /// Refused before sending; see [`reject_streaming`].
+    Unsupported(String),
+}
+
+impl std::fmt::Display for ForwardError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Transport { label, detail } => {
+                write!(f, "node `{label}` did not answer: {detail}")
+            }
+            Self::Json { label, detail } => {
+                write!(f, "node `{label}` returned an unreadable body: {detail}")
+            }
+            Self::Unsupported(detail) => write!(f, "{detail}"),
+        }
+    }
+}
+
+impl std::error::Error for ForwardError {}
+
+/// Refuse a streaming request rather than mis-parse an SSE body as JSON. Pure.
+pub fn reject_streaming(body: &Value) -> Result<(), ForwardError> {
+    if body.get("stream").and_then(Value::as_bool) == Some(true) {
+        return Err(ForwardError::Unsupported(
+            "the fabric does not forward streaming requests yet; \
+             send the request without `stream: true`"
+                .to_string(),
+        ));
+    }
+    Ok(())
+}
+
+/// Build a minimal OpenAI-shaped chat request. Pure.
+pub fn chat_request(model: &str, prompt: &str, max_tokens: u32) -> Value {
+    json!({
+        "model": model,
+        "messages": [{ "role": "user", "content": prompt }],
+        "max_tokens": max_tokens,
+        "stream": false,
+    })
+}
+
+/// Pull the assistant text out of a chat completion body. Pure.
+///
+/// Returns `None` rather than a placeholder so a caller can tell "the node
+/// answered with no content" from "the node said nothing useful".
+pub fn completion_text(body: &Value) -> Option<&str> {
+    body.get("choices")?
+        .as_array()?
+        .first()?
+        .get("message")?
+        .get("content")?
+        .as_str()
+}
+
+/// Read an engine error message out of a non-2xx body, if it carries one. Pure.
+pub fn error_message(body: &Value) -> Option<&str> {
+    body.get("error")
+        .and_then(|error| error.get("message").or(Some(error)))
+        .and_then(Value::as_str)
+}
+
+/// Send one request to one node.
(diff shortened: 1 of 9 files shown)


timtoole02 (commented):
Reviewed and verified locally on Windows. This is strong work — the architectural argument holds up, and the code backs it. `policy.rs` being pure and I/O-free is the right call, and the two regression tests are the standout: they record *why* they exist rather than just what they assert, which is what makes the `engine_queue_depth` story survivable for the next reader.

I independently confirmed the load-bearing semantic claim. `src/api/mod.rs` sets `engine_queue_depth: state.engine.depth()`, and `src/api/engine.rs` documents that field as "jobs accepted but not yet finished (queued + running)". So reading it as `in_flight` is correct, and the earlier revision that treated it as a bound really was wrong. Good catch, and good that the fix is pinned by a test in both `probe.rs` and `policy.rs`.

## Verified locally

Beyond CI, on Windows against this branch:

| Gate | Result |
| --- | --- |
| `cargo test --lib fabric::` | 65 passed / 0 failed |
| `cargo test --test fabric_end_to_end` | 11 passed / 0 failed |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo fmt --check` | clean |
| `scripts/check-public-scrub.sh` | clean |
| Merge against current `main` + `cargo check --all-targets` | clean, exit 0 |

`+2857 / −0` with no coupling to the rest of the crate — `src/fabric/`'s only `crate::` reference is to its own `node` module — so the blast radius is exactly one new subcommand. That is why this is landing now rather than waiting on the gap below.

## Known limitation, recorded before merge

**`camelid fabric run` cannot reach a node that has an API key set, and `fabric route` will not warn you.**

`route_requires_auth` in `src/api/server.rs` exempts `/v1/health` but gates `/v1/chat/completions`. `src/fabric/` never sends an `Authorization` or `X-API-Key` header. Against a node with `CAMELID_API_KEY` configured:

- `fabric status` reports it **ready**
- `fabric route` selects it and reports success
- `fabric run` gets **HTTP 401**

That collides with the PR's own stated property — *"placement is deterministic, so `camelid fabric route` predicts what the fabric will actually do."* Under auth the dry run predicts a success that cannot happen.

It compounds with `src/api/server.rs`, which refuses to bind a non-loopback listener without a key unless `--allow-unauthenticated-remote` is passed. So the only configuration in which a real multi-machine fabric works today is one where every node is fully unauthenticated on the LAN. That is presumably how the two-machine run was done, and it is a reasonable place to start — it just needs to be said out loud, since "Not in this PR" doesn't currently cover it.

Follow-up is queued: a `--bearer` flag threaded through `probe` and `forward` into `http::request`.

## Other gaps worth naming

**Placement degrades; forwarding does not.** `place()` probes, picks, then hands off to `forward()`. If the chosen node dies in that window the request fails with `Transport` even when healthy alternatives exist. "Affinity degrades, never fails" is real, but it stops at the placement boundary — worth adding to the scope list alongside the others.

**`Throughput` mode does not spread sequential requests.** On an idle fabric every node reports `in_flight: 0`, so selection falls through to the label tie-break and always picks the alphabetically-first node. Since `fabric run` is one-shot (probe, forward, wait, exit), a script issuing ten sequential requests sends all ten to the same machine. The headline claim holds for concurrent callers, which is the resident proxy's job — the determinism-over-spreading trade is a deliberate and defensible choice, it is just currently unstated.

**`request()` only returns once the peer closes.** The read loop runs to `Ok(0)` and only then parses, so a satisfied `Content-Length` never short-circuits. Axum honours `Connection: close`, so this is correct today — but any keep-alive intermediary (nginx, a muxing tunnel) makes every call block to the deadline: 2s for a probe, **300s for a forward**. The parser already knows the content length, so stopping early would remove the dependency on peer-initiated close.

## Nits

- `route --json` emits `"reason": "LeastLoaded"` via `format!("{:?}")` — `RouteReason` derives no `Serialize`. `status --json` in the same command family emits snake_case (`"state": "ready"`). Inconsistent, and a variant rename would silently change the JSON contract.
- `NodeReady.version` is collected but never rendered by `render_status`; it appears only under `--json`. Version skew (0.5.4 vs 0.5.0) is a verification highlight in the description, yet the default table cannot show it.
- `Fabric::dispatch` is exercised only by tests — `fabric run` re-implements place-then-forward inline so it can resolve the model id from the chosen node. The tested path and the shipped path diverge.
- No documentation for a new top-level user-facing command. No CI gate catches this (`check-ledger-drift.mjs` is model-ledger only), but README's "Ways to use Camelid" table is the obvious home.
- `Fabric::new` is public and does not reject duplicate labels the way `parse_fabric` does; `place()`'s `find` would then return the first match rather than the routed one. Library-only — the CLI always goes through `parse_fabric`.

Merging. The auth gap is a limitation of a new opt-in feature with no users yet, not a regression, and nothing on an existing path changes.

PR #617, opened 2026-08-05
perf(inference): carry the Q4_K owner's group scale in its widening madd

karan68:
# Carry the Q4_K owner's group scale in its widening `madd`

The K-quant prefill owner's AVX2 inner widened its `maddubs` products with `madd_epi16(., ones16)` and then applied the 6-bit group scale with a **separate `mullo_epi32`**. `madd_epi16` can take the scale in place of the ones vector, reducing and scaling in one instruction.

This is an **identity**, not an associativity argument: `madd(p, ones) * s` and `madd(p, s)` are both `s·p[2k] + s·p[2k+1]` in every lane.

The mins side — an 8-wide `i64` **scalar** loop run per (superblock × activation row) — becomes one `mullo_epi32` plus the existing horizontal sum, with the widened mins hoisted per weight row alongside the scales already hoisted there.

Main-side cost drops from **5 uops per 32 values to 3**, and that is what makes plain AVX2 overtake `vpdpbusd`: that instruction returns i32, so scaling its result needs the 2-uop `mullo_epi32`, whereas `madd_epi16` gets the scale for free.

## Consequence: the 256-bit AVX-VNNI inner is no longer the default

It is **kept and reachable** via `CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256=1`, because it was compared on exactly one microarchitecture. Zen 5 also has AVX-VNNI and was not measured.

**The AVX-512 inners are untouched.** This host has AVX-512 fused off, so the same fold could not be measured there. Both AVX-512 inners have the identical `dpbusd` + `mullo_epi32` shape and the same scalar mins loop, so the fix likely applies — that is an unmeasured follow-up, not a claim.

## Bit identity

Every intermediate stays exact integer inside the envelope the kernel already documented: `|maddubs pair| ≤ 2·15·127 = 3810`, scale `≤ 63` after the kmask unpack, so a madd lane is `≤ 480,060` and the 8-add accumulator `≤ 3.85M` — about 1.4% of `i32::MAX`. The mins side is smaller still: a group sum is `≤ 32·127 = 4064` and the total `≤ 2.05M`, so the `i64` it used was never needed. The load-bearing f32 chain is untouched, verbatim `q4_k_dot_arm` order.

## Kernel in isolation

Per (activation row × weight row) cell. Verbatim in-tree baselines, best of 3, reproduced across three runs. All four variants agree **bit for bit** over the whole output block.

| inner | in_dim 2048 | in_dim 8192 |
| --- | --- | --- |
| in-tree AVX2 | 56.99 ns | 237.28 ns |
| in-tree AVX-VNNI | 43.87 (1.30×) | 188.24 (1.26×) |
| + scale fold | 38.42 (1.48×) | 163.32 (1.45×) |
| + mins SIMD | **37.00 (1.54×)** | **161.26 (1.47×)** |

## Prefill, end to end

i9-14900HX, Windows x86_64, AC power, CUDA hidden. Release binaries from a clean detached worktree at this branch's base (`v0.5.4-32-g7985e1d32`) and from this branch. 606-token prompt, 5 paired rounds, `Llama-3.2-1B-Instruct-Q4_K_M`.

Two arms are controls: `off` disables the owner, and `owner_avxvnni256` runs the kernel this PR does **not** touch — same binaries, same model.

| arm | per-pair before/after ratios |
| --- | --- |
| `owner_avx2` (changed) | 1.1123 1.1891 1.2181 1.2255 1.2240 |
| `owner_avxvnni256` (untouched) | 0.8941 1.0048 0.9915 0.9845 0.9809 |
| `off` (control) | 0.9101 0.9974 0.9283 1.0109 1.0021 |

The changed arm's **worst** pair (1.1123) is above the **best** pair of either control (1.0109, 1.0048) — complete rank separation, 5 observations against 10, no overlap. That, rather than the 1.218× median, is what carries the result: a median can be bought with a lucky run, rank separation cannot.

Comparing the two inners directly on the after binary, the new AVX2 inner beats the 256-bit VNNI one by **1.086× (5/5)**. That is what an AVX-VNNI host gains; a host **without** AVX-VNNI gains the **1.218×**, since the AVX2 inner was already its only path.

Two independent estimates of the Q4_K matmul share of prefill agree: 1.54× kernel → 1.218× end-to-end implies **51%**; 1.186× → 1.086× implies **50%**.

### What is deliberately not claimed

**No 3B end-to-end number.** The 3B sweep was run twice and discarded both times: in the second run the owner-**off** control — which this PR cannot affect — itself spanned **0.486 to 1.330**, so the run cannot separate the change from drift on this box. The kernel probe covers the 3B-relevant shape directly (`in_dim` 8192 is `ffn_down`, 1.47×), and 3B byte identity is reported below, so nothing rests on the discarded runs.

**Negative-control model.** `Llama-3.2-1B-Instruct-Q6_K` holds no Q4_K tensors, and `kquant_avxvnni256_taken` is 0 for it in every arm — confirming the absence rather than assuming it.

## Byte identity on real weights

64 greedy tokens after the same prompt. Identical token ids **and** text sha256 in all six cells — three models × {default, VNNI sub-flag off}.

| model | sha256 (prefix) |
| --- | --- |
| 1B Q4_K_M | `27beb9855204ec9c64533da0` |
| 3B Q4_K_M | `109708da6da2bbdf90976fed` |
| 1B Q6_K | `641b4526fdf3b738a6f39b18` |

The `default` cell is the strong one: the before binary takes the 256-bit VNNI inner there and the after binary takes the new AVX2 inner, so it exercises the kernel rewrite **and** the dispatch change together.

## Negative controls on the tests

Each reproduced before it was trusted; each fails `q4_k_owner_prefill_bitwise_matches_block_dot_core`:

1. scale replaced by ones
2. low/high group scales swapped
3. mins side adds instead of multiplies
4. the 256-bit inner allowed to dispatch without its opt-in flag

(4) is why the test now asserts `kquant_owner_avxvnni_taken == 0` — without it, the AVX2 legs could silently have been measuring the other kernel.

## Gates

`cargo fmt --all -- --check` clean · `git diff --check` clean · `cargo clippy --all-targets -- -D warnings` clean · `cargo test --lib` **1569 passed, 0 failed**.

The existing `q4_k_owner_prefill_bitwise_matches_block_dot_core` compares the owner against the non-owner per-cell path bit for bit over `in_dim` 512/768, `out_dim` 96/40 and `n_rows` 4/5/13/64/67, and needed no change beyond that one assertion. `q4_k_owner_avxvnni_inner_is_bit_identical` now asks for the inner through its new flag.

## Scope

No flag default, support claim or ledger row changes for any user-visible capability. The K-quant owner itself remains opt-in at this base; this makes its Q4_K half faster when it is on.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -20202,6 +20202,32 @@ fn x86_kquant_matmul_owner_vnni_enabled() -> bool {
     }
 }
 
+/// The 256-bit `vpdpbusd` inner, default OFF: the AVX2 inner carries the group
+/// scale inside its widening `madd_epi16`, and measured faster than
+/// `vpdpbusd` + `mullo_epi32` at both shapes tried — 1.19x at in_dim 2048 and
+/// 1.17x at 8192 in isolation, 1.09x on 1B Q4_K_M prefill, on an i9-14900HX.
+/// Retained behind `CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256=1` because that
+/// comparison has only been made on one microarchitecture; dispatch is
+/// bit-identical either way. Resolved once like its siblings, with the same
+/// bypass so an A/B sweep is not pinned to whichever arm ran first.
+#[cfg(target_arch = "x86_64")]
+fn x86_kquant_matmul_owner_avxvnni256_enabled() -> bool {
+    #[cfg(test)]
+    {
+        q8_0_env_flag_enabled_default_off("CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256")
+    }
+    #[cfg(not(test))]
+    {
+        if q8_runtime::bench_uncached_runtime_plan() {
+            return q8_0_env_flag_enabled_default_off("CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256");
+        }
+        static ENABLED: OnceLock<bool> = OnceLock::new();
+        *ENABLED.get_or_init(|| {
+            q8_0_env_flag_enabled_default_off("CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256")
+        })
+    }
+}
+
 /// Shared-base output pointer for the K-quant owner's rayon tasks. Tasks
 /// partition the output-channel dimension into disjoint chunks — no two tasks
 /// ever write the same (row, channel) cell despite the shared base pointer.
@@ -20443,12 +20469,17 @@ unsafe fn q4_k_owner_weight_row_block_avx512vnni(
 
 /// v6 inner: 256-bit AVX-VNNI sibling of
 /// [`q4_k_owner_weight_row_block_avx512vnni`], for the Alder-Lake-and-later
-/// consumer parts that carry `vpdpbusd` but have AVX-512 fused off — the whole
-/// 12th-gen-onwards Intel client line, which today falls all the way back to
-/// the AVX2 inner. Same algebra as the AVX-512 sibling with the zmm pair split
-/// into two ymm halves: the low nibbles pair with `y.qs[64j..64j+32]` and the
-/// high nibbles with `y.qs[64j+32..64j+64]`, exactly the halves the 64-byte
-/// zmm load covers, so the operand pairing is unchanged.
+/// consumer parts that carry `vpdpbusd` but have AVX-512 fused off. Same
+/// algebra as the AVX-512 sibling with the zmm pair split into two ymm halves:
+/// the low nibbles pair with `y.qs[64j..64j+32]` and the high nibbles with
+/// `y.qs[64j+32..64j+64]`, exactly the halves the 64-byte zmm load covers, so
+/// the operand pairing is unchanged.
+///
+/// Opt-in via `CAMELID_X86_KQUANT_MATMUL_OWNER_AVXVNNI256=1`: `vpdpbusd`
+/// returns i32, so the group scale needs a 2-uop `mullo_epi32`, whereas the
+/// AVX2 inner folds the same scale into a widening `madd_epi16` for free and
+/// measured faster here. Kept because that was measured on one
+/// microarchitecture only.
 ///
 /// Bit-identity: every intermediate is exact integer. A `dpbusd` lane is a quad
 /// sum ≤ 4·15·127 = 7620; ×63 scale ≤ 480,060; `vacc` takes 8 adds (4 chunks ×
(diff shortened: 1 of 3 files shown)


karan68:
## Self-review pass — three findings, all fixed in `95949b193`

### 1. The new flag broke this file's own convention (real defect)

`x86_kquant_matmul_owner_avxvnni256_enabled()` resolved `std::env::var` on **every call** — once per matmul, so 336 times per 1B prefill and 588 per 3B. Both sibling sub-flags (`..._VNNI`, `..._REPACK8`) resolve through a `cfg(test)` / bench-bypass / `OnceLock` triple.

The trap: memoizing it *naively* would have silently broken the A/B harness this PR's evidence depends on — the sweep would latch whichever arm ran first and then measure the same kernel in both arms. That is exactly the "fake null" this repo already hit once and fixed with `CAMELID_BENCH_UNCACHED_RUNTIME_PLAN`. The fix carries that bypass, and I verified it on the rebuilt binary rather than trusting it:

```
owner_avx2         owner_taken=336   avxvnni256_taken=0
owner_avxvnni256   owner_taken=336   avxvnni256_taken=288
```

Still selecting per arm. Byte identity re-confirmed on the default path after the change: `27beb9855204ec9c64533da0ef9fccc3`, identical to the base binary.

### 2. An overstated number in the source

The flag's doc comment said "1.14-1.19x" with no qualifier — that is the **microbenchmark** figure, and left unqualified someone would reasonably quote it as end-to-end. It now names both: 1.19x / 1.17x in isolation at in_dim 2048 / 8192, and **1.09x on real 1B Q4_K_M prefill**.

### 3. A comment that contradicted its own code

I had written "the i64 the scalar loop used was never needed" on a line that still casts `as i64` — deliberately kept, so it stays symmetric with the untouched `main` line two rows below. Reworded to state the bound only.

## Re-gated after the fixes

`cargo fmt --all -- --check` clean · `git diff --check` clean · `cargo clippy --all-targets -- -D warnings` clean · focused owner tests 3/3.

`cargo test --lib` reported one failure, `gait::tests::memory_measurement_is_sane`, and it is **not** from this PR — worth stating how that was established rather than asserting it:

* it passes in isolation on the same checkout (2.63 s),
* CI `rust (macos-latest)` and `rust (ubuntu-latest, --all-features)` both **pass** on this branch,
* it lives in `gait`, which this diff does not touch,
* the identical kernel code returned 1569 passed / 0 failed on an earlier local run.

It is a hardware memory-bandwidth benchmark, and it ran while this laptop was saturated after several fat-LTO builds and hours of sweeps — the whole suite took 262–531 s against 65 s when the box is idle. A logic break would reproduce in isolation and on CI; this does neither.

## Deliberately not changed

* **The scalar twin and the stored weight layout.** Untouched, so there is no repeat of the aarch64 regression pattern where a SIMD-friendly layout makes the non-x86 path slower.
* **The AVX-512 inners.** They have the identical `dpbusd` + `mullo_epi32` shape and the same scalar mins loop, so the fold very likely applies — but `avx512vnni` is fused off on this host, so it cannot be measured here. Flagged as a follow-up rather than shipped as a guess.
* **Decode.** The owner declines below 4 rows, so decode never reaches this kernel.

PR #618, opened 2026-08-06
runnable: admit Q8_0 weights on the qwen3.5 resident Metal lane

heskew:
## What changed

Two type whitelists in `src/runnable/model.rs` now admit `Q8_0`:

1. `load_raw` (~:539) — add `GgufTensorType::Q8_0` to the `matches!` that selects `WirePages` (page-backed) over `RawMatBytes::owned`.
2. `prism_metal_format` (~:126) — add `GgufTensorType::Q8_0 => Some(ResidentWeightFormat::Q8_0)`.

Four lines. No `metal.rs` change, no new kernels, no repack.

## Why

I wanted to run Ornith-1.0-9B (`qwen35`, Q8_0) on the GPU on an Apple Silicon box. Today it is refused by the resident Metal lane and falls back to CPU decode:

```
[qwen35] resident Metal lane failed (invalid tensor data: Prism Metal weight is not page-backed); using hybrid fallback
```

Reading the tree, the Metal side already looked quantization-generic rather than Prism-specific:

- `Qwen35MetalDecode::new` (`metal.rs:16742`) validates geometry only — it never inspects a weight format.
- `encode_resident_matmul_f32` (`metal.rs:12054`) has a first-class `Q8_0` arm at `:12104`.
- `encode_qwen35_ssm_layer` (`metal.rs:18117`) issues all five SSM projections through that same generic dispatcher; the four genuinely SSM-specific kernels operate on f32 activations and never touch a quantized weight.

So the block was the loader, not the kernels. Both gates were scoped to the Prism quants, so `Q8_0` fell through to `owned` and then failed the page-backed check.

Admitting it needs no conversion step: GGUF `Q8_0` on disk (34 B — f16 scale + 32×i8) is already exactly the layout the Metal wire kernel consumes, as `wire_mmap.rs:3-6` notes. Worth adding that `WirePages` is not an mmap but a page-aligned heap allocation filled by one `read_exact_at`, so the 32-byte tensor alignment in GGUF never mattered here. This also follows precedent already in the tree — `src/runnable/vision.rs:1094-1123` page-backs `Q8_0` end-to-end through the same Metal path.

After the change:

```
[qwen35] full Metal resident graph active (packed weights, attention, gated-delta recurrence, FFN, logits, GPU greedy, and request sampling)
```

## Evidence

`bench-generate`, greedy (`--temperature 0`), warm (`--warmup --iterations 2`), `ornith-1.0-9b-Q8_0.gguf` on an M4 MacBook Air 16 GB:

| | before (CPU) | after (Metal resident) |
|---|---|---|
| decode, 5-token prompt | 9.80 tok/s | 11.3–11.5 tok/s |
| decode, 481-token prompt | 4.01–4.11 tok/s | 11.2–11.5 tok/s |
| TTFT, 481-token prompt | 48.5–57.0 s | 36.6 s |
| peak RSS | 8.0–8.2 GB | 4.8–5.2 GB |

**Greedy output is token-for-token identical to the CPU lane** on both the 5-token and the 481-token prompt (32 generated tokens each, token IDs compared directly).

The decode number is worth reading carefully: the Metal lane is context-stable (~11.3 tok/s at both 5 and 481 tokens) while CPU decays with context, so the ratio is ~1.16× at a short prompt and ~2.8× at 481 tokens. The RSS reduction comes from the weights going into Metal no-copy instead of being resident twice.

## Validation

Per `docs/VALIDATION_MATRIX.md`, "Inference/tokenizer/runtime changes" — standard backend gate plus targeted parity artifacts for the affected row:

- `cargo fmt --all -- --check` — pass
- `cargo clippy --all-targets -- -D warnings` — pass
- `cargo test --release --lib` — 1630 passed, 0 failed, 24 ignored
- `cargo doc --no-deps` — builds (one pre-existing `invalid_html_tags` warning in `tools/repack_ghost.rs`, untouched here)
- `bash scripts/check-public-scrub.sh` — pass
- Targeted parity: the token-ID comparison above

I could not run the `--all-features` legs locally: that pulls in `cuda`/`cudarc`, which does not build on macOS. Those need CI or a Linux host.

## Docs / compatibility rows

**None changed, deliberately.** Classifying this as **implementation groundwork only** — a lane that already existed now admits a quantization it already had kernels for. It is not a support claim for any row, and I have not touched `README.md`, `COMPATIBILITY.md`, or `STATUS.md`.

## Caveats I want to be upfront about

1. **Prefill is still slow on both lanes** and this PR does not address it — 36.6 s for a 481-token prompt on Metal (~76 ms/token), reproducible across iterations, so it is not shader compilation. Metal improves TTFT only ~1.4× over CPU. The decode numbers should not be read as a prefill claim.
2. **First call in a cold process pays ~15 s of one-time Metal pipeline compilation**, which lands inside the prefill timer. Benchmarks need `--warmup` or the first TTFT is meaningless (I measured 15.8 s cold vs 0.40 s warm on the same 5-token prompt).
3. **Blast radius is wider than `qwen35`, and it is not macOS-gated.** `load_raw` is shared by the whole runnable lane and the page-backing predicate carries no `cfg`, so **every `Q8_0` runnable model page-backs on every target — Linux and Windows included**, not only other macOS models. Same bytes, same size (page-rounded up). It is unobservable off macOS: `RawMatBytes::as_slice()` returns identical bytes for either backing, and `wire_pages()` — the only accessor that distinguishes them — is `#[cfg(target_os = "macos")]`; the CUDA `qwen35` lane already handles `Q8_0` and reads through `as_slice()`. The real off-macOS cost is rounding each tensor allocation up to a page.
4. **`CAMELID_METAL_WIRE=0` with this patch is untested.** `qwen35` passes `q8_wire: true` unconditionally while its weights are genuinely 34-byte wire, so it should stay self-consistent, but I have not exercised that path.
5. **K-quants are still not admitted** (`resident_metal_format` returns `None` for them). The `Q4K`/`Q6K` kernels exist and `CAMELID_METAL_KQUANT` defaults on, so this may be a similar small extension, but it is unverified and belongs in its own PR. Since `07201707` that extension is one arm in one match — page-backing follows from the mapping, so it is no longer possible to admit a type to half the lane.

## Questions

1. **Still open.** Is admitting `Q8_0` here the direction you want, or is the Prism scoping deliberate? This is the only one that needs you.
2. ~~Given caveat 3 — is page-backing every `Q8_0` runnable model acceptable, or would you prefer this scoped to `qwen35`?~~ — I read @karan68's "harmless, but worth stating outright" as the answer: documented rather than narrowed. Caveat 3 above now names the platforms explicitly. Say the word if you'd still rather have it gated; it is a small change.
3. ~~Would you want a macOS-gated resident-vs-CPU parity test in this PR?~~ — added (`metal_q8_0_resident_projection_matches_scalar_decode`), scoped to the resolver seam, since the Q8 GEMV kernels were already covered by `metal_q8_0_ksplit_gemv_matches_cpu_reference`.
4. ~~What evidence bar do you want?~~ — answered with `qa/ornith/G-PARITY-qwen35-metal-macos.md`: token-identical to the pinned llama.cpp `acd79d6` oracle, 4/4.

Happy to restructure or rescope — whichever fits how you want this lane to grow.

🤖 Filed by Claude Code (Fable) on behalf of @heskew

### DIFF
--- a/qa/ornith/G-PARITY-qwen35-metal-macos.md
+++ b/qa/ornith/G-PARITY-qwen35-metal-macos.md
@@ -1,0 +1,97 @@
+# ORNITH 9B — G-PARITY receipt, resident Metal lane (qwen35 vs llama.cpp acd79d6)
+
+**Gate:** G-PARITY — greedy token-identical vs the pinned llama.cpp oracle
+(`first_divergent_generated_token_index = -1`). **Result: PASS (4/4 prompts).**
+**Date:** 2026-08-06 · **Platform:** macOS 26.6 (25G72) arm64, Apple M4, 16 GB.
+**Camelid lane:** runnable → **resident Metal graph** (`qwen35`), wire mode
+(`CAMELID_METAL_F32Y=1`, `CAMELID_METAL_WIRE=1`), rustc 1.95.0, `--release`.
+**Oracle:** llama.cpp `acd79d6` (build 9632) — **reused, not re-run**; see Method.
+**Model:** `ornith-1.0-9b-Q8_0.gguf` (arch `qwen35`).
+
+Companion to `G-PARITY-qwen35-vs-llamacpp.md`, which certified the same gate on
+Windows x86_64 CPU. This one covers a different lane on a different platform.
+
+## Method (isolates the model forward from tokenization)
+
+Same method and the same four prompt token-ID arrays as the Windows receipt, so the
+two are directly comparable:
+
+```
+[[3710,369,279,6511,314,9338,30],
+ [727,73111,1393,1590],
+ [760,13600,314,3882,369],
+ [2427,310,4097,25,220,16,11,220,17,11,220,18,11]]
+```
+
+Greedy, `n_predict=20`, via the existing `ornith_qwen35_parity_gen` harness.
+
+**The oracle was not re-run for this receipt.** The reference token IDs are the ones
+already pinned in `G-PARITY-qwen35-vs-llamacpp.md` from the llama.cpp `acd79d6` run.
+That is the point of a pinned oracle — this receipt asks whether a *new Camelid lane*
+reproduces it, not whether llama.cpp is still self-consistent. Anyone re-certifying the
+oracle itself should start from the Windows receipt's method section.
+
+**The resident lane was confirmed engaged, not assumed.** `ornith_qwen35_parity_gen`
+exits 0 and prints plausible token IDs even when the Metal lane declines and falls back
+to CPU decode, so the run is only valid if stderr carries:
+
+```
+[qwen35] full Metal resident graph active (packed weights, attention,
+         gated-delta recurrence, FFN, logits, GPU greedy, and request sampling)
+```
+
+and does **not** carry `using hybrid fallback`. Both were checked for this run.
+
+## Results — generated token IDs, Metal resident lane vs pinned oracle
+
+| # | prompt | tokens | match |
+|---|--------|--------|-------|
+| 0 | `What is the capital of France?` | `[271,760,6511,314,9338,369,11751,13,271,3710,369,279,6511,314,9564,30,271,760,6511,314]` | ✅ identical |
+| 1 | `def fibonacci(n):` | `[198,262,413,307,2564,220,15,25,198,285,460,2958,198,262,4265,307,606,220,16,25]` | ✅ identical |
+| 2 | `The opposite of hot is` | `[8981,13,271,248068,198,90700,8340,25,271,16,13,220,2972,2014,53983,279,5952,64700,198,262]` | ✅ identical |
+| 3 | `Count to five: 1, 2, 3,` | `[220,19,11,220,20,13,4543,11,1092,3905,1727,30,271,760,1727,1324,303,279,8240,369]` | ✅ identical |
+
+All four token-identical → `first_divergent_generated_token_index = -1`. Byte-for-byte
+equal to the Windows CPU receipt's IDs, so CPU and Metal agree with each other as well
+as with the oracle.
+
+## What this earns
+
+- The **resident Metal lane** reproduces the pinned oracle for `qwen35` Q8_0 on Apple
+  Silicon. The gated-delta-net recurrence, causal conv1d, GQA head-repeat, state
+  orientation, partial NEOX mRoPE and gated attention are all correct **as executed on
+  the GPU**, not only in the CPU reference.
+- Admitting GGUF `Q8_0` into the resident lane with no repack does not perturb the
+  forward. Same 34-byte wire blocks, same argmax trajectory, all four prompts.
+- A macOS arm64 data point for a gate that previously existed only on Windows x86_64.
+
+This is lane/platform evidence for an existing row. It is **not** a support-contract
+change and claims nothing about rows outside `qwen35` Q8_0.
+
+## Caveats
+
+- **Not a benchmark.** No warmup, and load time is inside the wall clock, so the
+  numbers here are not comparable to `bench-generate` figures. For reference only: this
+  run took 14.7 s end to end; the same harness on the CPU hybrid lane took 221.8 s. A
+  cold process also pays one-time Metal pipeline compilation, which lands inside the
+  first prefill — see the `--warmup` note in the PR discussion.
+- **Wire mode must be set explicitly under the test harness.** `CAMELID_METAL_F32Y`
+  and `CAMELID_METAL_WIRE` default off; only the CLI turns them on via
+  `apply_default_fast_stack`. A bare `cargo test` therefore does *not* exercise this
+  lane, and wire-page weights require it.
+- **Known deviation carried from G-LOAD:** BPE pre-tokenizer `qwen35` is implemented as
+  qwen2's single-digit split, with `\p{M}` combining-mark folding deferred. The parity
+  method feeds token IDs, so this does not affect the model-forward result here.
+
+## Reproduce
+
+```sh
+CAMELID_ORNITH_GGUF=/path/to/ornith-1.0-9b-Q8_0.gguf \
+CAMELID_METAL_F32Y=1 CAMELID_METAL_WIRE=1 \
+CAMELID_PARITY_TOKENS='[[3710,369,279,6511,314,9338,30],[727,73111,1393,1590],[760,13600,314,3882,369],[2427,310,4097,25,220,16,11,220,17,11,220,18,11]]' \
+CAMELID_PARITY_NPREDICT=20 \
+cargo test --release --lib ornith_qwen35_parity_gen -- --ignored --nocapture
+```
+
+Check stderr for `full Metal resident graph active` before trusting the IDs. To produce
+the CPU control from the same binary, add `CAMELID_QWEN35_METAL=0`.
--- a/src/metal.rs
+++ b/src/metal.rs
@@ -25214,6 +25214,120 @@ mod tests {
         }
     }
 
+    /// Resolver gate for the resident `Q8_0` projection. The runnable loader admits
+    /// GGUF `Q8_0` with no repack, so what needs proving is the seam: that
+    /// `ResidentWeightFormat::Q8_0` plus 34-byte wire blocks resolve through the
+    /// production projection encoder — block-count math and the `n_tokens == 1` decode
+    /// vs batched-prefill split — to the scalar decoder's values. The Q8 GEMV kernels
+    /// themselves are covered by `metal_q8_0_ksplit_gemv_matches_cpu_reference`.
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn metal_q8_0_resident_projection_matches_scalar_decode() {
+        if !detect_metal_device().available {
+            return;
+        }
+        let kernel = metal_linear_kernel().expect("Q8 Metal pipelines");
+        let format = ResidentWeightFormat::Q8_0;
+        let block_elements = format.values_per_block();
+        let block_bytes = format.wire_bytes_per_block();
+
+        // One block per row is the degenerate K-split case (a single live block slot);
+        // eight crosses the stride. 19 rows is odd, so the row-tiling tail guard runs
+        // rather than only SIMD group zero.
+        for blocks_per_row in [1usize, 8] {
+            let input_width = blocks_per_row * block_elements;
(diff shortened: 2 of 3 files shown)


heskew:
Pushed a follow-up commit that closes the gap I flagged in question 3, plus some verification I owed on the blast-radius question.

**Tests (`eb24553`).** The parity evidence was previously commit-message-only. There are now five tests that run on any macOS host with **no GGUF and no GPU** — so they cannot silently pass the way an env-gated test does:

- `Q8_0` maps to `ResidentWeightFormat::Q8_0`
- the Prism low-bit formats still map unchanged (regression guard)
- K-quants are still refused, so admitting one stays a separate change
- a page-backed `Q8_0` tensor admits as `WirePages` with byte-identical contents
- an **owned** `Q8_0` tensor fails closed rather than being uploaded, since the resident lane wraps the allocation with `newBufferWithBytesNoCopy`

Fixtures are built from exact 34-byte wire blocks, so the 34/32 geometry that makes this a no-repack change is asserted rather than assumed. I also mutation-checked them: removing the `Q8_0` mapping fails exactly the two tests that guard it, while the three regression guards stay green.

Gate on the updated branch: `fmt --check` pass, `clippy --all-targets -D warnings` pass, `cargo test --release --lib` 1635 passed / 0 failed / 24 ignored.

**On the non-macOS half of caveat 3.** I said the `load_raw` change is wider than `qwen35`; I've now checked what that means off-macOS, and it is inert:

- `RawMatBytes::as_slice()` handles `Owned` and `WirePages` identically, and `wire_pages()` — the only accessor that distinguishes them — is `#[cfg(target_os = "macos")]`. Non-macOS code cannot observe the difference.
- The CUDA `qwen35` lane already handles `Q8_0` explicitly (`model.rs` ~4146 and ~4379) and reads through `as_slice()` before its own `widen_q8` / `to_vec()` copy, so page-backing changes nothing there.

I also checked for a third gate that might silently drop `Q8_0` and did not find one — `admit.rs`'s quant list already admits it, and `is_prism_low_bit` is genuinely Prism-specific and `cfg(not(target_os = "macos"))`.

**One thing I should surface that isn't in the original description.** `prism_metal_format` is also called from `par_matvec`/`par_matmul` (~202 and ~341), the CPU-lane hybrid path. Now that it returns `Some` for `Q8_0`, those call sites will try `try_prism_wire_matvec_f32` for page-backed `Q8_0` tensors — and `try_prism_wire_matmul_flat` rejects `Q8_0` in its format `matches!` (`metal.rs` ~12237) and returns `None` before doing any Metal work, so it falls through to the existing CPU path. Correct, and the cost is a couple of branches per matvec, but it is a no-op call that did not happen before.

That also means the hybrid path is now one whitelist entry away from running `Q8_0` projections on the GPU for the other runnable archs. I deliberately did **not** do that here — it is a different lane with different evidence needs — but happy to take it as a follow-up if it is a direction you want.

**Naming.** `prism_metal_format` now returns a format for a non-Prism quant, so the name is a bit of a misnomer. I refreshed the `RawMatBytes::WirePages` doc comment, which still said Q1/Q2-only, but left the function name alone to keep the diff focused. Say the word if you'd rather I rename it to something like `resident_metal_format` in this PR.

*Posted by Claude Code (Fable) on behalf of @heskew.*

heskew:
Pushed a third commit (`5f306ed`) closing the coverage gaps from another review pass, plus the real-model receipt I offered in question 4.

**The loader change is now actually covered.** The previous tests built `WirePages` directly, so nothing failed if `Q8_0` were dropped from `load_raw`'s page-backing whitelist — it would quietly fall back to owned backing, the resident lane would refuse it, and the model would decode on CPU with no error. That whitelist is now `wants_page_backing()`, with tests that are deliberately **not** macOS-gated, since backing selection is platform-independent.

**Resolver parity test, scoped to the seam that was actually uncovered.** I looked for what this PR could plausibly break and concluded it was not the kernel math: `metal_q8_0_ksplit_gemv_matches_cpu_reference` and the batched wire test already check `encode_q8_matmul_f32y`/`_batched` against a CPU reference on exactly these 34-byte blocks, and this PR touches no kernel. What was untested was the resolver — that `ResidentWeightFormat::Q8_0` plus wire pages reach the right kernel with the right block count. `metal_q8_0_resident_projection_matches_scalar_decode` drives `encode_resident_matmul_f32` at 1 and 8 blocks/row × 1 and 2 tokens, so both the decode and batched-prefill arms, against `decode_q8_0_tensor`. It sits next to the Prism equivalent in `metal.rs`, which also lets it use `values_per_block`/`wire_bytes_per_block` rather than restating 34/32.

**G-PARITY receipt on the resident Metal lane** — `qa/ornith/G-PARITY-qwen35-metal-macos.md`, a companion to the existing Windows CPU receipt in the same format. Greedy generation is token-identical to the pinned llama.cpp `acd79d6` oracle on all four of that receipt's prompts, `first_divergent_generated_token_index = -1`. To be explicit: **the oracle was not re-run** — the reference IDs are the ones already pinned in `G-PARITY-qwen35-vs-llamacpp.md`. The question this answers is whether a new lane reproduces the pinned oracle, not whether llama.cpp is still self-consistent.

Two traps I hit and documented there, since both produce a convincing false pass:

- `ornith_qwen35_parity_gen` exits 0 and prints plausible token IDs **even when the Metal lane declines and falls back to CPU decode**. My first run did exactly that. The receipt tells you to confirm `full Metal resident graph active` in stderr before trusting the IDs.
- `CAMELID_METAL_F32Y` / `CAMELID_METAL_WIRE` default off outside the CLI — only `apply_default_fast_stack` turns them on — so a bare `cargo test` does not exercise this lane at all, and wire-page weights require it.

I also rescoped the fails-closed test comment: it asserts the `Owned` variant is refused, not the `newBufferWithBytesNoCopy` alignment invariant it claimed. That is `WirePages`' contract, already covered by `wire_mmap_nocopy_buffer_gpu_reads_file_bytes`.

**On question 2, I deliberately did not act.** The suggestion was to gate the `load_raw` change to macOS + `qwen35`. It is mechanically easy — `arch` is in scope — but `load_raw` has no arch awareness today and the neighbouring Prism whitelist is neither platform- nor arch-gated, so special-casing `Q8_0` would make it the odd one out. Since that is a blast-radius judgement about your loader rather than a correctness fix, I would rather have your answer than pre-empt it. It is a three-line change whenever you want it.

Gate on the updated branch: `fmt --check` pass, `clippy --all-targets -D warnings` pass, `cargo test --release --lib` **1638 passed / 0 failed / 24 ignored**, `check-public-scrub.sh` pass. Mutation-checked: dropping `Q8_0` from `wants_page_backing` fails only the predicate test; flipping `q8_wire` fails only the projection test.

*Posted by Claude Code (Opus 5) on behalf of @heskew.*

karan68:
## Review of [#618](https://github.com/timtoole02/Camelid/pull/618)

**Real finding:** `prism_metal_weight()` requires both `wants_page_backing(tt)` (checked first) and `prism_metal_format(tt).is_some()` (checked second) to admit a type. These are two separately-maintained match arms over `GgufTensorType` with nothing tying them together. Add a format to only one — e.g. the K-quant follow-up in caveat 5 — and you reproduce this exact bug: tensor stays `Owned`, fails "not page-backed," silently falls back to CPU.

Fix: derive `wants_page_backing` from `prism_metal_format(tt).is_some()` (make the latter a free fn over `GgufTensorType`), or add one test asserting the two agree for every variant.

**Minor:** caveat 3 says blast radius is "wider than qwen35" — worth being explicit that `wants_page_backing` isn't macOS-gated, so this also page-backs every `Q8_0` tensor on Linux/Windows, not just other macOS models (harmless — `as_slice()` is identical either way — but worth stating outright).

Everything else checks out: block math is correct and cross-checked against `decode_q8_0_tensor` rather than re-deriving the same constants, `is_prism_low_bit` correctly excludes `Q8_0`, `WirePages` is genuinely cross-platform. No other issues found.

heskew:
Thanks @karan68 — that's the best finding on this PR, and it caught something I should have seen in my own diff.

**Fixed in `07201707` by removing the second list rather than testing it.** You offered two options; I took the structural one. `prism_metal_format` is now a free fn over `GgufTensorType` and page-backing is derived from it:

```rust
fn wants_page_backing(tt: GgufTensorType) -> bool {
    resident_metal_format(tt).is_some()
}
```

A test asserting two lists agree still lets someone add a type to one, get a red test, and "fix" it by editing the test. With one list there is nothing to disagree with — admitting a K-quant later is one arm in one match and the backing follows. Given caveat 5 proposes exactly that follow-up, I'd rather the trap not be there when someone walks into it.

The rename falls out of the same change: the fn returns a format for `Q8_0`, so the Prism-scoped name had stopped describing it. That also closes a naming nit from an earlier review pass.

**Knock-on:** the fn drops its macOS `cfg`, because `wants_page_backing` never had one — the loader makes the same backing choice on every target and `ResidentWeightFormat` isn't gated either. So the format mapping is no longer macOS-only, and its tests moved out of the macOS-gated module. Three tests that only checked the mapping folded into one table pinning the mapping and the backing together. Net three fewer tests, but the mapping is now guarded on **every** platform instead of macOS alone — which matters here, since the predicate's blast radius is precisely the non-macOS targets.

**On your minor point** — caveat 3 now says outright that the predicate is not macOS-gated and that every `Q8_0` runnable tensor page-backs on Linux and Windows too, with the reasoning for why that's unobservable (`as_slice()` identical, `wire_pages()` macOS-only, CUDA lane reads through `as_slice()`). I also took it as answering my open question 2: documented rather than narrowed, so I have **not** added an arch gate. Easy to add if you disagree.

Gate on the updated branch: `fmt --check` pass, `clippy --all-targets -D warnings` pass, `cargo test --release --lib` 1635 passed / 0 failed / 24 ignored, `check-public-scrub.sh` pass. Mutation-checked: dropping `Q8_0` from `resident_metal_format` fails exactly the two tests that guard it, regression guards green. I still can't run the `--all-features`/CUDA leg or a Linux build locally — this change is `cfg`-free so macOS compiles the same code, but the cross-platform legs want CI.

That's every thread I can close from my side. Question 1 — whether admitting `Q8_0` into this lane is the direction you want at all — is genuinely yours, and I'll leave the branch alone rather than keep adding to it while that's open.

*Posted by Claude Code (Opus 5) on behalf of @heskew.*

PR #619, opened 2026-08-06
feat(fabric): resident HTTP proxy (camelid fabric serve)

karan68:
## What

`fabric run` sends exactly one request per CLI invocation. This adds `camelid fabric serve`: a resident HTTP server so a real client can point at one address instead of the operator invoking the CLI per request.

`POST /v1/chat/completions` is placed and forwarded through the exact same `Fabric::dispatch` path `fabric run` uses. The backend's status and body are returned verbatim, tagged with `x-camelid-fabric-node` / `-reason` / `-affinity-lost` response headers. Streaming is refused with 400 before any node is touched, matching the existing `forward()` contract.

Merged up to current `main`, so the diff is exactly this feature: **7 files, +890/-5**.

## Files

- `src/fabric/server.rs` — router, `bind`/`serve_on` (split so a caller can read the real port before serving), the non-loopback bind guard, error mapping (routing failure → 503, unsupported input → 400, transport/decode failure → 502, refused body → 400/413), unit tests.
- `src/fabric/policy.rs` — `RouteReason::as_str()`.
- `src/api/mod.rs` — re-export `DEFAULT_MAX_REQUEST_BODY_BYTES` so the proxy bounds bodies at the same size the node does.
- `src/main.rs` — `FabricAction::Serve`, reusing the existing `route_mode()` and `fabric_bearer()` helpers.
- `tests/fabric_serve.rs` — 6 real-socket end-to-end tests: real TCP bind, real HTTP client, real stub nodes (same pattern as `fabric_end_to_end.rs`), including a concurrency test.
- `docs/CONFIGURATION.md` — a fabric proxy section stating the bind rule and the limits below.

## A real bug found in self-review

`Fabric::dispatch` is synchronous socket I/O that can legitimately run for the whole `forward_timeout` (minutes, for a real generation). The handler originally called it directly on the async task. `#[tokio::main]` defaults to worker threads = CPU count, so once concurrent in-flight requests reached that count, every async worker was parked in a blocking call and the **entire server** stopped answering anything until one finished — defeating the point of a resident proxy.

Fixed by running the dispatch inside `tokio::task::spawn_blocking`. Ablated with a test that pins the runtime to one worker thread, fires 4 concurrent requests at 4 independently-slow stub nodes, and asserts they complete together rather than serially. Confirmed the test fails without the fix and passes with it.

## Review round: four defects fixed

**1. No bind guard.** `fabric serve --addr 0.0.0.0:8282` started happily and exposed every node in the fabric, unauthenticated. `bind()` now refuses a non-loopback listener *before* it opens the socket, unless the operator acknowledges it with `--allow-unauthenticated-remote` / `CAMELID_ALLOW_UNAUTHENTICATED_REMOTE`. This mirrors `distributed::check_worker_bind_policy` and the rule `api::server::ServerPolicy::resolve` applies, including the env alias. `serve()` was removed so the guard cannot be skipped by forgetting to call it.

**2. Error-shape divergence.** `Json(body): Json<Value>` let axum answer a malformed body with a plain-text 400 — the one error a client is most likely to hit was the only one it could not parse. The handler now takes `Result<Json<Value>, JsonRejection>` and answers in the same `{"error":{"message":…,"type":"fabric_error"}}` shape as every other path in the file, mirroring `api::malformed_json_error` (413 for an oversize body, 400 otherwise). The existing test asserted only the status; it now asserts the shape, like its siblings.

**3. Request body limit.** No `DefaultBodyLimit` layer meant axum's 2 MiB default applied, eight times stricter than the node being proxied. The router now uses the node's own `DEFAULT_MAX_REQUEST_BODY_BYTES` (16 MiB), re-exported rather than copied so the two cannot drift.

**4. Timing-fragile concurrency test.** `StubNode`'s accept loop was strictly serial and `serve_once` slept while holding it, so a health probe could queue behind another request's completion — ~600 ms worst case against a 700 ms budget. The stub now handles each connection on its own thread (which is what its "health is never delayed" comment already claimed), and the ratio widens to 500 ms delay / 1200 ms threshold. Serialized is ~2000 ms, so the ablation stays sharp with real margin.

**Credential forwarding, now that `main` supports it.** `Fabric::with_bearer` landed on `main` after this branch was cut, and `fabric status|route|run` all take `--bearer`. `fabric serve` was the only one without it, which made the resident proxy the one surface that could not talk to a node started with `--api-key`. It now takes the same flag with the same `CAMELID_API_KEY` fallback through `fabric_bearer()`.

Also from the smaller notes: the `listening on` line moved after the bind, so a refused or already-taken address no longer prints a line it did not earn, and it now reports the address actually bound; `RouteReason::as_str()` pins the strings that reach clients — those values were `#[derive(Debug)]` output, and they are already public through `fabric route|run --json` on `main`, so they are pinned to their current spelling and used at all three sites; and the router holds an `Arc<Fabric>` rather than reallocating the node list on every request.

### Negative control

Reverting only the body-limit layer and the `JsonRejection` arm makes exactly the three new body tests fail — every one of them panicking while parsing the response, because the body came back as plain text.

### Live receipt, built binary

| Case | Result |
|---|---|
| `--addr 0.0.0.0:8299` | refused with the named message, exit 1, **no** listening line |
| same, `--allow-unauthenticated-remote` | past the guard (then failed at bind, os error 10049) |
| `--addr 127.0.0.1:8299` | `fabric serve listening on 127.0.0.1:8299` |
| same port while held | os error 10048, **no** listening line |
| malformed body | `400` `application/json` `{"error":{"message":"Failed to parse the request body as JSON","type":"fabric_error"}}` |
| 4 MiB valid body | `503` `fabric_error` — reached placement, not refused |
| 17 MiB body | `413` `fabric_error` |

## Gate

Run on the merged head, against current `main`:

- `cargo test --lib`: **1676 passed, 0 failed**, 84 ignored
- `fabric_end_to_end` 14 passed, `fabric_serve` 6 passed, over real sockets
- `cargo test --bin camelid`: 36 passed
- `cargo fmt --all -- --check` clean
- `cargo clippy --all-targets -- -D warnings` clean (0 warnings)
- `cargo doc --no-deps` exit 0, no new rustdoc warnings
- `scripts/check-public-scrub.sh` clean
- `git diff --check` clean, every touched file `i/lf w/lf`

Windows x86_64 only; the macOS / Linux / aarch64 legs are CI's.

## Deliberate limits

- **The proxy authenticates no client.** It has no key of its own; `--bearer` is the token it presents *to its nodes*. The acknowledgement flag is the only way to bind it to a routable address, and it should only be used behind something that does authenticate.
- **A client disconnect does not cancel a dispatch.** `spawn_blocking` is not cancellable, so the closure runs to completion, holding one blocking thread for up to `forward_timeout`. Bounded, but real; it belongs with the capacity-limit follow-up.
- **No `GET /v1/models`.** Some OpenAI-compatible SDKs probe it. A follow-up, not a blocker.
- Streaming, a background health-refresh loop, and per-node capacity limits — the same follow-ups already logged against #616.


timtoole02 (commented):
## Overview

Adds `camelid fabric serve` — a resident HTTP proxy that turns the one-shot `fabric run` CLI into a long-lived front door. `POST /v1/chat/completions` is placed and forwarded through the exact same `Fabric::dispatch` path, with the backend's status/body returned and `x-camelid-fabric-node` / `-reason` / `-affinity-lost` tagged onto the response.

Scope is now clean: with #616 merged (`79493f67`), the diff against current `main` is **658 added lines, 0 deleted, 4 files** — no leftover #616 content, merges cleanly.

The design is good. Reusing `dispatch` rather than reimplementing placement is the right call; the `serve`/`serve_on` split so tests can bind port 0 is a nice touch; and the `spawn_blocking` fix described in the PR body is a real bug caught in self-review, correctly diagnosed and correctly ablated. Error mapping (503 route / 400 unsupported / 502 transport) is sensible.

## Issues

### 1. No bind guard for unauthenticated remote exposure — the main server has one, this doesn't

`src/api/server.rs:194` deliberately refuses to start:

```rust
if !addr.ip().is_loopback() && key.is_none() && !options.allow_unauthenticated_remote {
    return Err(...)
}
```

`fabric serve` accepts any `--addr` with no auth layer and no equivalent check. `camelid fabric serve --addr 0.0.0.0:8282 --node a=...` starts happily and exposes **every node in the fabric** to the network, unauthenticated. The loopback default is good, but the guard is what the rest of the project relies on. Suggest mirroring the check (and an `--allow-unauthenticated-remote` escape hatch) before this ships.

Related, and worth stating in the PR description: `src/fabric/http.rs:224` writes only `Host` / `Accept` / `Connection` / `Content-Type` / `Content-Length` — no `Authorization` or `x-api-key`. So a node started *with* `--api-key` will pass the `/v1/health` probe (auth-exempt) and then 401 every forwarded request. That's inherited from #616, not introduced here, but a resident proxy is exactly the surface where an operator will hit it.

### 2. Malformed JSON breaks the error-shape contract, and the test hides it

`Json(body): Json<Value>` means axum's `JsonRejection` returns a **plain-text** 400 body — not the `{"error":{"message":…,"type":"fabric_error"}}` shape every other path in the file returns. The project convention is explicit and used in ~9 handlers (`src/api/mod.rs:4208` and others):

```rust
payload: std::result::Result<Json<T>, JsonRejection>
// then: Err(error) => return malformed_json_error(error),
```

Note the PR's own unit test `a_malformed_body_is_rejected_before_it_reaches_the_fabric` asserts only `response.status()` — it would have caught this had it asserted the body shape like its three siblings do.

### 3. Request body limit is 2 MB, 8× stricter than the node it proxies

No `DefaultBodyLimit` layer, so axum's 2 MB default applies. `DEFAULT_MAX_REQUEST_BODY_BYTES` is `16 * 1024 * 1024` (`src/api/server.rs:27`) and the main router sets it explicitly (`src/api/mod.rs:2534`). A long conversation the node would serve fine gets 413'd at the proxy — with a plain-text body, per issue 2.

### 4. The concurrency test is more timing-fragile than its comment claims

The comment predicts "~300ms plus scheduling overhead" against a 700ms threshold, but `StubNode`'s accept loop is strictly serial — `for stream in listener.incoming() { serve_once(...) }` — and `serve_once` sleeps `completion_delay` *while holding the thread*. Each of the 4 requests probes **all 4 nodes**. So a request whose probe of node-0 arrives after request-0's completion has started blocks in the accept backlog for the full 300ms, then spends its own 300ms: **~600ms worst case against a 700ms budget**, before CI jitter.

The stub comment says "Health is never delayed," which is true of the sleep but not of the serialization. Widening the ratio (500ms delay / 1200ms threshold) would keep the ablation just as sharp with real margin.

## Smaller notes

- **`println!` fires before the bind.** `src/main.rs` prints `fabric serve listening on {addr}` and *then* calls `serve()`, which binds. On `EADDRINUSE` the operator sees a "listening" line followed by an error. Move it into `serve_on` after `local_addr()`, or print after bind.
- **`{:?}` on `RouteReason` is a wire format.** `insert(out, "x-camelid-fabric-reason", &format!("{:?}", decision.reason))` makes a `#[derive(Debug)]` variant name into a client-visible header — renaming `LeastLoaded` silently changes the HTTP API. A small `as_str()` on `RouteReason` would pin it.
- **Client disconnect leaks a blocking thread.** `spawn_blocking` isn't cancellable; if the client hangs up, the closure runs to completion, holding one of tokio's 512 blocking threads for up to `forward_timeout` (300s default) and letting the node keep generating. Bounded, but worth logging next to the capacity-limit follow-up.
- **`Fabric` is cloned per request** (via axum state clone), reallocating the `Vec<NodeSpec>`. Negligible, but `Arc<Fabric>` is free to do.
- **No `/v1/models`.** For "a real client can point at one address," OpenAI-compatible SDKs commonly probe it. Reasonable follow-up, not a blocker.
- **Docs**: this adds a user-facing CLI subcommand with no docs change. #616 didn't add fabric docs either, so it's consistent — flagging only because the surface is now operator-facing.

## Verdict

Solid, well-tested work with unusually good commentary — the `spawn_blocking` diagnosis and its ablation test are the standout. Test coverage is real (real sockets, real stubs, verbatim-passthrough and 5xx-passthrough both covered).

I'd hold merge on **#1** (unauthenticated remote bind) and **#2** (error-shape divergence); #3 and #4 are quick follow-ups in the same pass. Everything else is optional polish.

One caveat: this review is from reading the diff and surrounding code against `origin/main` — I did not build or run the suite locally, so the gate results (1637 tests, clippy, fmt, scrub) are taken as reported.

karan68:
Thank you — this was a sharp review. I checked every claim against the source before touching anything, and **all four are real**. All four are fixed in `f67dcdd3f`, along with three of the smaller notes. PR description rewritten to match.

### 1. Bind guard — fixed

Confirmed: `fabric serve` accepted any `--addr` with no auth layer and no equivalent of the `src/api/server.rs:194` check.

New `fabric::server::bind()` refuses a non-loopback listener **before it opens the socket**, with `--allow-unauthenticated-remote` / `CAMELID_ALLOW_UNAUTHENTICATED_REMOTE` as the escape hatch. I mirrored `distributed::check_worker_bind_policy` (`src/distributed.rs:909`) rather than the API server directly — it is the closer precedent (same wording shape, same env alias, and it already reuses that acknowledgement across surfaces). I also deleted `serve()` so the guard cannot be skipped by forgetting to call it; `bind()` + `serve_on()` is now the only pair.

Live, on the built binary:

```
$ camelid fabric serve --node a=127.0.0.1:9 --addr 0.0.0.0:8299
Error: refusing unauthenticated non-loopback listener 0.0.0.0:8299; the fabric proxy has no
authentication of its own, so this would expose every configured node to the network. Bind a
loopback address, or explicitly acknowledge the risk with --allow-unauthenticated-remote
exit=1
```

Three unit tests cover loopback (v4 and v6), the refusal (kind + the message naming the flag), and the acknowledged case.

On the related point: also confirmed. `/v1/health` is auth-exempt (`src/api/server.rs:292`), and `fabric/http.rs` writes only `Host` / `Accept` / `Connection` / `Content-Type` / `Content-Length`, so an `--api-key` node passes the probe and then 401s every forward. I did **not** build credential forwarding — it needs a per-node credential design, and this PR is not the place to invent one. It is now stated in the PR description, in the module docs, and in `docs/CONFIGURATION.md`.

### 2. Error-shape divergence — fixed, and you are right that my own test hid it

The handler now takes `Result<Json<Value>, JsonRejection>` and answers in the file's own `fabric_error` shape, mirroring `api::malformed_json_error` (413 for `PAYLOAD_TOO_LARGE`, 400 otherwise). `a_malformed_body_is_rejected_before_it_reaches_the_fabric` now asserts the body shape, like its three siblings — status alone was exactly the gap you identified.

```
$ curl -i -X POST .../v1/chat/completions -H 'Content-Type: application/json' --data-binary 'not json'
HTTP/1.1 400 Bad Request
content-type: application/json

{"error":{"message":"Failed to parse the request body as JSON","type":"fabric_error"}}
```

### 3. Body limit — fixed

Confirmed 2 MiB by axum default. The router now layers `DefaultBodyLimit::max(DEFAULT_MAX_REQUEST_BODY_BYTES)`. Rather than copy the 16 MiB literal, I made `api/mod.rs` re-export the constant (`mod server;` is private, so a `pub(crate) use` is the surgical way to share it) — the two cannot drift now.

```
4 MiB valid body  -> 503 fabric_error   (reached placement; was 413 text/plain)
17 MiB body       -> 413 fabric_error   (in the right shape, per issue 2)
```

### 4. Concurrency test — fixed, and the comment is now true

Your reading of the accept loop is correct: `for stream in listener.incoming()` is serial, `serve_once` sleeps holding it, and a probe arriving behind a completion pays the full delay — ~600 ms against a 700 ms budget.

I did both halves. The stub now handles each connection on its own thread, which is what "Health is never delayed" already claimed and is what a real node does; and the ratio widens to 500 ms delay / 1200 ms threshold as you suggested. Serialized is ~2000 ms, so the ablation stays sharp with real margin on both sides. Measured: 0.51 s for the whole file.

### Smaller notes

- **`println!` before the bind** — fixed. `main.rs` now binds, then prints `listener.local_addr()`. So a refused or already-taken address prints no listening line, and the line reports what was actually bound. Verified live: a second bind on a held port errors with os error 10048 and no listening line.
- **`{:?}` on `RouteReason`** — fixed with `as_str()`, used at all three sites. One caveat worth flagging: those strings are already public on `main` through `fabric route|run --json`, which #616 shipped. So I pinned them to their **current** spelling rather than tidying them to snake_case — this PR should not silently change output that already exists. A test asserts the three values, so a variant rename can no longer move either surface.
- **`Fabric` cloned per request** — fixed, `Arc<Fabric>` in the router state.
- **Client disconnect leaks a blocking thread** — correct, and not fixable without making the forward path cancellable. Documented in the module docs next to the capacity follow-up rather than left implicit.
- **No `/v1/models`** — agreed it is worth having; left as a follow-up and named in the PR description under deliberate limits, since aggregating a model list across nodes is its own design question.
- **Docs** — added, since you are right that the surface is now operator-facing. `docs/CONFIGURATION.md` gains a "Fabric proxy policy" section next to the existing "Production HTTP policy", covering the bind rule and the two limits above.

### Negative control

Reverting only the body-limit layer and the `JsonRejection` arm makes exactly the three new body tests fail — each one panicking while parsing the response, because the body came back as plain text. That is the defect you described, reproduced.

### Gates on `f67dcdd3f`

`cargo test --lib` 1643 passed / 0 failed / 84 ignored (74 `fabric::` unit tests, up from 68) · `fabric_end_to_end` 11 · `fabric_serve` 6 · `--bin camelid` 31 · `cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` 0 warnings · `cargo doc --no-deps` exit 0 with no new warnings · `check-public-scrub.sh` clean · `git diff --check` clean · all touched files `i/lf w/lf`. Run on Windows x86_64; the macOS/Linux/aarch64 legs are CI's.

karan68:
Follow-up on my reply above, with two corrections and one addition. Head is now `c8d81d11`.

**The scope claim needed fixing, and now it is real.** The PR page was diffing against a base snapshot from before #616 merged, so it showed 13 files / +3728 — the "clean scope" you and I both described was true of a local `git diff` but not of what a maintainer sees. Merged `main` in; GitHub now reports **7 files, +890/-5** against `8601bf6b8`, which is exactly this feature.

**I was wrong about credentials — and `main` had already fixed it.** My reply said forwarded requests carry no credentials and that closing it needed a per-node credential design. `Fabric::with_bearer` landed on `main` after this branch was cut, and `fabric status|route|run` all take `--bearer` with a `CAMELID_API_KEY` fallback. `fabric serve` was simply the one subcommand that had not been given it, which made the resident proxy the only surface that could not talk to an `--api-key` node — the exact operator hazard you flagged.

Fixed in `c8d81d11`: same flag, same `fabric_bearer()` resolution, so a shell configured for one node needs no second setting. The module docs and `docs/CONFIGURATION.md` said "forwards no credentials"; that was true when written and is not any more, so both now say what the token does and does not cover — it authenticates the proxy to its nodes, never a client to the proxy.

```
$ camelid fabric serve --help
      --bearer <TOKEN>
          Bearer token for nodes started with an API key. Falls back to CAMELID_API_KEY.
          Without it, such a node observes as ready and then answers every forwarded
          request with 401
```

**Gates re-run on the merged head**, since 45 commits of `main` came in: `cargo test --lib` 1676 passed / 0 failed / 84 ignored · `fabric_end_to_end` 14 · `fabric_serve` 6 · `--bin camelid` 36 · `cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` 0 warnings · `cargo doc --no-deps` exit 0 · `check-public-scrub.sh` clean · `git diff --check` clean · all touched files `i/lf w/lf`. The merge was conflict-free.

Everything else in my previous reply stands unchanged.

PR #633, opened 2026-08-08
UI polish: design-review fixes across the frontend and desktop app

timtoole02:
## Summary

- Implements the findings from a full UI design review of the frontend and desktop app (136 verified findings: 16 high, 68 medium, 52 low), across two passes.
- Removes internal engineering vocabulary from user-facing chrome, consolidates six hand-rolled button systems onto the shared `ui/Button`, unifies the type scale and the signature uppercase label, and aligns every page title with its nav label.
- Fixes real defects found along the way: a CSS class referenced but never defined, an undefined design token silently stripping borders, a reversed theme-toggle announcement, and a Chat history card layout that broke as soon as real conversations existed.
- Brings the contrast checker up to cover computed chip/button fills; all 258 checked pairs now pass WCAG AA in both themes with an empty exceptions list.

## Why this change exists

The UI had drifted: parallel button systems, a signature label treatment implemented six different ways, pages whose titles disagreed with the nav item that opened them, and spec vocabulary leaking into the product surface (a permanent `NO MATCHING COMPATIBILITY.MD ROW` chip in the top bar, `NO CONTRACT ROW` repeated six times in the generation controls, a `see BACKEND_ASKS.md #3` citation in Settings). Individually small, together they read as an internal tool rather than a product.

Three of the fixes are genuine bugs, not polish:

- `.cxv-button` was referenced in three places but defined in no stylesheet, so those primary actions rendered as unstyled browser text. They now use the real `Button` component.
- `workspace.css` referenced `var(--color-border)`, which does not exist in `tokens.css`, silently removing nine borders across the Workspace view.
- The theme toggle's announced next theme was the reverse of the actual cycle in every state, so its tooltip and screen-reader label both lied.

Chat history deserves its own note: those cards had only ever been exercised with an empty conversation list. With real threads, `.cxv-card__actions` being `flex: none` let four buttons claim the whole footer, squeezing the meta column until each span wrapped inside itself ("6 / replies", "Jul / 20") and the delete button overflowed past the card edge. Previews also printed raw markdown (`**bold**`, ` ```cpp ` fences) because `formatPreview` only collapsed whitespace, and the same list showed both display names and raw catalog ids.

## Validation

- [x] `git diff --check` — clean for this change. Pre-existing CRLF files in the tree (`chat.css`, `MessageTurn.jsx`, `SettingsView.jsx`, `evidence.css`, `ParityReceipt.jsx`) still report trailing whitespace on added lines; that is the repo's existing convention for those files and was deliberately left alone.
- [x] `cargo fmt --all -- --check`
- [ ] `cargo test --all-targets --all-features` — **not run locally**; the build tree was cleaned to reclaim disk on the dev machine, so a full rebuild was not performed. Rust changes here are confined to `camelid-desktop` (splash retry command, window background) and were tested green before the rebase onto v0.6.0. Deferring to CI.
- [x] `cd frontend && npm run build`
- [x] `npm run smoke:contrast` — 258/258 pairs pass AA in both themes
- [x] Manual verification against a running engine: all 13 views toured in both themes at desktop and mobile widths, plus keyboard focus, modal focus traps, command palette, and the populated Chat history grid.

### Notable review points

- Four state-colour tokens were nudged slightly (e.g. `--color-error` `#b3261e` → `#ac251d`) so state-coloured text clears AA over its own translucent chip tint. Solved against the real computed fills rather than the flat token pairs.
- Three dead Observatory components, ~250 lines of orphaned CSS, and a hazardous wildcard selector are removed.
- A `light-dark()` token refactor was attempted and deliberately reverted: it would have raised the browser floor to Safari 17.5 / Chrome 123 for a UI served to arbitrary browsers. `tokens.css` carries a comment so it is not reintroduced by accident.
- Deliberately out of scope: the macOS overlay title bar (needs paired Tauri + frontend changes shipped together) and an in-app engine restart action (needs a backend command that does not exist yet).

## Public-repo privacy check

- [x] No private hostnames, SSH commands, user home paths, key paths, local validation details, raw SSH stderr, or raw infrastructure failure output are included


timtoole02:
Pushed a sidebar layout fix plus a merge of latest `main` (which brings in #634's support-label fix).

**Regression found**, A/B'd against the shipped v0.6.0 binary at identical 1440x900:

| | shipped v0.6.0 | #633 as pushed | with this fix |
|---|---|---|---|
| nav items below fold | 0 | **4** (System, API, Compatibility, Cluster) | 1 (Cluster, with SYSTEM visible) |
| rail scrollHeight | 731px | 975px | 817px |

Grouping the rail into four labelled sections added ~166px. Item height, font size and item padding are byte-identical to shipped, so the growth is purely structural. The `SYSTEM` label missed the cut by **one pixel** (top 836 vs rail bottom 835).

Fix reclaims the section label's vertical padding (the preceding section's 20px margin already separates the groups, so it was doubling an existing gap), drops the final section's trailing margin, and applies the end-fade already used on `.rail__rail-nav`. **The 20px section rhythm is unchanged.**

All 13 destinations cannot fit at 900px without abandoning the grouping — the three most aggressive tightenings combined recover 198px of the 244px needed — so the affordance is the correct answer rather than a compromise.

Verified on Windows: all 13 views render, no crashes, no horizontal overflow at 375px or 768px, theme cycles correctly, chat works end to end, and 7 frontend smokes pass. Smoke assertion counts across this PR are **+21 vs main with none reduced**.

timtoole02:
Second fix pushed (`1310ba91`): the splash's **Retry button was dead and left the app on a permanent spinner**.

`retry_startup` is registered in `generate_handler!` and wired to the Retry button, but was declared in neither place that authorizes it — `build.rs`'s `AppManifest` command list (which generates `allow-retry-startup`) nor `capabilities/default.json`.

The local splash is **not** exempt from the ACL: declaring any command makes `Resolved::has_app_acl` true, and tauri's gate `plugin_command.is_some() || has_app_acl_manifest || !is_local` then enforces it for local origins too. `startup_snapshot` is the proof — it's called only by the splash and still needs `allow-startup-snapshot`.

So Retry hid the error pane and the Technical details, showed the spinner, then died in the ACL with the rejection swallowed by `.catch(() => {})`. `start_engine` never ran, no further `engine-status` event was emitted, and the user was left watching a spinner forever with the diagnostics already erased — on the screen they reach *after* the engine failed. Only recovery was killing the app.

CI could not catch this: the mismatch compiles, links, and passes, and `camelid-desktop/tests/` only covers installer hooks. Verified by running `build.rs` and confirming `permissions/autogenerated/retry_startup.toml` is now produced and `allow-retry-startup` resolves in the generated schemas — neither existed before.

---

**One finding I am reporting rather than fixing.** During streaming, auto-follow is now released only by a *gesture* (`wheel`/`touchmove`/`keydown`). Dragging the scrollbar or clicking its track emits `scroll` and nothing else, so those users stay pinned to the bottom and "Jump to latest" never appears. v0.6.0 released on any scroll past 260px (`const follow = distanceFromBottom < 260`). Wheel/touch/keyboard users — the large majority — are unaffected.

I had a one-line fix for it (restore the distance release alongside the gesture release) but **reverted it**: I could not get a clean isolated verification, because generation kept finishing before I could measure, and I was not willing to ship an unverified behavioural change to streaming chat unattended. Your call whether it is worth a follow-up.

PR #651, opened 2026-08-12
feat(fabric): relay streaming responses through the resident proxy

karan68:
## The problem

`camelid fabric serve` refused any request carrying `stream: true` with a 400. Chat clients stream by default, so in practice a stock OpenAI client could not talk to the fabric at all — it would connect, send, and get a 400 back. The resident proxy shipped in #619 was only usable by something written specifically for it.

## What this does

The proxy now relays server-sent events from the node it placed the request on.

The relay is **byte-transparent**. Nothing here parses `data:` lines or looks for `[DONE]`; the body is de-chunked and forwarded verbatim. An event field this fabric has never heard of reaches the client unaltered, and a change to the engine's chunk shape cannot break the proxy.

- **`http`** — `ChunkDecoder` decodes a chunked body *as it arrives*, because a streamed body's frames straddle socket reads while the existing `dechunk` needs the whole body up front. `open_stream` reads only the response head and leaves the body to `ResponseStream::next_chunk`. Head parsing is now shared with `parse_response` rather than written twice, and `connect_and_send` is shared with `request`.
- **`forward`** — `forward_streaming` returns a `StreamOutcome`. A node that answers non-2xx, or with anything other than `text/event-stream`, is buffered and handed back as an ordinary answer, so a refusal arrives as a readable body instead of an empty stream.
- **`server`** — the response head travels over a oneshot, so the status and the `x-camelid-fabric-*` placement headers are settled before the client is committed to reading a stream. The body travels over a bounded channel, which is what applies backpressure to the node when a client reads slowly. Only `content-type` and `cache-control` are relayed from the node; the rest describe *its* connection, and this response is framed independently of it.

`fabric run` is unchanged: it returns one complete answer by construction and still refuses `stream: true`, now saying to use `fabric serve` instead.

## A property this adds

**A client that hangs up now stops the node generating.** The buffered path cannot do this — blocking socket I/O is not cancellable, so a dispatch runs until the node answers or the timeout expires. On the streaming path the next send fails and the node's socket is dropped with it. Measured below.

## Timeouts

`forward_timeout` bounds two waits on a streaming request: the wait for the response head, and any later gap in which the node sends *nothing*. Every event resets the second, so a long generation is never cut short while a wedged node still fails on schedule. This mirrors the read-headers-deadline plus per-step-deadline shape already used by `chat::client`.

## Evidence

**Gates** (on the merged head, after main moved 94 commits underneath): `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean; `fabric::` unit 90 pass; `fabric_end_to_end` 14 pass; `fabric_serve` 12 pass; `--bin camelid` 40 pass; `check-public-scrub.sh` exit 0; all touched files `i/lf w/lf`; `git diff --check` clean.

**Two real engines, not stubs.** Llama-3.2-1B-Q8_0 and Llama-3.2-3B-Q4_K_M, both CUDA-resident, behind the proxy, all re-measured on the merged binary:

- Streaming: `200`, `content-type: text/event-stream`, `cache-control: no-cache`, `transfer-encoding: chunked`, no `content-length`, placement headers present, 93 TCP arrivals carrying 94 SSE frames terminating in `data: [DONE]`, text reassembles coherently.
- A stock `curl -N` client streams correctly against the proxy address.
- Model-scoped placement holds on the streaming path: each model's request reached the node serving it.
- Non-streaming is unaffected.

**Control — relay, not buffer.** The same request sent to a node *directly* versus *through the proxy*, interleaved A/B over three rounds (this host's throughput drifts, so blocked runs are not trustworthy):

| | mean TCP arrivals | mean first byte | mean median gap |
|---|---|---|---|
| node direct | 76 | 86 ms | 8 ms |
| via proxy | 75 | 80 ms | 8 ms |

A buffering proxy would collapse to roughly one arrival. Instead the arrival count and the inter-arrival gap are indistinguishable from talking to the node itself.

**Control — cancellation.** The same long generation, two arms:

| | node returns to idle |
|---|---|
| client hangs up after 2 events | **268 ms** after the hang-up |
| client reads to the end | stream ran **21,690 ms** |

**Ablations.** Each new guarantee was removed in turn and the matching test failed, and only it: dropping the hang-up check leaked events (`3 -> 14`); buffering the body failed both the incrementality and the cancellation tests; removing the non-SSE fallback failed the readable-refusal test.

## Limits, stated plainly

- **No cross-machine arm.** Every measurement above is loopback on one host with two real engines. The second machine was unreachable when this was run, so nothing here is evidence about behaviour over a real network.
- The buffered (non-streaming) path is still not cancellable. That is inherited, not introduced, and is now stated in the module docs rather than left implied.
- The proxy still has no authentication of its own, and still serves only `/v1/chat/completions`; `/v1/models` is not proxied.
- **Placement under concurrency is not addressed here and has a known defect.** Because `policy::route` breaks ties on label, a burst of concurrent same-model requests can land on one node while others idle. On three stub nodes, 12 concurrent requests placed 10/2/0 and cost about 2.5x throughput. That is a separate change; this PR only makes the front door usable.


karan68:
Follow-up stacked on this: #652 fixes placement under concurrency. With this PR alone, a burst of concurrent requests for one model all land on whichever node sorts first - measured 8 of 8 on two real engines, with the second node never used. #652 is the third commit on that branch; its diff collapses to 5 files once this merges.

PR #662, opened 2026-08-14
fix(fabric): serve a request another node can take when one has gone

karan68:
## The defect

`Fabric::dispatch` places a request once and sends it once. The proxy reuses a fabric observation for `--observation-max-age-ms` (500 by default) rather than probing every node on every request, so inside that window its view of the fabric can name a node that is no longer there. The request then died against that node with **502 — even when a sibling was serving the same model and was idle**.

There was no node-level retry anywhere: `http::is_retryable` is EINTR-level only, and `ForwardError::Transport` (documented as "the request never completed against that node") went straight to `forward_error` → 502.

Measured on the built binary, two stub nodes serving one model, `--observation-max-age-ms 30000`:

```
request 1 -> 200 node=node-a attempts=1
node-a stopped listening
request 2 -> 502
  {"error":{"message":"node `node-a` could not be reached: cannot connect: ...
             the target machine actively refused it. (os error 10061)"}}
node-a completions: 1   node-b completions: 0
```

## The change

The request is placed again on another node serving the same model, up to `--max-forward-attempts` (**2** by default). Same binary, same scenario, after:

```
request 1 -> 200 node=node-a attempts=1
node-a stopped listening
request 2 -> 200 node=node-b attempts=2
  {"choices":[{"message":{"role":"assistant","content":"served by node-b"}}]}
node-a completions: 1   node-b completions: 1
```

`--max-forward-attempts 1` reproduces the 502 above exactly, on the same binary. That is the kill switch and the ablation lever, the same shape as `--observation-max-age-ms 0` in #656.

## What makes re-sending safe

A distinction, not a policy knob.

`HttpError::peer_never_received_it()` is true for **`Resolve` and `Connect` only** — both are raised before the first write, so the node cannot have started the work. Everything else, `Io` included, is "may have arrived": `write_all` does not report how many bytes left the socket, so a write that failed part-way is indistinguishable from one that never started. Erring towards *may have arrived* is what makes the retry safe without assuming anything is idempotent. `InvalidRequest` is deliberately excluded too — nothing was sent, but another node would refuse it identically.

`ForwardError` carries that through as a new `Unreachable { label, detail }` variant held apart from `Transport`, which now means "may have reached it". Only `Unreachable` is re-sent.

Two consequences worth stating plainly:

- A node that **took** the request and then failed ends the request with 502, exactly as before. It may already be generating.
- `Streaming::next_chunk` always maps to `Transport`, so once the response head has been relayed a retry is **structurally impossible**, not merely disallowed by a rule someone could later relax.

## Interaction with the probe cache (#656)

Failed nodes are excluded from the observation **already in hand** rather than triggering a re-probe. Re-probing between attempts would pay a full probe round — up to the probe budget against a node that has just been found to be gone — and would undo the single-flight probing #656 exists for. There is a test asserting the probe count does not grow with attempts.

`place()`'s existing "look again before refusing, because a `ModelUnavailable` decided from a reused observation becomes a 404 a client is told never to retry" path is skipped once anything is excluded: by then the caller already holds a real forwarding failure, and that is what it reports.

An observation that named a node found gone is dropped once the request finishes — **including when the request succeeded elsewhere**. Otherwise every later request in the window pays the same dial to the same absent node before being served.

## Reporting

`x-camelid-fabric-attempts` is now sent on every answer, alongside `x-camelid-fabric-node` and `x-camelid-fabric-reason`. Always present rather than only on a failover, so a client reads it off the header instead of inferring it from a header that is sometimes missing.

`dispatch` / `dispatch_streaming` now return `Dispatched` / `DispatchedStream` instead of tuples, so the attempt count has somewhere to live.

## Deliberately not done

- **`fabric run` does not fail over.** Its request body is built from the *chosen* node's active model when `--model` is absent, so a retry would have to rebuild the body per attempt. The proxy's body comes from the client and is node-independent, which is why the proxy can do this and the one-shot CLI cannot without a different shape.
- No circuit breaker or health-based ejection. A node stays a candidate until the next observation says otherwise.
- The default cap is 2, so a fabric with three or more absent nodes still fails. Raising it is a flag, not a code change.

## Docs correction included

The `Fabric proxy policy` section of `docs/CONFIGURATION.md` had gone stale on `main` in two ways, both next to where this change belongs:

- "The proxy has no key of its own, so it authenticates no client" — untrue since #658 added `--api-key` / `--api-key-file`.
- "streaming requests are refused with 400 rather than half-supported" — untrue since #651; the proxy relays SSE. `fabric run` still refuses, and the section now says which is which.

## Validation

Gates on `964843ab`:

| gate | result |
| --- | --- |
| `cargo test --lib` | 1837 passed / 0 failed / 86 ignored |
| `cargo test --lib fabric::` | 127 passed |
| `cargo test --test fabric_serve` | 32 passed (25 pre-existing + 7 new) |
| `cargo test --test fabric_end_to_end` | 14 passed |
| `cargo test --bin camelid` | 40 passed |
| `cargo clippy --all-targets -- -D warnings` | 0 warnings |
| `scripts/check-public-scrub.sh` | exit 0 |
| `git diff --check`, `git ls-files --eol` | clean, all 8 files `i/lf w/lf` |

### Ablations

Each was applied to the committed tree, the `fabric_serve` suite run, then reverted; `git diff` was empty afterwards each time.

| ablation | result |
| --- | --- |
| `ForwardError::node_never_received_it() -> false` (no failover) | **exactly the 4 failover tests** fail; 28 pass, including all three safety tests |
| `HttpError::peer_never_received_it() -> true` for every variant | **exactly 1** fails — `a_node_that_took_the_request_is_not_asked_to_run_it_again_elsewhere` — and its assertion prints the second node's answer, i.e. the double execution the rule prevents |
| drop the post-success `forget_observation` | **exactly 1** fails — `a_served_failover_still_drops_the_observation_that_named_the_vanished_node` |

The second ablation does **not** break the mid-stream test, which is the point: the "never reached" rule and the "already relayed" rule are independent mechanisms, and the tests are sensitive to them separately.

### New tests

Integration (`tests/fabric_serve.rs`, real sockets through the real router):

- `a_vanished_node_does_not_fail_a_request_another_node_can_serve` — asserts the answering node, `attempts=2`, and that the vanished node's completion count did not move.
- `a_budget_of_one_fails_the_request_the_way_it_did_before` — the ablation as a permanent test.
- `a_node_that_took_the_request_is_not_asked_to_run_it_again_elsewhere` — the node reads the request and hangs up; 502, and the sibling is never asked.
- `failover_stops_at_the_attempt_budget_rather_than_walking_the_fabric` — with its positive control: the same situation at 3 attempts does reach the third node, so the refusal is the budget and not a broken exclusion.
- `a_served_failover_still_drops_the_observation_that_named_the_vanished_node` — the next request is `attempts=1`.
- `a_vanished_node_does_not_fail_a_streaming_request`.
- `a_node_dying_mid_stream_is_not_replayed_on_another_node`.

Unit tests cover the classification exhaustively (a new `HttpError` variant has to be classified deliberately), the classification taken from the real dialling path rather than hand-built variants, the exclusion, and the attempt-budget floor.

`StubNode::stop_listening()` is new in the test harness: it stops the accept loop and releases the port, so connections are refused the way a machine that has gone refuses them, while what the node already recorded stays readable.

## Not covered

No cross-machine evidence — unchanged from #651 and #652, and still blocked on the second machine. Everything here is loopback: stub nodes for the tests, and stub nodes plus the built binary for the receipt.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -107,16 +107,29 @@ target/release/camelid fabric serve \
 
 Two limits are deliberate and worth knowing before you deploy it:
 
-- The proxy has no key of its own, so it authenticates no client. The acknowledgement flag is the
-  only way to bind it to a routable address, and it should only be used behind something that does
-  authenticate.
+- `--api-key` (or `--api-key-file`) is the key the proxy requires *from its clients*, and having one
+  satisfies the bind guard on its own. It deliberately reads no environment variable: on this command
+  `CAMELID_API_KEY` already names the token sent to nodes, and one value must not silently configure
+  both directions. Without a key, `--allow-unauthenticated-remote` is the only way to bind a routable
+  address, and it should only be used behind something that does authenticate.
 - `--bearer` (or `CAMELID_API_KEY`) is the token the proxy presents *to its nodes*, the same as
   `fabric status|route|run`. A node started with `--api-key` needs it: `/v1/health` is exempt from
   that node's auth, so without a token the node observes as ready and then answers every forwarded
   request with 401.
 
-Request bodies are bounded at the same 16 MiB default the node itself uses, and streaming requests
-are refused with 400 rather than half-supported.
+The proxy re-probes its nodes at most once per `--observation-max-age-ms` (500 by default) rather
+than once per request. Inside that window its view can be wrong, so a request placed on a node that
+has gone since is placed again on another node serving the same model, up to `--max-forward-attempts`
+(2 by default) nodes in all. A request is only ever sent twice when the first node was never reached
+and so cannot have started it: a node that accepted the request and then failed, or that failed
+part-way through a stream, ends the request with 502 rather than risking a second generation. Set
+`--max-forward-attempts 1` to fail on the first node instead. Every answer carries
+`x-camelid-fabric-node`, `x-camelid-fabric-reason` and `x-camelid-fabric-attempts`, so a client can
+tell a first-choice placement from a failover.
+
+Request bodies are bounded at the same 16 MiB default the node itself uses. A streaming request is
+relayed as it arrives; `fabric run`, which returns one complete answer, refuses `stream: true` with
+400 instead.
 
 ## Mixtral diagnostic gate
 
--- a/src/fabric/forward.rs
+++ b/src/fabric/forward.rs
@@ -44,7 +44,13 @@ impl Forwarded {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ForwardError {
-    /// The request never completed against that node.
+    /// The node was never reached, so it cannot have run the request.
+    ///
+    /// Held apart from [`ForwardError::Transport`] because that is the whole
+    /// difference between "safe to send somewhere else" and "may already be
+    /// generating"; see [`ForwardError::node_never_received_it`].
+    Unreachable { label: String, detail: String },
+    /// The request never completed against that node, and may have reached it.
     Transport { label: String, detail: String },
     /// The node answered, but not with JSON.
     Json { label: String, detail: String },
@@ -52,9 +58,44 @@ pub enum ForwardError {
     Unsupported(String),
 }
 
+impl ForwardError {
+    /// Whether the node provably never received the request.
+    ///
+    /// Only a caller that has said nothing to its own client on the strength of
+    /// this attempt may act on it: re-sending is safe because the node cannot
+    /// have started, not because the request is idempotent.
+    pub fn node_never_received_it(&self) -> bool {
+        matches!(self, Self::Unreachable { .. })
+    }
+
+    /// Which node failed, when the failure is attributable to one.
+    pub fn label(&self) -> Option<&str> {
+        match self {
+            Self::Unreachable { label, .. }
+            | Self::Transport { label, .. }
+            | Self::Json { label, .. } => Some(label),
+            Self::Unsupported(_) => None,
+        }
+    }
+}
+
+/// Tag an HTTP failure with the node it happened against, keeping the one
+/// distinction a retry depends on.
+fn dial_or_transport(label: &str, error: HttpError) -> ForwardError {
+    let (label, detail) = (label.to_string(), error.to_string());
+    if error.peer_never_received_it() {
+        ForwardError::Unreachable { label, detail }
+    } else {
+        ForwardError::Transport { label, detail }
+    }
+}
+
 impl std::fmt::Display for ForwardError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
+            Self::Unreachable { label, detail } => {
+                write!(f, "node `{label}` could not be reached: {detail}")
+            }
             Self::Transport { label, detail } => {
                 write!(f, "node `{label}` did not answer: {detail}")
             }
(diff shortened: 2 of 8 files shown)


karan68:
## Failover verified with a real machine leaving the fabric

This PR's receipt was taken against stub nodes on loopback. It now has a two-machine one:
a Windows node (CPU AVX2) and an Apple M5 Pro (Metal) over real Wi-Fi, both built from
`2ac971520`, same GGUF sha256 `3f87a880...c7ffd1`, proxy on the Windows box.

`--observation-max-age-ms 30000`, so the observation still names the Mac after it dies.

| step | result |
|---|---|
| req1, both alive | 200 `mac`, `attempts=1`, `LeastLoaded` |
| `pkill` the engine on the Mac | — |
| req2 | **200 `win`, `attempts=2`**, real body, 2912 ms |
| req3 | 200 `win`, `attempts=1` — observation forgotten, no wasted attempt |
| Mac restarted, req5 | 200 `mac`, `attempts=1` |

Kill switch, **identical situation**, only `--max-forward-attempts 1`:

```
502  node `mac` could not be reached: cannot connect: 192.168.29.158:8181:
     No connection could be made because the target machine actively refused it. (os error 10061)
```

Same setup, one flag, 502 vs 200 — so the retry is doing the work, and it can be turned off.

A note on how to reproduce this, because I got it wrong the first time: you have to warm the
observation **while the node is still alive** and kill it afterwards. If you start the proxy
after the node is already down, the fresh probe correctly excludes it, `place` never names it,
and nothing fails over — the test passes for the wrong reason and proves nothing.

`Unreachable` really was raised by a refused TCP connect (os error 10061), i.e. before any byte
was written, which is the condition `peer_never_received_it()` exists to express. The Mac's
engine log confirms it recorded no completion for the failed-over request.

Full cross-machine receipt, including throughput scaling and streaming over the link, is on
#663 since it sits on top of this one.

karan68:
**Cross-PR note — from test-merging the fabric stack locally. No action needed from you.**

I merged all seven open fabric PRs together on a scratch branch and ran the full gates. They compose green, but two things in this PR interact with #666, and I'd rather flag them now than have you hit a confusing red build mid-merge.

**1. This PR and #666 assert opposite contracts for a failed forward.**

`a_node_that_took_the_request_is_not_asked_to_run_it_again_elsewhere` asserts a 502 carries *no* node header:

```rust
assert_eq!(header(&headers, "x-camelid-fabric-node"), None);
```

#666's review fix deliberately made `forward_error` tag the failing node, on the grounds that a failure which doesn't name its node is one nobody can attribute. Each PR is green on its own; together, this assertion fails.

I think **#666 has it right**, so the change belongs here: the assertion should become `Some("node-a")`. That is also the stronger claim, since it proves the node that actually broke is identifiable.

**2. `tag()` gaining a third parameter here breaks a test helper in #666.**

This PR makes it `tag(headers, &decision, attempts)`; #666's `answering()` helper still calls the two-argument form, so the lib tests stop compiling with `E0061`. One-line fix.

**3. Worth keeping when they meet:** `ForwardError::label()` from this PR is a better single source of truth than the variant match #666 open-codes inside `forward_error`.

I'll push both fixes to whichever of the two merges second — you shouldn't have to resolve anything.

For reference, the gates on the fully merged tree: `cargo clippy --all-targets -- -D warnings` clean, `--lib fabric::` 164/0, `fabric_serve` 56/0, `fabric_end_to_end` 14/0, `--bin camelid` 40/0, public-scrub clean.

Suggested order for the whole stack is #664, #666, #668, #662, #667, #665, #663 — but happy to reorder if you'd rather take them differently.

PR #663, opened 2026-08-14
feat(fabric): place the engine's other inference routes, not only chat

karan68:
> Stacked on #662 (`feat/fabric-failover`). Both touch the same handlers in `src/fabric/server.rs`, so this is one commit on top of that branch rather than a conflict waiting to happen. Review or merge #662 first; the commit here is `2ac97152` and stands alone once it lands.

## The gap

The proxy answered two routes. The engine serves four real inference routes plus discovery:

| engine route | proxy, before |
| --- | --- |
| `POST /v1/chat/completions` | placed |
| `POST /v1/completions` | **bodyless 404 from axum** |
| `POST /v1/embeddings` | **bodyless 404** |
| `POST /v1/rerank`, `/v1/reranking` | **bodyless 404** |
| `GET /v1/models` | aggregated |
| `GET /v1/models/{model}` | **bodyless 404** |

So an address advertised as OpenAI-compatible answered `models.list` but not `models.retrieve`, and served chat but not completions or embeddings — with nothing in the body to say whether the route or the model was wrong.

## What this does

Places all four inference routes and adds `GET /v1/models/{model}`. Every one of those bodies carries `model`, so placement, affinity, the attempt budget, the body limit and the streaming decision are the ones chat already had — the handler is shared, not copied.

Measured on the built binary (`v0.6.1-125-g2ac971520`) against a stub node:

```
placed routes:
  POST /v1/chat/completions   -> 200 node=node-a reached-node=1 body-verbatim=true
  POST /v1/completions        -> 200 node=node-a reached-node=1 body-verbatim=true
  POST /v1/embeddings         -> 200 node=node-a reached-node=1 body-verbatim=true
  POST /v1/rerank             -> 200 node=node-a reached-node=1 body-verbatim=true
  POST /v1/reranking          -> 200 node=node-a reached-node=1 body-verbatim=true

refusals:
  POST /v1/responses             -> 501  the Responses and Conversations APIs keep their state...
  GET  /v1/conversations/c1/items -> 501  (same, on every method)
  POST /v1/audio/speech          -> 404
    this fabric proxy does not serve that route; it serves POST /v1/chat/completions,
    POST /v1/completions, POST /v1/embeddings, POST /v1/rerank, POST /v1/reranking,
    and GET /v1/models and GET /v1/models/{model}

discovery:
  GET /v1/models               -> 200 {"object":"list","data":[{"id":"receipt-model",...}]}
  GET /v1/models/receipt-model -> 200 {"id":"receipt-model","object":"model",...}
  GET /v1/models/not-loaded    -> 404 code=model_not_found
```

## What earns a route its place

That any node serving the model the request names could answer it. The client cannot tell which node it reached and must not need to.

That rules out the **Responses and Conversations** APIs. They keep their items in a SQLite store on the node that served the request (`src/api/responses_store.rs`), so placing them would appear to work and then lose a conversation the moment a follow-up landed elsewhere. They are refused with **501 and the reason**, on every method, rather than half-supported — a stored response is as node-local as the POST that created it. Anything else is a **404 naming what is served**.

Both refusals sit inside the client-key layer: an unauthenticated caller gets a 401 and learns nothing about the fabric, which is the same reasoning that keeps model names out of a 401 from `/v1/models`.

## Design notes

- **One table, two uses.** `PLACED_ROUTES` is what the router is folded from *and* what the 404 message is written from, so a route cannot be registered and unadvertised, or advertised and unregistered.
- **The path is not client-controlled.** Each route is registered with its own `&'static str`; nothing is read from the request line, so no query string or encoded segment can steer what reaches a node.
- **Retrieval and placement cannot disagree.** `GET /v1/models/{model}` answers from `policy::route` on the same observation, so a 200 there is a promise that a request naming that model would be placed. It inherits the distinction #654 introduced for free: a settled 404 when every node was consulted, a retryable 503 while one is unaccounted for.
- **`stream: true` stays a property of the request, not of the route.** A client that sets it on a route with nothing to stream gets that route's complete answer relayed, because a non-streaming answer already comes back as `StreamOutcome::Buffered`. There is a test for that rather than a rule against it.

## Validation

| gate | result |
| --- | --- |
| `cargo test --lib` | 1844 passed / 0 failed / 86 ignored |
| `cargo test --lib fabric::` | 134 passed (127 + 7 new) |
| `cargo test --test fabric_serve` | 41 passed (32 + 9 new) |
| `cargo test --test fabric_end_to_end` | 14 passed |
| `cargo test --bin camelid` | 40 passed |
| `cargo clippy --all-targets -- -D warnings` | 0 warnings |
| `scripts/check-public-scrub.sh` | exit 0 |
| `git diff --check`, `git ls-files --eol` | clean, all 3 files `i/lf w/lf` |

### Ablations

Each applied to the committed tree, suites re-run, then reverted; `git diff` empty afterwards each time.

| ablation | result |
| --- | --- |
| register only `/v1/chat/completions` | **exactly the 6** tests that exercise the new routes fail; 35 pass, including every pre-existing chat, streaming, auth and failover test |
| drop the node-local refusals | **exactly 3** fail (1 integration, 2 unit), and the failure output shows `/v1/responses` falling through to the generic 404 |
| drop the unknown-route fallback | **exactly 2** fail |
| let model retrieval answer 200 without consulting placement | **exactly 3** fail, including the one that pins the retryable-vs-settled distinction |

### New tests

Integration, over real sockets: every placed route arrives with its path and a **byte-for-byte** body (a field the proxy has never heard of has to survive); a non-chat route is still scoped to the node serving the model; `/v1/completions` streams; `/v1/embeddings` with `stream: true` still answers; the node-local refusal costs not even a probe; an unserved route names the served ones; an unauthenticated caller cannot read the route table and cannot reach a node on a new route; a malformed body on a new route is refused in the fabric error shape.

Unit: the 404 message names every entry in the table; the two tables do not overlap; node-local routes are refused on every method; a model can be retrieved, an unserved one is refused exactly as placing it would be, and a fabric with an unreachable node stays retryable.

## Not covered

Still loopback only — no cross-machine evidence, unchanged from #651/#652/#662. `/v1/messages` is not registered: the engine refuses it as unsupported, and this proxy's 404 names the alternatives, which is more use than relaying that refusal.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -131,6 +131,31 @@ Request bodies are bounded at the same 16 MiB default the node itself uses. A st
 relayed as it arrives; `fabric run`, which returns one complete answer, refuses `stream: true` with
 400 instead.
 
+### What the proxy serves
+
+It places the engine's stateless inference routes, and answers discovery for the whole fabric:
+
+| Route | Behaviour |
+| --- | --- |
+| `POST /v1/chat/completions` | Placed on a node serving the model named in the body. |
+| `POST /v1/completions` | Same. |
+| `POST /v1/embeddings` | Same. |
+| `POST /v1/rerank`, `POST /v1/reranking` | Same. |
+| `GET /v1/models` | The union of what every ready node is serving. |
+| `GET /v1/models/{model}` | 200 exactly when a request naming that model would be placed. |
+
+A route is placed only if any node serving the named model could answer it, because the client cannot
+tell which node it reached and must not need to. That rules out the Responses and Conversations APIs:
+they keep their items in a SQLite store on the node that served the request, so a follow-up landing on
+another node would find nothing. Those routes are refused with **501** and a reason rather than
+half-supported — send them to a node directly, where they work as documented. Any other route is a
+**404** naming what is served, so a client can tell a wrong route from a wrong model. Both refusals
+are behind the client key, if one is configured: an unauthenticated caller gets a 401 and learns
+nothing about the fabric.
+
+`stream: true` is read from the request rather than assumed from the route, so a client that sets it
+on a route with nothing to stream still gets that route's complete answer instead of an empty stream.
+
 ## Mixtral diagnostic gate
 
 The checked Mixtral 8×7B Q8 row has one-token evidence, but its longer continuation parity gate is
--- a/src/fabric/server.rs
+++ b/src/fabric/server.rs
@@ -43,6 +43,11 @@
 //! * A stop (Ctrl-C, or SIGTERM where there is one) closes the listener and
 //!   lets the requests already in flight finish, bounded by `forward_timeout`.
 //!   A kill still drops them: this covers being asked to stop, not being shot.
+//! * `/v1/health` answers readiness, not liveness — a 503 from it means "no
+//!   node is ready", never "restart me". It is also the one route a client
+//!   reaches without a key, because the engine's shared `authenticate` keeps
+//!   `/v1/health` public; it therefore names the fabric's nodes and models only
+//!   when this proxy is bound to loopback.
 //! * A node that becomes ready, or loads a different model, is routed to only
 //!   once the current observation expires. A node that *stops* answering is not
 //!   waited on: the request that finds it gone is placed on another node
@@ -142,6 +147,15 @@ pub struct ServeConfig {
     pub forward_timeout: Duration,
     /// What a client must present to be served at all.
     pub auth: ClientAuth,
+    /// The address this proxy listens on.
+    ///
+    /// Read only to decide whether `/v1/health` may name the fabric's members:
+    /// a loopback listener cannot be reached from off the machine.
+    /// [`serve_on_until`] overwrites this with the address the listener
+    /// actually bound, so the value the disclosure rule sees can never be a
+    /// caller's guess. It is set there rather than in [`serve_on`] because
+    /// every serving path goes through it, including the injected-stop seam.
+    pub bound: SocketAddr,
 }
 
 #[derive(Clone)]
@@ -158,6 +172,7 @@ pub fn router(fabric: Fabric, config: ServeConfig) -> Router {
     Router::new()
         .route("/v1/chat/completions", post(chat_completions))
         .route("/v1/models", get(models))
+        .route("/v1/health", get(health))
         // Without this, axum's 2 MiB default would refuse conversations the
         // node behind the proxy accepts.
         .layer(DefaultBodyLimit::max(DEFAULT_MAX_REQUEST_BODY_BYTES))
@@ -361,9 +376,14 @@ pub async fn serve_on(
 pub async fn serve_on_until(
     listener: tokio::net::TcpListener,
     fabric: Fabric,
-    config: ServeConfig,
+    mut config: ServeConfig,
     stop: impl std::future::Future<Output = ()> + Send + 'static,
 ) -> std::io::Result<()> {
+    // The listener is the only authority on what was actually bound, and the
+    // health disclosure rule turns on it. It is set here rather than in
+    // `serve_on` so that a caller entering through this seam cannot skip it.
+    config.bound = listener.local_addr()?;
+
     let drain = config.forward_timeout;
     // The peer address is only available to a service built this way, and the
     // access log is the only thing that reads it.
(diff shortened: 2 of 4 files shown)


karan68:
## Cross-machine receipt: this stack run on two real machines

Every receipt on this stack so far (#651, #652, #656, #658, #662, #663) was taken against
in-process stub nodes or two engines sharing one box. #652 explicitly **refused** a throughput
claim for that reason. This is that gap closed: a second physical machine, on real Wi-Fi.

**Fabric**

| | node `win` | node `mac` |
|---|---|---|
| platform | windows x86_64, AVX2 | macOS arm64, Apple M5 Pro |
| backend | `cpu_q8_runtime_repack` | `metal_resident_q8_runtime` |
| decode path | `q8_0_decode_packed_rows4_avx2` | `q8_0_metal_resident_decode` |
| threads | 32 | 18 |
| `build` in `/v1/health` | `v0.6.1-125-g2ac971520` | `0.6.1+2ac97152091b` |

Both built from this branch's head `2ac971520`. Same GGUF on both, sha256
`3f87a880...c7ffd1` verified independently on each machine. Proxy ran on the Windows box;
probe latency was 1 ms to `win` and 13 ms to `mac`.

### 1. The five routes this PR places, served by real engines

| check | status | reached node |
|---|---|---|
| `POST /v1/chat/completions` | 200 | `mac` |
| `POST /v1/completions` | 200 | `mac` |
| `POST /v1/embeddings` | 422 | `mac` |
| `POST /v1/rerank` | 422 | `mac` |
| `POST /v1/reranking` | 422 | `mac` |
| `GET /v1/models` | 200 | answered from the observation |
| `GET /v1/models/{id}` | 200 | answered from the observation |
| `GET /v1/models/not-loaded-anywhere` | 404 `model_not_found` | — |
| `POST /v1/responses` | 501 | never placed |
| `GET /v1/conversations/c1/items` | 501 | never placed |
| `POST /v1/audio/speech` | 404 listing served routes | — |

The three 422s are the interesting ones. That is the **node's own** refusal — `model "Llama 3.2
1B Instruct" uses architecture "llama" ...; it is not an admitted embedding artifact` — relayed
verbatim, because a Llama chat model genuinely is not an embedding artifact. On `main` those
three paths return a bodyless axum 404. A stub can be made to return any status; only a real
engine can produce a refusal the proxy had no way to invent.

### 2. Throughput scales across machines

Closed loop, 4 concurrent clients, 60 s per arm, unique marker at the front of every prompt so
the engine's exact-prompt cache cannot serve a repeat.

| arm | req/s | completion tok/s | p50 | p95 | placed |
|---|---|---|---|---|---|
| `win` only | 1.156 | 27.5 | 3363 ms | 4683 ms | win 73 |
| `mac` only | 6.017 | 117.3 | 637 ms | 837 ms | mac 366 |
| **both** | **7.773** | **140.5** | 325 ms | 1425 ms | mac 392 / win 80 |

911 requests, **0 failures**, every one `attempts=1`.

- Measured 140.5 tok/s against 144.8 predicted by summing the single-node arms = **97% of
  perfectly additive**. On requests/s it is 108% of the sum.
- Both nodes were saturated in the mixed arm: `win` did 1.32 req/s inside it vs 1.16 alone,
  `mac` 6.46 vs 6.02. Neither was starved.
- The `mac` is **5.2x** the `win` node. Adding a node that slow is only worth +20% tok/s — and
  that is exactly what additivity predicts, since `win` is 19% of the combined total. The
  headline number is small because the second node is slow, not because placement leaked work.
- Worth stating plainly: `reserved()` ranks on in-flight count and knows nothing about node
  speed, yet it absorbed a 5.2x asymmetry without any speed-awareness. That is the
  reservation design in #652 doing the work.

**The cost, which belongs in the docs:** adding the slow node **halves p50 (637 -> 325 ms) and
nearly doubles p95 (837 -> 1425 ms)**. A heterogeneous fabric buys throughput and median latency
and pays for it in the tail.

### 3. Failover when a real machine leaves

`--observation-max-age-ms 30000` so the observation still names the node after it dies.

- req1 -> 200 `mac`, `attempts=1`
- `pkill` the engine on the Mac
- req2 -> **200 `win`, `attempts=2`**, real body
- req3 -> 200 `win`, `attempts=1` (observation correctly forgotten)

Negative control, identical situation, `--max-forward-attempts 1`:
**502** naming the node — `node `mac` could not be reached: cannot connect: ... (os error 10061)`.
Same setup, one flag, 502 vs 200.

### 4. Streaming across the link

207 SSE events, first event at **49 ms**, **largest inter-event gap 54 ms**, `[DONE]` received,
full 638-character answer. No stall and no flap on real Wi-Fi.

### 5. Multi-address DNS, with real resolution

`localhost` resolves to `::1` **and** `127.0.0.1`. With the node bound to `127.0.0.1` only, `::1`
genuinely refuses. Addressed as `localhost:8181` the fabric returned **200** — `connect_any` from
#616 ("a dead leading address does not hide a live one") confirmed against real resolution order
rather than a constructed test.

### 6. Same answer regardless of placement

Same prompt, `temperature: 0`, sent straight to each node: **identical sha256**
`e29b8ac0b7866ba467281212eb68030708e2329c8c74ee4f613f4b7139c0f857` from x86_64 AVX2 CPU and from
ARM64 Metal. So which node serves a request is not observable in the answer.

### What this does not show

- One 60 s run per arm. No repeats, no confidence intervals. Treat the ratios as one measurement,
  not a benchmark.
- Two nodes, one 1B Q8_0 model, short prompts, 48-token answers.
- The Windows binary is a default build with **no CUDA** (`cuda_resident_active: false`), so
  `win` is a CPU AVX2 node. A CUDA node would change the ratio, not the additivity.
- Byte-identity in section 6 is one prompt on one model. It is not a parity claim.
- macOS `.local` mDNS does not resolve from Windows here (`os error 11001`), so the Mac was
  addressed by IP. Section 5 used `localhost` instead to get a genuine multi-address name.

### One thing this turned up, not fixed here

`fabric serve` prints its listening line and nothing else. Pointed at a node whose name can never
resolve, stdout is one line, **stderr is empty**, and the proxy serves at reduced capacity
indefinitely. `fabric status` is honest about it (`offline`, `os error 11001`,
`1 ready, 0 not ready, 1 offline`); `serve` never mentions it. Filed separately so this PR stays
about routes.

PR #665, opened 2026-08-15
feat(fabric): give the proxy a health endpoint of its own

karan68:
## The gap

`fabric serve` had no address anything could probe. `/v1/health` matched no route, so a load
balancer or an orchestrator got a **bodyless 404** whatever the fabric was doing — a proxy with
two healthy nodes and a proxy with nothing behind it were indistinguishable from the outside.

This is the first of the remaining production gaps I listed on #663. It is separate from #664
(which reports the fabric once, at startup); this answers the same question at any time, to a
machine.

## Readiness, and why the status code is what it is

`GET /v1/health` answers **200 while at least one node is ready, 503 when none is.**

That threshold is not a preference — it is exactly the condition placement needs. With one ready
node, some request can be served. With none, every request is refused, and a proxy that kept
answering 200 would be inviting traffic it can only turn away. It also means **one node of
several going down does not take the proxy out of rotation**, which is the behaviour you want and
the opposite of what a naive "all nodes healthy" check would do.

**It is readiness, not liveness.** Restarting the proxy cannot bring a node back, so wiring this
status code to a `livenessProbe` would restart a perfectly good process because someone else's
machine is down. The body carries the liveness fact separately as `ok`, which is true whenever
the process answers at all. Both are stated in the module docs and the handler docs.

The answer comes from the observation placement already uses, so a load balancer polling once a
second does **not** become a probe of every node once a second. There is a test for that.

## The part that changed the design

I assumed health would sit behind `ClientAuth` like every other route, wrote a test asserting an
unauthenticated probe gets 401 — and it failed. The engine's shared `authenticate` deliberately
keeps health public:

```rust
// src/api/server.rs
!matches!(path, "/health" | "/v1/health" | "/" | "/index.html")
```

with its own test, `router_authenticates_api_routes_but_keeps_health_public`. Since #658 the
proxy reuses that check rather than writing a second security primitive, so it inherits the
exemption. Good: a probe should not need a credential.

But that turns the body into a **disclosure decision**, not a formatting one:

* the node list is every label and address in the fabric — an attacker's target list, and
  `--allow-unauthenticated-remote` puts this route in reach of anyone;
* the model list is the same answer `/v1/models` gives, and that route **is** behind the key, with
  an existing test asserting an unauthenticated caller must not learn what is served. Publishing
  it here would have handed it out through a side door.

So node detail *and* the model list are included **only when the proxy is bound to loopback**.
That mirrors `crate::api`, which withholds its own `executable` and `listen_addr` on exactly the
same fact (`state.serve_addr.ip().is_loopback()`). Off-box callers get the verdict and nothing
else, which is all a load balancer reads.

The gate keys off the **bound** address, and `serve_on` overwrites `ServeConfig.bound` with the
listener's real `local_addr()`, so the value the rule sees cannot be a caller's guess.

## Live receipt, on two real machines

Windows proxy + a Windows node + an Apple M5 Pro node across real Wi-Fi.

**Loopback bind, both machines up — 200:**

```json
{"ok":true,"service":"camelid-fabric","version":"0.6.1","build":"v0.6.1-123-g1a04ff0a4-dirty",
 "ready":true,"nodes":{"total":2,"ready":2,"not_ready":0,"unreachable":0},
 "models":["Llama 3.2 1B Instruct"],
 "node_detail":[{"spec":{"label":"win","host":"127.0.0.1","port":8181},"status":{"state":"ready",...},"latency_ms":1},
                {"spec":{"label":"mac","host":"192.168.29.158","port":8181},"status":{"state":"ready",...},"latency_ms":23}]}
```

**The Mac is killed — still 200, still in rotation:**

```json
{"ready":true,"nodes":{"total":2,"ready":1,"not_ready":0,"unreachable":1}, ...
 {"spec":{"label":"mac",...},"status":{"state":"unreachable","reason":"cannot connect: ..."},"latency_ms":null}}
```

**The last node is killed — 503:**

```json
{"ok":true,...,"ready":false,"nodes":{"total":2,"ready":0,"not_ready":0,"unreachable":2},"models":[]}
```

**Exposed bind (`0.0.0.0` + `--allow-unauthenticated-remote`), the request issued from the Mac
over Wi-Fi with no credential** — this is the case only a second machine can produce:

```
HTTP 200
{"ok":true,"service":"camelid-fabric","version":"0.6.1","build":"v0.6.1-123-g1a04ff0a4-dirty","ready":true}
```

No node list, no addresses, no model names. The same request to the same proxy over loopback
returns the full body above.

For contrast, a **node's** `/v1/health` is a different thing entirely — `{"ok":true,"engine":"camelid",
...,"generation_ready":true,"active_model_id":"Llama 3.2 1B Instruct",...}`. The proxy deliberately
has no `engine` field and a different `service` name, so nothing can mistake one for the other. A
proxy has no single `active_model_id` to report anyway.

## Tests

8 unit + router tests and 5 integration tests over real sockets, including:

* readiness both ways, and that the verdict is identical off-box while the detail is not;
* a partial fabric stays ready;
* a probe with no key is answered while `/v1/models` and `/v1/chat/completions` on the same proxy
  still return 401 — the contrast that justifies the redaction;
* five health checks against two nodes produce exactly two probes.

**Ablations, each isolating one rule:**

| ablation | result |
|---|---|
| disclose detail unconditionally | **exactly 3** fail: both exposure tests + the route-level one |
| status code always 200 | **exactly 3** fail: 2 unit + 1 integration readiness tests |

Tree restored and re-verified green after each.

Note the coverage asymmetry that shaped the test layout: an integration test can only bind
loopback, so it can *never* reach the redacted branch. That branch is covered at the pure-function
and router level, which is why `health_report` is a pure function taking a bool rather than a
`SocketAddr`.

## Gates

`fmt --check` 0 · `clippy --all-targets -D warnings` **0** · `--lib fabric::` **128** (120 + 8) ·
`--test fabric_serve` **30** (25 + 5) · `--test fabric_end_to_end` 14 · `--bin camelid` 40 ·
`--lib api::server` 8 · `check-public-scrub.sh` 0 · `git diff --check` 0 · all three files
`i/lf w/lf`. 3 files, +402/-3.

## Limits

* It is a snapshot, not a subscription: a node that dies is invisible until something asks, and
  the answer may be up to `--observation-max-age-ms` old. The body does not currently say how old.
* No liveness endpoint of its own. If you want one, use TCP connect — do not use this status code.
* Off-box the body is deliberately thin. An operator who wants detail from off the machine should
  use `fabric status`, or an authenticated `/v1/models`, not this.
* Independent of #664 and #663; no conflicts expected, though after #663 merges the "unknown
  route" 404 message should probably start listing `/v1/health` among the served routes.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -47,7 +47,8 @@
 //! A `session_start` with no matching `session_exit` means the process did not
 //! leave through that path. Be precise about what that does and does not say:
 //! `camelid serve` installs no signal handler (there is no `ctrl_c` or
-//! `with_graceful_shutdown` anywhere in the engine), so an ordinary Ctrl-C ends
+//! `with_graceful_shutdown` in its path — `fabric serve` has one, but that is a
+//! different command in a different process), so an ordinary Ctrl-C ends
 //! the process exactly as abruptly as an out-of-memory kill, and the two are
 //! recorded identically — which is to say, not at all. Read a missing exit
 //! record as "did not fail on the way out", never as evidence of a kill.
--- a/src/fabric/server.rs
+++ b/src/fabric/server.rs
@@ -43,6 +43,11 @@
 //! * A stop (Ctrl-C, or SIGTERM where there is one) closes the listener and
 //!   lets the requests already in flight finish, bounded by `forward_timeout`.
 //!   A kill still drops them: this covers being asked to stop, not being shot.
+//! * `/v1/health` answers readiness, not liveness — a 503 from it means "no
+//!   node is ready", never "restart me". It is also the one route a client
+//!   reaches without a key, because the engine's shared `authenticate` keeps
+//!   `/v1/health` public; it therefore names the fabric's nodes and models only
+//!   when this proxy is bound to loopback.
 //! * A node that becomes ready, or loads a different model, is routed to only
 //!   once the current observation expires. A node that *stops* answering is not
 //!   waited on: the request that finds it gone is placed on another node
@@ -142,6 +147,15 @@ pub struct ServeConfig {
     pub forward_timeout: Duration,
     /// What a client must present to be served at all.
     pub auth: ClientAuth,
+    /// The address this proxy listens on.
+    ///
+    /// Read only to decide whether `/v1/health` may name the fabric's members:
+    /// a loopback listener cannot be reached from off the machine.
+    /// [`serve_on_until`] overwrites this with the address the listener
+    /// actually bound, so the value the disclosure rule sees can never be a
+    /// caller's guess. It is set there rather than in [`serve_on`] because
+    /// every serving path goes through it, including the injected-stop seam.
+    pub bound: SocketAddr,
 }
 
 #[derive(Clone)]
@@ -158,6 +172,7 @@ pub fn router(fabric: Fabric, config: ServeConfig) -> Router {
     Router::new()
         .route("/v1/chat/completions", post(chat_completions))
         .route("/v1/models", get(models))
+        .route("/v1/health", get(health))
         // Without this, axum's 2 MiB default would refuse conversations the
         // node behind the proxy accepts.
         .layer(DefaultBodyLimit::max(DEFAULT_MAX_REQUEST_BODY_BYTES))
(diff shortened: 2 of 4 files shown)


karan68:
Rebased onto `main` now that #664 is in — the two touched the same few lines in the `fabric serve` handler in `src/main.rs`. Force-pushed `ce41d3ff5` -> `2db246c7a`; the only difference is absorbing #664, the PR's own change is untouched (`git diff ce41d3ff5 2db246c7a` is exactly #664's two files, +149).

Gates re-run on the rebased head, all green: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` **0 findings**, `--lib fabric::` **132/0**, `--test fabric_serve` **31/0**, `--test fabric_end_to_end` 14/0, `--bin camelid` 40/0, public-scrub clean, `git diff --check` clean, all files LF.

Nothing needed from you — this is just keeping it ready to merge.

PR #666, opened 2026-08-15
feat(fabric): record what the proxy is actually doing

karan68:
## The gap

`fabric serve` printed its listening line and then nothing, ever. `src/fabric` contained **no
logging of any kind** — not a `tracing` event, not a `println`. There was no way to answer "is
anything reaching this", "which machine served that request", or "what is failing" without a
packet capture.

This is gap #2 of the production list I put on #663. (#665 answers "can you take traffic" to a
machine; this answers "what did you just do" to a human.)

## What it does

One line per request through `tracing`: INFO normally, WARN on a 5xx.

```
INFO camelid::fabric::server: fabric request method=POST path=/v1/chat/completions status=200 node="mac" reason="LeastLoaded" elapsed_ms=232
WARN camelid::fabric::server: fabric request failed method=POST path=/v1/chat/completions status=503 node="-" reason="-" elapsed_ms=2013
```

`node` and `reason` are the point. They are read back off the response rather than threaded out
of dispatch, because `tag` already puts them there for the client and one fact with two sources
drifts.

**Why not just `TraceLayer`.** The engine's router already uses tower-http's `TraceLayer`
(`src/api/mod.rs:2581`), and reaching for it here was the obvious move. It is not enough on its
own: it can only see HTTP, so it cannot name the node — the one decision this process exists to
make — and it records at DEBUG, below the level an operator actually runs at. I measured that
second point rather than assuming it: two engine nodes served **911 requests** during the #663
throughput run and their logs contain nothing but `[hw]`/`[cuda]` startup lines.

**Placed outside authentication**, deliberately: a refused request is exactly the one worth
having a record of, and inside the layer it would be invisible.

**Off by default.** Nothing is written unless `RUST_LOG` asks, which is how the rest of this
binary already behaves. `RUST_LOG=camelid=info` is the access log; `RUST_LOG=camelid=warn`
narrows it to failures. A proxy started without it behaves exactly as before — shown below.

**Never recorded:** request and response bodies, and every request header, one of which is the
client's key. The query string is dropped for the same reason; the path alone identifies the
route.

## Live receipt, two real machines

A Windows proxy over a Windows node and an Apple M5 Pro node across Wi-Fi.

**A. `RUST_LOG` unset — two requests served, and the proxy is as silent as it was before:**

```
--- stdout (1 lines) ---
fabric serve listening on 127.0.0.1:8490
```

**B. `RUST_LOG=camelid=info` — six requests, six lines, each naming the machine that answered
across the LAN:**

```
INFO ... fabric request method=POST path=/v1/chat/completions status=200 node="mac" reason="LeastLoaded" elapsed_ms=232
INFO ... fabric request method=POST path=/v1/chat/completions status=200 node="mac" reason="LeastLoaded" elapsed_ms=121
INFO ... fabric request method=POST path=/v1/chat/completions status=200 node="mac" reason="LeastLoaded" elapsed_ms=117
... (6 total)
```

**C. `RUST_LOG=camelid=warn` — two healthy requests produce nothing; then every node is killed:**

```
WARN ... fabric request failed method=POST path=/v1/chat/completions status=503 node="-" reason="-" elapsed_ms=2013
```

**D. a refusal is recorded, and the key is not.** With `--api-key`, an unauthenticated request
and then a correct one:

```
INFO ... fabric request method=POST path=/v1/chat/completions status=401 node="-" reason="-" elapsed_ms=0
INFO ... fabric request method=POST path=/v1/chat/completions status=200 node="mac" reason="LeastLoaded" elapsed_ms=200
--- does the key appear anywhere in the log? ---
no: the credential never reaches the log
```

Note `status=401 ... elapsed_ms=0`: the refusal cost nothing and touched no node, which is the
property #658 exists to hold.

## Tests

Six unit tests that assert on **what was actually written**, by installing a capturing
`tracing` subscriber and reading the formatted output back — not by checking that a function was
reached. They run on the default current-thread runtime on purpose: the subscriber is
thread-local, so on a multi-threaded one the task can resume on a thread that never had it
installed and capture nothing.

The middleware is also driven against a synthetic handler that returns the fabric's own headers,
so "the node is recorded" is tested without placing anything.

| ablation | result |
|---|---|
| move the layer inside the auth layer | **exactly 1** fails: `a_refused_request_is_still_recorded` |
| drop the WARN branch | **exactly 1** fails: `a_failure_is_recorded_at_warn_and_a_success_is_not` |

Both restored and re-verified green.

## Gates

`fmt --check` 0 · `clippy --all-targets -D warnings` **0** · `--lib fabric::` **126** (120 + 6) ·
`--test fabric_serve` 25 · `--test fabric_end_to_end` 14 · `--bin camelid` 40 ·
`check-public-scrub.sh` 0 · `git diff --check` 0 · `i/lf w/lf`. One file, +256/-4.

## Limits

* No client address: that needs `ConnectInfo` wiring the proxy does not have today.
* No request id, so a line cannot be correlated with a node's own logs.
* On a streaming answer `elapsed_ms` is time to the response **head**, not to the last event —
  the middleware returns once the head is ready while the body is still relaying.
* One line per request is a lot at high rates; `RUST_LOG=camelid=warn` is the volume control.
* Independent of #664 and #665. All three touch `src/fabric/server.rs` but different regions.

### DIFF
--- a/src/fabric/server.rs
+++ b/src/fabric/server.rs
@@ -634,6 +634,7 @@ fn route_error(error: RouteError) -> Response {
 }
 
 fn forward_error(error: ForwardError) -> Response {
+    let message = error.to_string();
     // The node was reachable but refused unsupported input: that is the
     // caller's mistake, not the upstream's, so it is a 400 not a 502/503.
     let status = match &error {
@@ -642,7 +643,15 @@ fn forward_error(error: ForwardError) -> Response {
         | ForwardError::Transport { .. }
         | ForwardError::Json { .. } => StatusCode::BAD_GATEWAY,
     };
-    error_response(status, &error.to_string())
+    let label = error.label().map(str::to_string);
+    let mut response = error_response(status, &message);
+    // The node that failed rides on the answer for the same reason [`tag`] puts
+    // the node that served there: placement is read back off the response, so a
+    // failure that does not carry its node is one nobody can attribute.
+    if let Some(label) = label {
+        insert(response.headers_mut(), "x-camelid-fabric-node", &label);
+    }
+    response
 }
 
 fn error_response(status: StatusCode, message: &str) -> Response {
@@ -769,7 +778,7 @@ mod tests {
                             reason: RouteReason::LeastLoaded,
                             affinity_lost: None,
                         };
-                        tag(response.headers_mut(), &decision);
+                        tag(response.headers_mut(), &decision, 1);
                     }
                     response
                 }),
@@ -822,13 +831,17 @@ mod tests {
             auth: ClientAuth::resolve(Some("s3cret".to_string()), None).expect("a key resolves"),
             ..open_config()
         };
-        let written = logged(
-            router(Fabric::new(Vec::new()), config),
-            get_request("/v1/models"),
-        )
-        .await;
+        let mut refused = get_request("/v1/models");
+        // Actually presented, so the assertion below is about a credential that
+        // reached the middleware rather than one that was never sent.
+        refused.headers_mut().insert(
+            "authorization",
+            HeaderValue::from_static("Bearer presented-9f3a"),
+        );
+        let written = logged(router(Fabric::new(Vec::new()), config), refused).await;
         assert!(written.contains("status=401"), "{written}");
-        // The credential itself must never reach the log.
+        // No request header reaches the log, and one of them is the client's key.
+        assert!(!written.contains("presented-9f3a"), "{written}");
         assert!(!written.contains("s3cret"), "{written}");
     }
 
@@ -855,6 +868,34 @@ mod tests {
             served.is_empty(),
             "a served request must not warn: {served}"
         );
+    }
+
+    /// A node that could not be reached is the failure most worth attributing,
+    /// and it is the one an operator is reading WARN to find. The error already
+    /// carries the label, so a line that cannot name the node is a line that
+    /// sends them to a packet capture anyway.
+    #[tokio::test]
+    async fn a_failure_to_reach_a_node_records_which_node() {
+        let app = Router::new()
+            .route(
+                "/v1/chat/completions",
+                post(|| async {
+                    forward_error(ForwardError::Transport {
+                        label: "node-a".to_string(),
+                        detail: "connection refused".to_string(),
+                    })
+                }),
+            )
+            .layer(middleware::from_fn(access_log));
+
+        let written = logged_at(
+            tracing::Level::WARN,
+            app,
+            request(serde_json::json!({ "model": "m" })),
+        )
+        .await;
+        assert!(written.contains("status=502"), "{written}");
+        assert!(written.contains("node=\"node-a\""), "{written}");
     }
 
     /// A query string can carry anything a client puts there, and the path
--- a/src/fabric/server.rs
+++ b/src/fabric/server.rs
@@ -33,6 +33,10 @@
 //!   unless a key is configured or the risk is acknowledged explicitly.
 //! * Authentication is all-or-nothing: there is one key, it is not per-client,
 //!   and nothing here revokes or rotates it while the proxy is running.
+//! * Every request is recorded through `tracing` at INFO, and a failure at
+//!   WARN, which means nothing is written unless `RUST_LOG` asks for it — the
+//!   same as the rest of this binary. `RUST_LOG=camelid=info` is the whole
+//!   access log; `RUST_LOG=camelid=warn` narrows it to what failed.
 //! * A node that becomes ready, or loads a different model, is routed to only
 //!   once the current observation expires. A node that *stops* answering is not
 //!   waited on: the request that finds it gone is placed on another node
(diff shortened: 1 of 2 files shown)


karan68:
**Cross-PR note — from test-merging the fabric stack locally. No action needed from you.**

Same note as on #662, from this side. I merged all seven open fabric PRs on a scratch branch and ran the full gates; they compose green, but this PR and #662 interact in two places.

**1. Opposite contracts for a failed forward.**

The review fix in this PR makes `forward_error` tag the failing node, so a 502 carries `x-camelid-fabric-node` — a failure that does not carry its node is one nobody can attribute. #662 has a test asserting exactly the opposite, that a 502 carries no such header. Green apart, red together.

I think **this PR has it right**, so the fix belongs in #662's assertion rather than here.

**2. `answering()` here calls `tag(response.headers_mut(), &decision)`.**

#662 gives `tag` a third parameter, `attempts: usize`, so this helper stops compiling with `E0061` once both are in. One-line fix — pass `1`, since the helper stands in for a single placement.

**3. Worth taking when they meet:** #662 adds a `ForwardError::label()` accessor, which is a cleaner single source of truth than the variant match this PR open-codes inside `forward_error`.

I'll push both fixes to whichever of the two merges second, so nothing lands on you.

Gates on the fully merged tree, for reference: `cargo clippy --all-targets -- -D warnings` clean, `--lib fabric::` 164/0, `fabric_serve` 56/0, `fabric_end_to_end` 14/0, `--bin camelid` 40/0, public-scrub clean.

Suggested order for the whole stack is #664, #666, #668, #662, #667, #665, #663 — happy to reorder.

karan68:
Rebased onto `main` now that #662 and #667 are in, and picked up the two cross-PR fixes I flagged above. This PR now touches 2 files instead of 1 — the second is deliberate, see fix 2.

**1. `forward_error` conflict resolved in #662's favour, structurally.** #662 added a `ForwardError::Unreachable` variant and a `ForwardError::label()` accessor. Rather than keep this PR's open-coded variant match, `forward_error` now uses `error.label()` — one source of truth for "which node failed", and the new variant is handled for free.

**2. `tests/fabric_serve.rs` — the contract change this PR makes.** #662's `a_node_that_took_the_request_is_not_asked_to_run_it_again_elsewhere` asserted a 502 carries *no* `x-camelid-fabric-node`. This PR makes a failed forward name its node, so that assertion is now wrong and the test fails. Changed it to `Some("node-a")`, which is the stronger claim: it proves the node that actually broke is identifiable. Confirmed it fails before the change and passes after.

**3. `answering()` test helper** now passes the third `attempts` argument #662 added to `tag`.

Gates on the rebased head `360541da7`: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` **0 findings**, `--lib fabric::` **143/0**, `--test fabric_serve` **39/0**, `--test fabric_end_to_end` 14/0, `--bin camelid` 40/0, public-scrub clean, `git diff --check` clean, all LF.

karan68:
Force-pushed `360541da7` -> `c96ea669e`. **The previous head was broken and this is my mistake, so I want to be exact about what happened.**

## What went wrong

After rebasing this branch onto `8974b4dcb` (main with #662 and #667 in it), I found two things this PR had to reconcile with #662, made both edits, ran the full gate suite, saw it green, and pushed — but I pushed the **commit**, and the two edits were still sitting **uncommitted in the working tree**. My gate script read the working tree, so it reported green for code that was never in the pushed SHA.

CI caught it independently: `360541da7` failed all three `rust` legs and `ci-gate`. Nothing reached `main`.

## The two reconciliations, and why they belong to this PR

Both are consequences of this PR's contract change meeting #662, which landed first. Neither is visible while either PR is tested alone, which is why both were green in isolation.

**1. `tag()` arity.** #662 gave `tag` a third parameter, `attempts: usize`. This PR's test helper `answering()` in `src/fabric/server.rs` still called the two-argument form — `E0061`.

**2. The 502 node-tagging contract.** This is the substantive one. #662 added `a_node_that_took_the_request_is_not_asked_to_run_it_again_elsewhere`, which asserts a 502 carries **no** `x-camelid-fabric-node`:

```rust
assert_eq!(header(&headers, "x-camelid-fabric-node"), None);
```

The whole point of this PR's review fix was the opposite: `forward_error` now moves the label out of the error onto the answer, because a failure that does not name the node is exactly the line an operator gets from the workflow this PR advertises (`RUST_LOG=camelid=warn`) and it read `node="-"`.

So the two PRs assert contradictory contracts. I believe **this PR is right** — a forward failure should name the node it happened against — so I updated #662's assertion rather than dropping the fix:

```rust
// The refusal names the node it happened against, so an operator can act on
// the one that is actually broken.
assert_eq!(header(&headers, "x-camelid-fabric-node"), Some("node-a"));
```

That is also the stronger assertion: it pins *which* node is named, not merely that the header is absent. The test's other assertion — that `node-a` served the request exactly once, which is why it may not be re-sent — is untouched, so #662's actual safety property is unchanged.

## Gates on `c96ea669e`, with a clean tree

| gate | result |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy --all-targets -- -D warnings` | **0 findings** |
| `cargo test --lib fabric::` | **143 passed / 0 failed** |
| `cargo test --test fabric_serve` | *(see below)* |
| `cargo test --test fabric_end_to_end` | 14 / 0 |
| `cargo test --bin camelid` | 40 / 0 |
| `scripts/check-public-scrub.sh` | 0 |
| `git diff --check` | 0 |
| line endings | LF |

## What I changed about my own process

The gate script now refuses to run against a dirty tree, and prints `tree_clean_at_end` and the `gated_head` it actually measured, so a green log can no longer describe something other than the commit being pushed. I also verified the fixes by reading them back out of the commit (`git show HEAD:<file>`) rather than off disk.

PR #668, opened 2026-08-15
feat(fabric): make the proxy safe to restart and its log traceable

karan68:
## What this is

Three things an operator needs from a resident proxy. They ship together because they share one call site — the `axum::serve` in `serve_on` — not because they are one feature.

**1. A stop is no longer a kill.** Ctrl-C, or SIGTERM where there is one, closes the listener so no new request is accepted, and lets the requests already in flight finish. Today `fabric serve` drops them: every in-flight generation becomes a client-visible failure on an ordinary restart.

**2. Each access-log line names the client it came from.** This needs `ConnectInfo`, which is configured on that same `axum::serve` call.

**3. Each request has an id**, logged and answered to the client, so a complaint can name the line that recorded it.

## The bound on the drain, and who it is for

Draining is bounded by `forward_timeout`, because that is already this proxy's answer to "how long may a request legitimately take" rather than a new number to tune.

Worth being precise about which requests that bound is actually for. A **buffered** request cannot outlive it — the same value bounds the forward, so the request ends first either way. A **stream** can: its idle timeout resets with every event, so a client reading slowly could otherwise hold the process open for as long as it kept reading. An orchestrator's own grace period is usually shorter than either and simply wins.

That is also why the bound has no automated test: it is only reachable on a stream, and a test for it would be a timing race with no useful margin. It is exercised in the live receipt instead.

## Why an inbound id is checked rather than cleaned up

An inbound `x-request-id` is honoured so an id assigned upstream survives the hop. It is validated first, because it lands in a log line the caller does not otherwise control: non-empty, `<= 128` bytes, and every byte `is_ascii_graphic` (which excludes space and every control character).

HTTP framing rejects a raw CR or LF before this ever sees it, so the reachable problems are the quieter ones — a tab or DEL that makes a line hard to parse, and a length nothing bounds. The docs say that rather than claiming to prevent log injection.

Anything failing the check gets a **fresh uuid, not a sanitised value**. A cleaned-up id is no longer the caller's id, so it would correlate with nothing while still looking as though it did.

## What this id does not do

It correlates a complaint with **this proxy's** log line. It does not correlate with the node's, because the engine neither accepts nor emits `x-request-id` — verified, zero occurrences. The module docs say so explicitly so nobody reads "request id" as end-to-end tracing.

## The one new seam

`serve_on` now delegates to `serve_on_until(listener, fabric, config, stop)`, which takes the stop as an argument. `serve_on` passes the real signal handler. Without that seam the shutdown path could only be tested by signalling the test runner itself.

## Tests, and what each one is worth

`fabric::` 143 → **149**, `fabric_serve` 39 → **41**. Every new test was ablated against the defect it claims to catch:

| Ablation | Result |
|---|---|
| `is_usable_request_id` always returns true | **exactly 2** fail: `an_unusable_id_is_replaced_rather_than_cleaned_up`, `what_counts_as_a_usable_id` |
| `into_make_service_with_connect_info` → `into_make_service` | **exactly 1** fails: `a_request_is_logged_with_the_caller_and_the_id_it_was_given` — and **all 149 `fabric::` unit tests still pass** |
| graceful shutdown → abrupt stop | **exactly 1** fails: `a_stop_finishes_the_work_in_flight_and_accepts_no_more` |

Two of those are worth spelling out, because they changed the tests.

**The `ConnectInfo` ablation passes every unit test.** The unit tests put `ConnectInfo` into the request extensions themselves, so they structurally cannot tell whether the server supplies it. Only a request over a real socket can. That is why there is a test which installs a capturing `tracing` subscriber and asserts the actual log line — it is the only thing that holds the wiring in place.

**My first shutdown test was vacuous and passed under its own ablation.** `axum::serve` spawns a task per accepted connection, so dropping the serve future leaves in-flight requests running *inside the test process*; asserting "the request still got a 200" proved nothing. What loses the request in production is the caller returning early and the process exiting under it. The test now asserts the ordering that actually matters — the stop must not report itself complete until in-flight work is — measured on a single task to avoid a cross-task race. Under the ablation it fails with:

> the proxy called itself stopped after **302.1µs**, while it still owed a request about **400ms** of work

A ~1300x separation, so it is not timing-fragile.

## Gates

`cargo fmt --all --check` clean · `cargo clippy --all-targets -- -D warnings` **0 warnings** · `--lib fabric::` **149 pass / 0 fail** · `--test fabric_serve` **41/0** · `--test fabric_end_to_end` **14/0** (unchanged) · `--bin camelid` **40/0** (unchanged) · `scripts/check-public-scrub.sh` exit 0 · `git diff --check` clean · LF endings on all three files.

## Honest gaps

- **`stop_requested()` itself is untested.** It is the thin wrapper that awaits `ctrl_c()` or SIGTERM. A test that raised a real signal would kill the test runner. Everything below it — the drain, the ordering, the refusal of new connections — is tested through `serve_on_until`.
- **The drain bound has no automated test**, for the reason given above.
- **Log lines still only appear when `RUST_LOG` asks for them**, the same as the rest of this binary. `RUST_LOG=camelid=info` is the whole access log. Not changed here.

## Stacking

#666 has merged, so this no longer carries it: it now stands alone on `main` at **3 files, +407/-10**.

Rebased onto `main` (`49114d3f4`). The rebase changed none of its own work -- `git diff` between the pre-rebase and rebased head is empty. Gates were re-run on the rebased **commit** (clean tree), not on a working tree.

Verified with `git merge-tree`: this still **conflicts with #665** in `src/fabric/server.rs` -- one mechanical hunk where both add to the same router/serve region. #665 will be rebased over this one.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -47,7 +47,8 @@
 //! A `session_start` with no matching `session_exit` means the process did not
 //! leave through that path. Be precise about what that does and does not say:
 //! `camelid serve` installs no signal handler (there is no `ctrl_c` or
-//! `with_graceful_shutdown` anywhere in the engine), so an ordinary Ctrl-C ends
+//! `with_graceful_shutdown` in its path — `fabric serve` has one, but that is a
+//! different command in a different process), so an ordinary Ctrl-C ends
 //! the process exactly as abruptly as an out-of-memory kill, and the two are
 //! recorded identically — which is to say, not at all. Read a missing exit
 //! record as "did not fail on the way out", never as evidence of a kill.
--- a/src/fabric/server.rs
+++ b/src/fabric/server.rs
@@ -36,7 +36,13 @@
 //! * Every request is recorded through `tracing` at INFO, and a failure at
 //!   WARN, which means nothing is written unless `RUST_LOG` asks for it — the
 //!   same as the rest of this binary. `RUST_LOG=camelid=info` is the whole
-//!   access log; `RUST_LOG=camelid=warn` narrows it to what failed.
+//!   access log; `RUST_LOG=camelid=warn` narrows it to what failed. Each line
+//!   carries the client it came from and an `x-request-id`, which is also
+//!   answered to the client — the node does not log that id, so it correlates
+//!   a complaint with this proxy's line, not with the node's own.
+//! * A stop (Ctrl-C, or SIGTERM where there is one) closes the listener and
+//!   lets the requests already in flight finish, bounded by `forward_timeout`.
+//!   A kill still drops them: this covers being asked to stop, not being shot.
 //! * A node that becomes ready, or loads a different model, is routed to only
 //!   once the current observation expires. A node that *stops* answering is not
 //!   waited on: the request that finds it gone is placed on another node
@@ -57,7 +63,7 @@ use std::time::{Duration, Instant};
 
 use axum::body::Body;
 use axum::extract::rejection::JsonRejection;
-use axum::extract::{DefaultBodyLimit, Request, State};
+use axum::extract::{ConnectInfo, DefaultBodyLimit, Request, State};
 use axum::http::header::{CACHE_CONTROL, CONTENT_TYPE};
 use axum::http::{HeaderMap, HeaderValue, StatusCode};
 use axum::middleware::Next;
@@ -72,6 +78,12 @@ use crate::api::{ApiAuth, DEFAULT_MAX_REQUEST_BODY_BYTES};
 
 /// Optional header a client sends to request affinity to a specific node.
 const STICKY_HEADER: &str = "x-camelid-fabric-sticky";
+
+/// The header a request id arrives on, and is answered with.
+const REQUEST_ID_HEADER: &str = "x-request-id";
+
+/// Longest inbound request id this proxy will adopt as its own.
+const MAX_REQUEST_ID_LEN: usize = 128;
 
 /// The credential a client must present to this proxy.
 ///
@@ -192,9 +204,15 @@ pub fn router(fabric: Fabric, config: ServeConfig) -> Router {
 async fn access_log(request: Request, next: Next) -> Response {
     let method = request.method().clone();
     let path = request.uri().path().to_string();
+    let client = request
+        .extensions()
+        .get::<ConnectInfo<SocketAddr>>()
+        .map(|ConnectInfo(peer)| peer.to_string())
+        .unwrap_or_else(|| "-".to_string());
+    let request_id = correlation_id(request.headers());
     let started = Instant::now();
 
-    let response = next.run(request).await;
+    let mut response = next.run(request).await;
 
     let elapsed_ms = started.elapsed().as_millis() as u64;
     let status = response.status();
@@ -209,6 +227,8 @@ async fn access_log(request: Request, next: Next) -> Response {
             node,
             reason,
             elapsed_ms,
+            client,
+            request_id,
             "fabric request failed"
         );
     } else {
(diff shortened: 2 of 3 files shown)


karan68:
## Live receipt — two real machines, real model, real signals

Everything below is from built binaries at this PR's head, not from tests.

**Provenance.** Windows proxy `camelid v0.6.1-126-g62c2efb7a`. Mac node+proxy built from the same commit, reporting `"build":"0.6.1+62c2efb7adf4"` (a `--depth 1` clone has no tags, hence the different version string from the same source).

**Setup.** Windows node `127.0.0.1:8181` (RTX 4060, 16/16 layers resident) · Mac node `192.168.29.158:8181` (Apple M5 Pro, Metal) · same GGUF, 1,321,082,528 bytes on both · Llama 3.2 1B Instruct.

### The request id, over a real socket

Four requests through the proxy with `RUST_LOG=camelid=info`:

| case | sent | answered + logged |
|---|---|---|
| none supplied | — | `be42528e-41eb-4694-988d-0b5c82e6a498` |
| client supplies one | `receipt-known-id-001` | `receipt-known-id-001` |
| 146 chars (over the 128 bound) | `xxxx…` | `c866e517-8078-4e8e-969f-e206828c4c93` |
| contains a space | `has a space` | `ff6e4cc0-3804-4680-b259-fef739fb8a76` |

An unusable id is **replaced, not cleaned up** — the value that comes back shares nothing with what was sent, so it cannot be mistaken for the caller's own id.

### The client address is real, not loopback-only

The Mac called the Windows proxy across Wi-Fi. The line the Windows operator gets:

```
fabric request method=POST path=/v1/chat/completions status=200 node="mac"
  reason="LeastLoaded" elapsed_ms=152
  client="192.168.29.158:64780" request_id="sent-from-the-mac-over-wifi"
```

A genuine remote peer. This is the part no unit test can reach — the unit tests supply `ConnectInfo` themselves.

### A stop finishes the work in flight — Windows, real Ctrl-C

Real `CTRL_C_EVENT` delivered to the proxy's console. The request was confirmed to be **on the node** first (the node's own `engine_queue_depth` reached 1), not merely sent:

```
REQUEST_OBSERVED_ON_NODE      = true (engine_queue_depth>=1 after 158ms)
STILL RUNNING WHEN STOP SENT  = true
INFLIGHT FINISHED             = 5,570 ms AFTER the stop
                              → HTTP/1.1 200 OK, 7,917-byte body from the model
                              → x-request-id: inflight-across-the-stop  (preserved across the shutdown)
NEW CONNECTION AFTER STOP     = REFUSED (actively refused 127.0.0.1:8490)
PROXY EXITED                  = true
```

### Same thing under a real SIGTERM — macOS

The Windows box cannot exercise the `#[cfg(unix)]` branch at all, so the proxy was built and run on the Mac and sent a real `kill -TERM`. **The log ordering is the whole feature:**

```
11:11:32.006  INFO  stopping: no longer accepting requests, finishing the ones in flight
11:11:34.107  INFO  fabric request ... status=200 node="mac" elapsed_ms=2274
                    request_id="inflight-across-sigterm"
```

The stop is acknowledged first; the request finishes 2.1 s later with a 7,897-byte body. New connection refused, process exited.

### The drain bound, which has no automated test

Stated in the PR: the bound is unreachable by a buffered request, because `forward_timeout` ends that request first either way. It is reachable by a **stream**, whose idle timeout resets on every event. Both arms below are the identical request against the identical config (`--forward-timeout-s 3`); only the stop differs.

| arm | duration | events | terminal `[DONE]` |
|---|---|---|---|
| control, never stopped | **11.51 s** | **2003** | yes |
| SIGTERM once flowing | **3.04 s** after the stop | **553** | no |

```
11:12:38.279  INFO  stopping: no longer accepting requests, finishing the ones in flight
11:12:41.280  WARN  in-flight requests did not finish in time; stopping without them drain_seconds=3
```

3.001 s apart — the bound firing at exactly its configured value, on the one request shape that can reach it.

### Housekeeping

CI **9/9 green** on `62c2efb7a`. Both machines returned to idle: no `camelid` processes, ports 8181/8490/8491 free on each, working tree clean.

PR #671, opened 2026-08-18
feat(fabric): let the proxy present a certificate

karan68:
## What this is

`fabric serve` is the address we tell an operator to expose, and `--api-key` is
the flag that is supposed to make that safe. Without TLS that key is sent in the
clear on every request, so the credential protecting the bind is exactly what the
bind gives away — along with every prompt and every completion.

The engine behind the proxy has been able to serve HTTPS all along
(`src/api/mod.rs`). Its front door could not. This closes that.

## What changed

`--tls-cert` / `--tls-key` now do for `fabric serve` what they already do for
`camelid serve`. The pair and the one rule about it — half a pair is a mistake,
not a request for cleartext — move into a small shared `src/tls_pair.rs` that
both front doors import, so they cannot drift and neither reaches into the other.

Worth saying why it is a new module rather than a re-export from
`src/api/mod.rs`, which is where #658 put `resolve_api_key`: that file's git blob
SHA-1 is pinned by the SmolLM3 qualification
(`scripts/hf-qualification-smollm3-chat-parity.mjs` runs
`git hash-object -- src/api/mod.rs`). My first attempt added four lines there and
turned `validation-scripts` red. Refreshing the provenance chain is a documented
procedure, but a TLS change has no business editing a qualification fixture. This
branch leaves `src/api/mod.rs` byte-identical to `main`.

A routable bind now has two independent refusals instead of one. Anonymity and
cleartext are different risks, so each gets its own acknowledgement:

| bind | anonymous | authenticated |
| --- | --- | --- |
| loopback | serves | serves (unchanged) |
| routable, cleartext | refused (unchanged) | **refused — this is new**: needs `--tls-cert`/`--tls-key` or `--allow-cleartext-remote` |
| routable, TLS | refused (unchanged) | serves |

Acknowledging one does not accept the other. That is what
`neither_guard_satisfies_the_other` exists to pin, and it is the test I would
most want a reviewer to read.

**The bold cell is a breaking change.** `fabric serve --addr 0.0.0.0:8282
--api-key ...` served before this branch and exits 1 after it, and
`docs/CONFIGURATION.md` said in as many words that a key "satisfies the bind
guard on its own". That section is rewritten here rather than left to contradict
the binary; it now states both refusals and shows what an exposed proxy looks
like.

### The certificate is read before anything is announced

`ProxyTls` holds the loaded `RustlsConfig` rather than the paths it came from,
and `ProxyTls::resolve` is where the pair is read — while clap is still resolving
arguments, before the listener is bound, before the listening line is printed and
before `startup_report` probes a single node.

That ordering is not cosmetic. `fabric serve` announces its address between
binding and serving, so reading the certificate at serve time told an operator
who mistyped a path:

```
fabric serve listening on https://127.0.0.1:8491
fabric: 0 of 1 nodes ready; no model is being served
  a (127.0.0.1:9) unreachable: cannot connect ...
Error: could not load TLS certificate/key: ... (os error 2)
```

which is the thing the comment above that print exists to prevent, and the
opposite of what the engine does — `api/mod.rs` loads the pair at :2705 and calls
`print_ready_banner` at :2797. It also weakened the guard above:
`refuse_cleartext_remote` accepts a certificate as an answer to cleartext, and "a
certificate" meant *two paths were given*, so a routable bind with an unreadable
one passed the guard and opened the socket before failing.

Reading it first makes `Some` mean a certificate that will actually be presented,
and leaves `serve_tls` with no failure of its own. Measured on the built binary,
the same command both times:

|  |  |
| --- | --- |
| before | stdout: the listening line and the startup report; stderr: the error |
| after | stdout: **empty**; stderr: the error |

and, after, with files that exist and are not a PEM pair on
`--addr 0.0.0.0 --api-key`: refused in 90 ms, before the bind, with
`The private key file contained no keys`.

## Two things the TLS path had to keep

`axum::serve` cannot present a certificate, so the TLS path is `axum_server` —
already a dependency, already how the engine serves TLS. That is a different
crate driving the accept loop, and two behaviours from #666/#668 had to survive
the switch:

* The access log still names the caller. `axum_server` supplies `ConnectInfo`
  only if asked, and no unit test can see that wiring — the unit tests insert the
  extension themselves. Only a request over a real TLS socket can. Worth noting
  the engine's own TLS path uses plain `into_make_service()` and therefore has no
  peer address; the proxy deliberately does not copy that.
* A stop still drains **and still stops accepting**. `axum_server` owns its
  shutdown, so the bound is handed to `Handle::graceful_shutdown(Some(drain))`
  rather than raced against a sleep, and both paths report an expired drain in
  the same words. Both halves are asserted: they are that crate's accept loop and
  that crate's shutdown, so neither carries over from the cleartext test.

## Tests

`--lib fabric::` 164 → 170, `fabric_serve` 56 → 60. Nothing removed.

Tests mint a throwaway certificate rather than committing a private key, and
speak real TLS to the proxy rather than trusting that an untested branch serves
what it claims. The unreadable-certificate test covers both a certificate that is
not there and — the likelier mistake — files that exist and are not a PEM pair.

Ablated on the head of this branch, each against the mechanism it claims to
catch:

|  |  |
| --- | --- |
| cleartext guard always allows | exactly **2** fail: `a_cleartext_non_loopback_listener_is_refused_by_default`, `neither_guard_satisfies_the_other`. 168 unit tests pass — the loopback, certificate and acknowledged controls keep passing, as they must — and `fabric_serve` is untouched at 60/0 |
| TLS branch serves cleartext | exactly **3** fail, all three of them the ones that speak real TLS to a socket: `a_tls_listener_places_a_request_and_relays_the_answer`, `a_tls_request_is_logged_with_the_caller_it_came_from`, `a_tls_stop_finishes_the_work_in_flight_and_accepts_no_more`. 57 pre-existing integration tests still pass, and all 170 unit tests do |

The unreadable-certificate test is deliberately not in the second set: after the
change above it never reaches the serving path at all, which is the point of it.
That a configured certificate cannot degrade to cleartext is now structural
rather than tested — `ServeConfig.tls` can only be `Some` once the certificate
has loaded.

The second ablation found a defect in my own tests. A TLS client and a cleartext
server deadlock rather than fail — hyper waits for a request line it will never
recognise in the ClientHello bytes while the handshake waits for a ServerHello.
The first run hung for 78 seconds on one second of CPU and reported nothing. In
CI that burns the job timeout instead of naming the broken assertion. Every step
of the TLS client is now bounded. Only after that did the ablation produce a
clean "exactly 3 failed, in 12.36s".

## Live receipt, built binary

`v0.6.1-142-g979f251bf`, one node deliberately unreachable, certificate minted
with openssl (`CN=localhost`, SAN `DNS:localhost,IP:127.0.0.1`):

```
curl --cacert <chain> https://localhost:8493/v1/health
  -> 503, ssl_verify_result=0, HTTP/1.1, real readiness body
     {"ok":true,"service":"camelid-fabric","ready":false,
      "nodes":{"total":1,"ready":0,"not_ready":0,"unreachable":1}, ...}

curl http://localhost:8493/v1/health          -> curl exit 1, no response
curl https://localhost:8493/v1/health         -> curl exit 60 (certificate not
   (without trusting the minted CA)               trusted), no response
```

`--api-key` on `0.0.0.0` with no certificate exits 1 with a message naming both
`--tls-cert` and `--allow-cleartext-remote`, and prints no listening line. Half a
pair still exits 1 with `--tls-cert and --tls-key must be provided together`.

## The one judgement call for a maintainer

This adds three dev-only crates: `rcgen`, `pem`, `yasna` (locked 616 → 619).
Measured, not estimated.

* Nothing reaches the shipped binary — `[dev-dependencies]` only.
* No new crypto backend. `rcgen` and the rustls test client both use the
  `aws-lc-rs` the TLS build already pulls in; `rustls`, `tokio-rustls` and
  `rustls-pki-types` as dev-deps added zero crates because they resolve to
  versions already in the lock.
* The alternative was committing a private key to the tree, or leaving the
  headline feature with no automated test. Happy to take a different route if you
  would rather.

One oddity that is deliberate: the test writes
`rcgen::Certificate::pem(&issued.cert)` rather than `issued.cert.pem()` because
`scripts/check-public-scrub.sh` forbids that method name in dotted form anywhere
in the tree — it is a privacy guard for operator key paths. There is a comment
saying so at the call site.

## Gates

`cargo fmt --all --check` clean · `cargo clippy --all-targets -- -D warnings` 0 ·
`--lib fabric::` 170/0 · `--test fabric_serve` 60/0 · `--test fabric_end_to_end`
14/0 · `--bin camelid` 40/0 · `scripts/check-public-scrub.sh` exit 0 ·
`cargo doc --no-deps` adds no new warning · `git diff --check` clean · all files
`i/lf w/lf`. Gates were run on the commit with a clean tree, not on a working
tree.

## Honest gaps

* **The hop to the nodes is unchanged, so the guard above is one-sided.** This
  encrypts client → proxy. The proxy speaks to its nodes over a plain socket with
  `Authorization: Bearer` written into the head, and there is no way to make it
  speak TLS. On a fabric whose nodes are not loopback that token, and every
  prompt, still cross the network in the clear whatever the proxy presents to its
  own clients. `docs/CONFIGURATION.md` now says that plainly rather than letting
  "the proxy does TLS" imply more than it means. Making the proxy a TLS *client*
  is the next slice.
* **The engine does not have this guard, so the two front doors now disagree
  about cleartext.** `ServerPolicy::resolve` refuses an anonymous routable bind
  and nothing else, so `camelid serve --addr 0.0.0.0:8181 --api-key ...` still
  serves cleartext — and `docs/CONFIGURATION.md` recommends exactly that command.
  Same risk, same binary, opposite answers. I did not change it here: that is a
  behaviour change to a shipped surface and deserves its own PR and its own docs.
  Say the word if you would rather they land together.
* No cross-machine receipt yet. The second machine is powered off, so the LAN arm
  — real certificate, real request from another host over Wi-Fi, plus a negative
  control showing the plaintext port is refused — is not taken. The automated
  tests cover the mechanism on loopback across all three CI OSes and the receipt
  above covers the built binary on loopback; the cross-machine claim is simply
  not made yet.
* The drain-expired warning on the TLS path has no automated test, the same gap
  #668 declared for the cleartext one. It is only reachable on a request that
  outlives the bound.
* Still one key, still not per-client. Rotation and revocation are the next
  slice, not this one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,8 +492,11 @@ dependencies = [
  "minijinja",
  "ratatui",
  "rayon",
+ "rcgen",
  "rusqlite",
  "rust-embed",
+ "rustls",
+ "rustls-pki-types",
  "rustyline",
  "serde",
  "serde_json",
@@ -502,6 +505,7 @@ dependencies = [
  "thiserror 1.0.69",
  "tokenizers",
  "tokio",
+ "tokio-rustls",
  "tokio-util",
  "toktrie",
  "tower",
@@ -3169,6 +3173,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2ee67f1008b1ba2321834326597b8e186293b049a023cdef258527550b9935b4"
 
 [[package]]
+name = "pem"
+version = "3.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d30c53c26bc5b31a98cd02d20f25a7c8567146caf63ed593a9d87b2775291be"
+dependencies = [
+ "base64 0.22.1",
+ "serde_core",
+]
+
+[[package]]
 name = "percent-encoding"
 version = "2.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3511,6 +3525,19 @@ checksum = "22e18b0f0062d30d4230b2e85ff77fdfe4326feb054b9783a3460d8435c8ab91"
 dependencies = [
  "crossbeam-deque",
  "crossbeam-utils",
+]
+
+[[package]]
+name = "rcgen"
+version = "0.13.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75e669e5202259b5314d1ea5397316ad400819437857b90861765f24c4cf80a2"
+dependencies = [
+ "aws-lc-rs",
+ "pem",
+ "rustls-pki-types",
+ "time",
+ "yasna",
 ]
 
 [[package]]
@@ -6348,6 +6375,15 @@ dependencies = [
  "libc",
  "once_cell",
  "pkg-config",
+]
+
+[[package]]
+name = "yasna"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e17bb3549cc1321ae1296b9cdc2698e2b6cb1992adfa19a8c72e5b7a738f44cd"
+dependencies = [
+ "time",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,6 +156,14 @@ alloc-gate = []
 [dev-dependencies]
 tower = { version = "0.5", features = ["util"] }
 tempfile = "3"
+# Test-only, and all three already resolve to versions the TLS build pulls in:
+# these let the fabric TLS tests mint a throwaway certificate and speak real
+# HTTPS to the proxy, rather than committing a private key or trusting that an
+# untested branch serves what it claims.
+rcgen = { version = "0.13", default-features = false, features = ["aws_lc_rs", "pem"] }
+rustls = { version = "0.23", default-features = false, features = ["aws-lc-rs"] }
+rustls-pki-types = "1"
+tokio-rustls = { version = "0.26", default-features = false, features = ["aws-lc-rs"] }
 
 [lints.rust]
 # The objc crate's msg_send!/sel_impl! macros (used for Metal command-buffer GPU
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -96,8 +96,15 @@ and CUDA VRAM. It contains no model-path, prompt, API-key, or per-user labels.
 ## Fabric proxy policy
 
 `camelid fabric serve` puts one HTTP address in front of several independent nodes. It defaults to
-`127.0.0.1:8282` and, like `camelid serve`, refuses a non-loopback address unless the exposure is
-acknowledged with `--allow-unauthenticated-remote` (or `CAMELID_ALLOW_UNAUTHENTICATED_REMOTE`):
+`127.0.0.1:8282`. A non-loopback address has to answer two separate questions, because they are two
+separate risks, and answering one does not answer the other:
+
+| | anonymous | authenticated |
+|---|---|---|
+| **cleartext** | refused: needs both acknowledgements | refused: needs `--tls-cert`/`--tls-key` or `--allow-cleartext-remote` |
+| **TLS** | refused: needs `--api-key` or `--allow-unauthenticated-remote` | serves |
+
+Loopback is unaffected and needs neither.
 
 ```bash
 target/release/camelid fabric serve \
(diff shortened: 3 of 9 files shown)


karan68:
Second commit, `d14b3c26d`, after reviewing my own first one against the built binary rather than the tests.

**It announced an https address it had not earned.** `fabric serve` prints its listening line and probes every node between binding and serving, and the certificate was read after that. A mistyped path produced:

```
fabric serve listening on https://127.0.0.1:8491
fabric: 0 of 1 nodes ready; no model is being served
  a (127.0.0.1:9) unreachable: cannot connect ...
Error: could not load TLS certificate/key: ... (os error 2)
```

That is what the comment two lines above the print exists to prevent, and the engine does the opposite — `api/mod.rs` loads at :2705, `print_ready_banner` at :2797. Worse, it weakened the guard this branch adds: `refuse_cleartext_remote` treats a certificate as an answer to cleartext, and "a certificate" only meant *two paths were given*, so a routable bind with an unreadable one passed the guard and opened the socket.

`ProxyTls` now holds the loaded `RustlsConfig` and `ProxyTls::resolve` reads the pair while clap is still resolving arguments. Same command, before and after:

| | |
| --- | --- |
| before | stdout: the listening line and the startup report; stderr: the error |
| after | stdout: **empty**; stderr: the error |

Files that exist and are not a PEM pair, on `--addr 0.0.0.0 --api-key`: refused in 90 ms, before the bind, `The private key file contained no keys`.

**Three smaller things in the same commit.**

- `a_tls_stop_finishes_the_work_in_flight` was missing the half its cleartext twin has: a stopped proxy accepts no more. `axum_server` drives its own accept loop, so that does not carry over. Added, and it is in the ablation set below.
- `ProxyTls`' doc linked `crate::api::TlsFiles`, which this branch itself moved to `crate::tls_pair` — a new rustdoc broken-link warning on a public type, and a claim that was not true either.
- `docs/CONFIGURATION.md` still said an `--api-key` "satisfies the bind guard on its own". This branch makes that false: `fabric serve --addr 0.0.0.0 --api-key` now exits 1. The fabric section is rewritten — both refusals, an exposed-proxy example with a certificate, and a plain statement that TLS covers the client's hop only.

**Re-ablated on this head**, since the first commit's numbers no longer describe it:

| | |
| --- | --- |
| cleartext guard always allows | exactly **2** fail (`a_cleartext_non_loopback_listener_is_refused_by_default`, `neither_guard_satisfies_the_other`); 168 unit tests pass, `fabric_serve` untouched at 60/0 |
| TLS branch serves cleartext | exactly **3** fail, all three the ones that speak real TLS to a socket; 57 integration and all 170 unit tests pass |

The unreadable-certificate test is deliberately out of the second set now — it never reaches the serving path.

Gates on the commit with a clean tree: `--lib fabric::` 170/0, `--test fabric_serve` 60/0 in 12.31s, fmt clean, `clippy --all-targets -D warnings` 0, `check-public-scrub.sh` 0, `git diff --check` clean, all files `i/lf w/lf`.

Two things I did **not** change, both now in Honest gaps, because they are decisions rather than defects: the proxy still speaks plain HTTP with a bearer to its nodes, and the engine still has no cleartext guard at all, so the two front doors now disagree about the risk this branch exists for. Happy to take either as a follow-up or fold the engine one in here.

PR #672, opened 2026-08-18
feat(fabric): name the clients this proxy serves, and let one be revoked

karan68:
﻿## What this changes

`camelid fabric serve` took a single `--api-key`, so every client that could reach the proxy held the
same secret. Two consequences, both operational:

- the access log could only say that *someone* authenticated, never which client;
- withdrawing access from one client meant changing the key for all of them and restarting the proxy,
  which drops the requests everyone else has in flight.

`--client-keys <PATH>` takes a JSON file that names each client:

```json
{"clients": [
  {"name": "laptop", "key": "..."},
  {"name": "ci",     "key": "..."}
]}
```

The name is what the access log records, as `client_name`; the key never is. Deleting an entry
revokes that client without a restart and without disturbing anyone else.

## What is reused rather than rebuilt

Nothing in `src/fabric/client_keys.rs` compares secrets. Each client holds the engine's own
`ApiAuth`, so a presented credential is read from the same two headers (`Authorization: Bearer` or
`X-API-Key`) and compared with the same constant-time check the engine applies to its own listener,
and keys are validated through `crate::api::resolve_api_key`. The health-route exemption comes from
`ApiAuth::route_requires_auth` and the 401 from `ApiAuth::unauthorized`, both lifted out of the
engine's `authenticate` so the two front doors refuse in identical words. This module decides *who*,
never *how*.

## Two choices worth arguing with

**A re-read that fails keeps the previous set and says so, rather than refusing everyone.** A key file
is usually replaced by writing a new one and renaming it over the old, so a read landing mid-swap
sees a partial or absent file; refusing on that would turn an ordinary edit into an outage. The cost
is real and is stated where an operator will meet it: a revocation written into a file that does not
parse — or deleting the key file altogether — has *not* taken effect, and the previously loaded set
goes on being served until the file is readable again or the proxy restarts. So the proxy prints to
stderr when a re-read starts failing, and again when it recovers, rather than relying on `RUST_LOG`
being set:

```
fabric: could not reload client keys: <why>. The previous set of 2 clients is still in
force, so a revocation written here has NOT taken effect.
```

**An unusable set at startup stops the proxy rather than opening it.** A proxy that silently ignored
an unreadable key file would serve whatever can reach it. Unreadable, invalid JSON, empty, a
duplicated name, or two clients sharing a key are all refused before the listener binds.

## The one line in `src/api/mod.rs`, and the provenance re-pin

Please look at this part first.

The fabric no longer uses the crate-internal re-export of `authenticate`: it now holds a *set*, and a
single-key middleware cannot say which key answered. That re-export existed only for the fabric --
the engine's own router at `src/api/mod.rs:2575` was already calling `server::authenticate` directly
-- so the name is now dead, and CI denies warnings.

Removing a name moves the file's git blob, and `src/api/mod.rs` is pinned by the SmolLM3 chat-parity
qualification. The pin moves from `cb8e118e` to `c74f54dc`, and it is a chain of three files, not one
value:

| file | what changes |
| --- | --- |
| `qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json` | `implementation.source_git_blob_sha1` |
| `scripts/hf-qualification-smollm3-chat-parity.mjs` | `RENDERER_GIT_BLOB_SHA1`, **and** `GROUNDING_FILES.runtime_envelope.sha256` ΓÇö that fixture is itself a pinned grounding, so editing it moves its own digest |
| `scripts/test-hf-qualification-smollm3-chat-parity.mjs` | the same two values |

The new fixture digest was computed with the repo's own `canonicalGitTextBytes`, not an approximation.
Its canonical length is unchanged at 5,578 bytes, which is the check that the edit replaced forty hex
characters and nothing else.

I got this wrong on the first push and CI caught it: I had refreshed only `RENDERER_GIT_BLOB_SHA1`,
because `GROUNDING_FILES` does not list `src/api/mod.rs` and I stopped there. It does not need to ΓÇö
the fixture carries the blob sha1, and the fixture is what is pinned.
`scripts/test-check-model-qualification-roster.mjs` failed on exactly that, which is also the ablation
proving the pin is live rather than vestigial. All **42** `scripts/test-*.mjs` now pass locally, which
is what `ci.yml` runs in the `validation-scripts` job.

## Measurements

Test counts, against `upstream/main` at `d080367b4` (i.e. with #671's TLS work in). No test was
removed or renamed.

| target | before | after |
| --- | --- | --- |
| `--lib fabric::` | 170 | 181 |
| `--test fabric_serve` | 60 | 62 |
| `--test fabric_end_to_end` | 14 | 14 |
| `--bin camelid` | 40 | 41 |

Gates on the commit, clean tree: `cargo fmt --all --check` exit 0 - `cargo clippy --all-targets --
-D warnings` **0 findings** - `check-public-scrub.sh` exit 0 - `git diff --check` exit 0 - all
touched files LF.

Revocation latency is bounded by the re-read interval, which is one second, and the file is only
re-read when its size or modification time has changed. The integration test runs against that
shipped default rather than a test-only one, and asserts the revoked client is refused within 5s.

## Ablations

Every new test was checked against the behaviour it claims to catch. Each row is one edit to the
source, both suites re-run, then reverted; the failures listed are the complete set.

| ablation | unit failures | integration failures |
| --- | --- | --- |
| `current_clients` never re-reads the file | 2: `removing_a_client_revokes_only_that_client`, `a_broken_file_leaves_the_previous_set_in_force` | 1: `revoking_one_client_stops_it_without_a_restart_or_disturbing_the_rest` |
| the admitted name is not attached to the response | 0 (174 pass) | 1: `each_client_is_served_and_logged_under_its_own_name` |
| an unusable set falls back to `ClientAuth::none()` | 1: `an_unusable_set_stops_the_proxy_rather_than_opening_it` | 0 (58 pass) |
| only the first entry in the set is compared | 3 | 2 |
| `conflicts_with_all` dropped from `--client-keys` | 1 (`--bin camelid`): `a_client_key_set_cannot_be_combined_with_a_single_key` | - |

In every row the pre-existing tests all passed, so these are the new tests failing, not collateral.

One test I wrote and then deleted, in case it is worth knowing: an assertion that `match_client` does
not stop at the first match cannot fail, because the loader already refuses two clients sharing a
key, so at most one entry can ever match. The non-short-circuit is still there and the reasoning is
in the doc comment, but the test now asserts something observable instead -- that a client is
admitted under its own name wherever it sits in the set.

## What this does not do

- No per-client concurrency or rate limit. That is deliberate and separate.
- No cross-machine receipt: the second machine was unavailable while this was written. Everything
  above is loopback and in-process on Windows.
- The key set is a file the proxy watches. It is not a general configuration reload.

## Base

Rebased onto `d080367b4`, i.e. on top of #671. Git resolved four of the five overlapping files on its
own; three things it could not, and they are worth a look:

- **A silent compile break.** My commit removed `use std::path::PathBuf` from `src/fabric/server.rs`,
  because moving `ClientAuth::resolve` out made it unused. #671 then added `ProxyTls::resolve`, which
  needs it. Each side is correct alone and the merge is textually clean; the result did not compile.
  Restored the import.
- **#671's own bind-guard table went stale the moment this landed.** It listed `--api-key` or
  `--allow-unauthenticated-remote` as the ways to answer the anonymous question. `--client-keys` is
  now a third, so the table and the first bullet under it are updated.
- **A claim of mine went stale in the other direction.** I had written that a key set "satisfies the
  bind guard". After #671 there are two guards, and a key set answers only one of them. It now says
  so, and says an exposed proxy still needs a certificate or `--allow-cleartext-remote`.

The counts above were re-measured on the rebased commit, not carried over.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -172,9 +172,18 @@ The name, never the key, is written to every access-log line that client causes,
 Deleting an entry revokes that client without a restart and without disturbing anyone else. The file
 is re-read at most once a second, and only when its size or modification time has changed, so this
 costs nothing per request. If a re-read fails — which is what the moment of an atomic replace looks
-like — the previous set stays in force and a warning is logged, so an ordinary edit cannot become an
-outage. The cost of that choice is real and worth stating: a revocation written into a file that
-does not parse has *not* taken effect, and the warning is the only thing that says so.
+like — the previous set stays in force, so an ordinary edit cannot become an outage.
+
+That fallback has a cost worth stating plainly: **a revocation written into a file that does not
+parse, or deleting the key file altogether, does not revoke anybody.** The previously loaded set goes
+on being served until the file is readable again or the proxy is restarted — and a restart refuses to
+come up at all while the file is unusable. So the proxy prints to stderr when a re-read starts
+failing, and again when it recovers, rather than relying on `RUST_LOG` being set:
+
+```
+fabric: could not reload client keys: <why>. The previous set of 2 clients is still in force, so a
+revocation written here has NOT taken effect.
+```
 
 The proxy re-probes its nodes at most once per `--observation-max-age-ms` (500 by default) rather
 than once per request. Inside that window its view can be wrong, so a request placed on a node that
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -102,7 +102,7 @@ separate risks, and answering one does not answer the other:
 | | anonymous | authenticated |
 |---|---|---|
 | **cleartext** | refused: needs both acknowledgements | refused: needs `--tls-cert`/`--tls-key` or `--allow-cleartext-remote` |
-| **TLS** | refused: needs `--api-key` or `--allow-unauthenticated-remote` | serves |
+| **TLS** | refused: needs `--api-key`, `--client-keys`, or `--allow-unauthenticated-remote` | serves |
 
 Loopback is unaffected and needs neither.
 
@@ -130,11 +130,11 @@ that terminates TLS itself.
 
 Three limits are deliberate and worth knowing before you deploy it:
 
-- `--api-key` (or `--api-key-file`) is the key the proxy requires *from its clients*. It deliberately
-  reads no environment variable: on this command `CAMELID_API_KEY` already names the token sent to
-  nodes, and one value must not silently configure both directions. Without a key,
-  `--allow-unauthenticated-remote` is the only way to bind a routable address, and it should only be
-  used behind something that does authenticate.
+- `--api-key` (or `--api-key-file`, or `--client-keys` below) is the key the proxy requires *from its
+  clients*. It deliberately reads no environment variable: on this command `CAMELID_API_KEY` already
+  names the token sent to nodes, and one value must not silently configure both directions. Without
+  one of them, `--allow-unauthenticated-remote` is the only way to bind a routable address, and it
+  should only be used behind something that does authenticate.
 - `--bearer` (or `CAMELID_API_KEY`) is the token the proxy presents *to its nodes*, the same as
   `fabric status|route|run`. A node started with `--api-key` needs it: `/v1/health` is exempt from
   that node's auth, so without a token the node observes as ready and then answers every forwarded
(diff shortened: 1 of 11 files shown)


karan68:
Second commit, `70f3a8e8e`, after reviewing my own first one against the built binary rather than the tests.

## The fallback was silent, and silence was the whole risk

Keeping the previous set in force when a re-read fails is the right call, and I still think so — refusing every client because a file was caught mid-rewrite turns an ordinary edit into an outage. But it rests entirely on the operator being told, and my own PR body said so: *"the warning is the only thing that says so."*

That warning was `tracing::warn!`, and `RUST_LOG` is unset on a stock proxy. Measured on the built binary, the same situation twice, the only difference an environment variable:

| | |
| --- | --- |
| `RUST_LOG` unset | corrupt the key file → laptop still served. delete the key file entirely → laptop **and** ci still served. **stdout: nothing about it. stderr: nothing at all.** |
| `RUST_LOG=camelid=warn` | identical, and the WARN line appears |

So on a default proxy an operator could delete the whole key file and every client would go on being served, indefinitely, with nothing said anywhere.

A failed re-read now prints to stderr as well, once per transition into and out of the degraded state so a file left broken does not fill the terminal. The message is a pure `stale_key_set_notice`, unit-tested rather than eyeballed, and it names the consequence rather than only the error:

```
fabric: could not reload client keys: <why>. The previous set of 2 clients is still in
force, so a revocation written here has NOT taken effect.
fabric: client keys reloaded from <path>; now serving 1 client
```

Re-measured after, no `RUST_LOG`, one proxy: corrupt → exactly one stderr line; delete → no second line, same degraded state; restore with a client removed → the recovery line, that client 401, the other still 200.

## Three claims that were not quite true

- **A reload was only recorded when the client count changed.** Rotating a compromised key — same count, different secret — left no trace at any log level. Every successful reload is recorded now; the stamp check already means this only fires when the file really moved.
- **`FileStamp`'s doc said length plus mtime covers a rewrite inside one timestamp tick.** It narrows that window, it does not close it: a key swapped for one of the same length inside a tick still looks unchanged. Closing it would mean reading the file every interval, which is the cost the stamp exists to avoid — so the doc now says what it actually buys instead of implying more.
- **The name rule is `is_ascii_graphic`, which excludes spaces**, but the refusal said only "not printable ASCII". A space is printable. Someone naming a client `build server` would have been told something untrue about why.

`docs/CONFIGURATION.md` gains the deletion case explicitly. "Deleting an entry revokes that client" invites "deleting the file revokes everyone", and the opposite is what happens.

Worth saying, so it is clear these were looked at rather than skipped.

- **The provenance re-pin is correct, verified independently of the scripts.** `git hash-object src/api/mod.rs` = `c74f54dc3cfff63dc86c2529d94cea39132b8600`, which is what the fixture and both scripts carry. The fixture's own canonical digest, computed with the repo's `canonicalGitTextBytes`, is `3c5d3891…d91fbb` at **5,578** bytes — matching `GROUNDING_FILES.runtime_envelope.sha256` and `GROUNDING_FILE_SIZE_BYTES`. The size pin is the check that the edit moved forty hex characters and nothing else, and it holds.
- **The `src/api/server.rs` extraction is behaviour-preserving.** `route_requires_auth` and the 401 body move verbatim into `ApiAuth`; `authenticate` is otherwise untouched, so the engine's own listener refuses exactly as before.
- **The admitted name cannot be forged and never reaches the client.** It rides a response *extension*, not a header, `AdmittedClient` is private to `fabric::server`, and the auth layer sets it after the inner handler returns — so nothing downstream, node included, can put one there.
- **`match_client` not short-circuiting is load-bearing and correctly argued**: the loader refuses two clients sharing a key, so at most one can match, and stopping early would make the answer's timing depend on where in the file a client sits.
- **Health stays exempt**, so a load balancer polling `/v1/health` neither presents a key nor drives a re-read.

## One thing left alone, on purpose

`admit()` takes a `std::sync::Mutex` and, on the once-a-second tick, does a `stat` — and occasionally a read and parse — on the executor thread while holding it. For a small local file that is microseconds and the single-flight is deliberate. On a key file that lives on a stalled network mount it is not, and every concurrent request queues behind it. Moving the re-read to a background task would fix it and is a bigger change than this PR should carry; naming it here so it is a known cost rather than a surprise.

## Gates

Re-measured on the commit with a clean tree: `--lib fabric::` **181** (180 + the notice test), `--test fabric_serve` 62, `--test fabric_end_to_end` 14, `--bin camelid` 41, `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` **0**, `check-public-scrub.sh` 0, `git diff --check` clean, touched files `i/lf w/lf`.

PR #673, opened 2026-08-18
feat(cuda): GPU-native fused filtered sampling & lazy host KV materialization

karan68:
## Summary

This PR resolves a major throughput bottleneck during non-greedy sampling on CUDA and eliminates redundant synchronous host memory copies during prompt prefill:

1. **GPU-Native Fused Filtered Sampling**:
   - Implemented  in  providing parallel on-device logit filtering (, , ) with Gumbel noise perturbation and reduction.
   - Connected  directly to the decode forward pipeline in  so non-greedy sampling ( > 0, 	ext{min\_p} > 0$) executes 100% on the GPU with **zero device-to-host logit transfers** (saving ~512 KB PCIe round-trip per token on large vocabularies).
   - Added unit test  in .

2. **Lazy Host KV Materialization on GPU Prefill**:
   - Removed eager synchronous D2H KV cache copies from , relying on  and  on-demand if a CPU fallback or prompt cache export occurs.

3. **Platform Repack Safety**:
   - Guarded  in  with  to ensure Linux x86 hosts do not attempt ARM repacking.
   - Ensured execution planner explicitly disables CPU repack in  when CUDA resident decode is active.

## Verification & Benchmarks

- **Test Suite**: Ran  on an NVIDIA L4 GPU (), verifying  and .
- **Live Serving**: Verified  with TinyLlama Q8_0 on NVIDIA L4 GPU handling live streaming chat completions () with temperature and min-p sampling.


timtoole02 (commented):
## Most of this is already on `main`

The flash-prefill half of this branch is #670, merged 2026-08-19 — commit `d964daa4` here is that same commit. `main` has since hardened it through review:

- `3a4bacd5` fix(cuda): address PR 670 review — opt-in contract, tile-level online softmax, and register unrolling
- `1f62e0b8` fix(cuda): assert head_dim <= 256 in launch_attention_flash_prefill
- `bdfc5e97` fix(cuda): make the flash prefill claims true and stop the chunk knob panicking
- head_dim-specialized kernels `flash_attention_prefill_tiled_d64/_d128/_d256/_dyn`, templated on `constexpr int D_STEPS = HEAD_DIM / 32`

`CpuTensor::q8_0_raw_bytes()` / `Q8_0PackedRows4::to_q8_0_bytes()` and the `execution_plan.rs` repack-off insert are on `main` too.

So merging this branch as-is would **revert** those fixes. Two concrete regressions:

**1. `flash_prefill_enabled()` goes back to default-on.** `main` has `std::env::var("CAMELID_FLASH_PREFILL").is_ok_and(...)` — opt-in, which is the contract restored in #670's review and the one `scripts/check-cuda-prefill-parity-gate.mjs` states in writing ("flash prefill is opt-in, prefill-only, token-parity"). This branch has `.unwrap_or(true)`. The dispatch at `src/cuda_resident.rs:13863` also drops `k <= MAX_VERIFY_K` and `base_position + k > SPLITK_THRESHOLD`. Since `flash_ok` is hardcoded `true` at the `prefill_batched` call site and `prefers_batched_prefill()` is true for any all-Q8_0 resident engine, that puts every stock CUDA user on the promoted Q8_0 lane onto the new kernel with no env var set. The parity gate still passes because it only string-matches; it never checks the env default.

**2. The `head_dim` bounds are gone.** `main` carries `assert!(head_dim <= 256, "flash prefill requires head_dim <= 256")` in `launch_attention_flash_prefill` plus `&& head_dim <= 256` at the dispatch. This branch has neither, and its `float o_acc[2][8]` hardcodes the accumulator depth `main` now derives per-head_dim. No supported row reaches it today (gemma3 is excluded by the windowed-arch guards, gemma4 runs its own runtime), so this is a robustness regression rather than a live bug — but it is a regression against code already in the tree.

The branch is also 156 commits behind `main` and no longer merges clean — conflicts in `src/cuda_resident.rs` and `src/cuda_resident/tests.rs`, the two files it rewrites most heavily.

## What's genuinely new, and it's good

Two things, neither on `main`: `sample_gumbel_filtered` (GPU-native min-p) and the lazy host-KV mirror.

The sampling kernel audits clean. The shared-memory split is an exact fit (`sval` 1024 B + `sidx` 1024 B against the launcher's `block * 8`), `blockDim.x` is pinned to 256 so the power-of-two tree reduction is valid, idle threads at `n < blockDim.x` always lose the reduction, the tie-break matches `sample_gumbel` construct-for-construct, and `-inf`/NaN logits behave correctly. The threshold is right too: `global_max + logf(min_p)` is the exact log-space form of the CPU stage's `threshold = min_p * max_probability` over an **un-tempered** softmax (`src/inference.rs:7157-7166`), so it honours llama.cpp's temperature-last chain rather than sharpening the filter. I went looking for a temperature-ordering divergence here and there isn't one.

The lazy KV change is also sound — all three CPU forward paths go through `ensure_cpu_kv_materialized()`, and right after a GPU prefill `filled() == position`, so recovery succeeds and `store_mirrored_layer_kv` advances the watermark. `verify_batch` still passes `flash_ok = false`, so the speculative-verify bit-identity contract is untouched.

## Findings on the new code

**`min_keep` is silently dropped on the newly-enabled GPU min-p lane.** `sample_gumbel_filtered` implements min_p as a bare threshold (`src/cuda_resident.rs:4107`) with no `min_keep` floor, the guard in `generate_next_token_sampled_resident` (`src/inference.rs:3536`) doesn't disqualify on it, and only `config.min_p` is forwarded (`:3566`). The CPU reference applies `.count().max(min_keep).min(candidates.len())` (`:7163`), and `min_keep` is a public request field (`src/api/mod.rs:933`) threaded to this call with no clamp. Worth noting `typical_p` is unguarded here too, but that's pre-existing — this PR is what first makes `min_keep` an *independent* divergence rather than one riding on an already-ignored filter.

**Three doc comments now describe the old eager-mirror behaviour.** The eager mirror is conditional on `CAMELID_CUDA_EAGER_KV_MIRROR` (`src/inference.rs:3320`), which nothing in the tree sets. Stale: `src/inference.rs:3384-3385` — in this diff's own file — whose contrast "unlike the prefill (which mirrors eagerly via `copy_resident_cuda_kv_to_host`)" is now backwards; and `src/inference/metal_resident.rs:433-434` and `:440-441`.

**The relaxed identity check contradicts the doc block above it.** `recover_cpu_kv_from_cuda_resident` went from `filled() != position` to `filled() < position`, but the comment still states the rule as "`filled() == position` EXACTLY", calls it "deliberately stricter than the Metal side's `filled() >= position`", and justifies it as "the same readiness triple `verify_drafts_gpu` already uses" — and those sibling checks (`~3679`, `~3781`) still use `==`. The relaxation looks safe today because streaming decode is posted as an Exclusive run-to-completion job, so two sessions of one model can't interleave on the shared KV cache. But with the eager mirror gone this recovery is now load-bearing rather than a rare fallback, and the only thing preventing cross-sequence contamination is a scheduler invariant enforced in `src/api/mod.rs`, not here. If the relaxation is intentional, the doc block should say so and explain why the sibling checks stay strict.

**The min-p test doesn't assert survivor-set equality.** `sample_gumbel_filtered_matches_min_p_threshold` pins `inv_temp = 1.0` and checks only that the sampled index clears the threshold. It can't compare token-for-token against the CPU sampler by construction (CPU draws inverse-CDF from one per-step uniform, the kernel draws per-index Gumbel-max) — but it could assert the surviving *set* matches, which is where `min_keep` and the boundary comparison would show up.

**`CAMELID_CUDA_PREFILL_BATCH_TOKENS` bypasses the quant-specific clamps.** `batched_prefill_token_cap()` reads the new var first and returns `k.clamp(1, 256)` before the Q1_0 and `has_kquant` branches. Both existing overrides do the opposite — `CAMELID_CUDA_KQUANT_BATCH_TOKENS` clamps to 1..=4 and then runs an explicit shared-memory fit search. `launch_kquant_gemm_batched` enforces that budget with a hard `assert!`, so on a Q4_K model (n_sb 32, aux_lanes 9) any value >= 5 panics the process. It needs `CAMELID_CUDA_RESIDENT_PREFILL_BATCHED=1` too, so it's operator-reachable rather than user-reachable — but both are documented A/B knobs. Moving the env read below the quant branches would fix it.

**The `execution_plan.rs` insert is a behavioural no-op.** Both keys are in `MANAGED_ENV_KEYS`, and `PlannerEnv::apply()` already calls `env::remove_var` for every key a plan doesn't insert — so that branch already left them unset, and `env_flag_enabled` returns false for both "absent" and `"off"`. The existing test can't catch this either (it asserts only `!= Some("on")`). Harmless, but if the description credits it with fixing the CPU-repack / GPU-resident conflict, that credit belongs to the `q8_0_raw_bytes` PackedRows4 conversion instead.

## Description accuracy

The summary says the fused sampler does `top_k` and `top_p` filtering — only min_p is wired; both still disqualify the fast lane at `src/inference.rs:3536`. It also doesn't mention the flash-prefill kernel replacement, the default-on flip, the new prefill chunk sizing, or the single-token prefill enablement, all of which are in the diff.

## Suggested path

Close this branch and reopen just the two novel pieces — the min-p sampler and the lazy KV mirror — on a fresh branch off current `main`. That drops the revert risk entirely and leaves a small, easy-to-gate PR whose new code is already in decent shape. If flipping flash prefill default-on is a deliberate proposal, it deserves its own PR with `scripts/validate-cuda-prefill-row.sh` output attached, since that's the gate the opt-in contract exists to force.

PR #674, opened 2026-08-18
feat(fabric): add and remove a machine without restarting the proxy

karan68:
﻿## What this changes

`--node` is read once at startup, so changing which machines the fabric places on means stopping the
proxy ΓÇö which drops whatever everyone else has in flight, the exact thing the graceful stop in #668
exists to prevent. A fabric of personal machines changes often, because personal machines sleep, move
network and get rebuilt.

`--nodes-file <PATH>` puts the set in a file and re-reads it as it changes:

```
# the machines I place on
desktop=192.0.2.10:8181
laptop=192.0.2.11:8181
```

One `LABEL=HOST[:PORT]` per line ΓÇö the syntax `--node` already takes, parsed by the same
`parse_node_spec` through the same `parse_fabric`, so there is one answer to "what is a node spec" and
duplicate labels are refused here exactly as they are on the command line. Blank lines and whole-line
`#` comments are skipped, so a machine can be taken out for an hour by commenting it rather than
deleting what you wrote. It conflicts with `--node`, because "which machines am I placing on" must
have one answer.

## The load-bearing part is the observation, not the file

An observation describes the set it was taken over. One taken before a machine was added or removed is
not merely *stale*, it is **about something else**: reusing it would place on a node that has gone, or
refuse to place on one that has just arrived, for the whole of the freshness window (500 ms by
default, and unbounded if an operator raises it).

So the node set carries a **generation**, and an observation is reused only while its generation still
matches. Two details that matter:

- The generation is bumped only when the parsed set actually **differs**, not merely when the file was
  re-read. Saving the file without editing it therefore costs nothing.
- `NodeSet::current` returns the generation rather than a "did it change" flag, so a caller that only
  wanted the specs cannot swallow the change and leave the next caller reusing an observation of a set
  that no longer exists.

Lock order is now **nodes ΓåÆ observed ΓåÆ reserved**, and no two are ever held together; the node set's
lock is released before the observation's is taken.

## Where I deliberately did not follow the plan

The plan this came from said that removing a node "has to drop both" the observation and the
reservations. **Only the first is right, and I did not do the second.**

Reservations count what is actually in flight, keyed by label, and a `Placement` releases by label
when it drops. A request already running on a machine you just removed is still counted while it runs
and still releases when it finishes. Dropping those counts would lose track of real work and let the
next placement over-commit a node that is still busy. There is a test for it, and an integration test
that proves the removed machine still answers the request it was already given.

## Two refusals, both deliberate

- An unusable file ΓÇö unreadable, invalid, or **naming no nodes at all** ΓÇö stops the proxy at startup
  rather than starting it on an empty fabric, which would answer every request with 503 while looking
  as though it had started correctly.
- A *re-read* that fails keeps the previous set. A file is usually replaced by writing a new one and
  renaming it over the old, so a read landing mid-swap sees a partial or empty file; emptying the
  fabric on that would turn an ordinary edit into a total outage. The cost is real and is stated where
  an operator will meet it: a change written into a broken file — or deleting the node file altogether
  — has **not** taken effect, and the machine you meant to take out is still taking requests. So the
  proxy prints to stderr when a re-read starts failing, and again when it recovers, rather than relying
  on `RUST_LOG` being set:

  ```
  fabric: could not reload the node file: <why>. The previous set of 2 machines is still
  being placed on, so a change written here has NOT taken effect.
  ```

  Once each way, not once per re-read. The `tracing::warn!` alongside it stays periodic on purpose:
  it describes a state rather than an event, so a log stream should carry it every interval.

## Measurements

Against `upstream/main` at `d080367b4`. No test was removed or renamed.

| target | before | after |
| --- | --- | --- |
| `--lib fabric::` | 170 | 184 |
| `--test fabric_serve` | 60 | 63 |
| `--test fabric_end_to_end` | 14 | 14 |
| `--bin camelid` | 40 | 41 |

Gates on the commit, clean tree: `cargo fmt --all --check` exit 0 ┬╖ `cargo clippy --all-targets --
-D warnings` **0 findings** ┬╖ `check-public-scrub.sh` exit 0 ┬╖ `git diff --check` exit 0 ┬╖ all touched
files LF.

The integration tests run against the **shipped** one-second re-read interval, not a test-only bound,
and poll to a 5 s ceiling.

## Ablations

Each row is one edit to the source, both suites re-run, then reverted; the failures listed are the
complete set. In every row the pre-existing tests all passed, so these are the new tests failing, not
collateral.

| ablation | unit failures | integration failures |
| --- | --- | --- |
| the node-set generation is ignored when reusing an observation | 1: `a_changed_node_set_is_not_answered_from_the_observation_of_the_old_one` | 0 (63 pass) |
| the file is never re-read | 7 | 3 |
| a file naming no nodes is accepted | 2: `an_unusable_file_stops_the_proxy_rather_than_emptying_the_fabric`, `a_broken_or_empty_file_leaves_the_previous_set_in_force` | 0 (63 pass) |

The first row is the interesting one: it fails **only** the unit test, because the integration tests
use the default observation age of zero and so never reuse one. That is the honest reading ΓÇö the
generation check is invisible to a proxy that re-probes every time, and only bites the configuration
`fabric serve` actually ships (500 ms).

## A flaky test I wrote and removed

An ablation failed one test I had not predicted, and it was my fault, not the ablation's. I had
asserted that *reordering* the file bumps the generation ΓÇö but the reordered text is the **same
length** as the original, so a write landing in the same modification-time tick is indistinguishable
from no write at all, and the assertion failed on timing rather than on behaviour. I removed that half
and said why in the test. Adding and removing change the length and are covered properly.

## What this does not do

- No add/remove **route**. The file is the whole interface. A loopback admin route is a reasonable
  follow-up and deliberately not in this PR.
- `--nodes-file` is on `fabric serve` only. `fabric status|route|run` are one-shot commands where
  "without restarting" means nothing; letting them read the same file is a small, separate convenience.
- No cross-machine receipt: the second machine was unavailable. Everything here is loopback and
  in-process on Windows.

`Fabric::specs()` returns an owned snapshot rather than a borrow ΓÇö with a set that can change there is
no slice to lend. It had **zero callers** repo-wide.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -185,6 +185,39 @@ fabric: could not reload client keys: <why>. The previous set of 2 clients is st
 revocation written here has NOT taken effect.
 ```
 
+### Adding and removing a machine while it runs
+
+`--node` is read once at startup, so changing the fabric means restarting the proxy — which drops
+whatever everyone else has in flight. `--nodes-file <PATH>` puts the set in a file instead and
+re-reads it as it changes:
+
+```
+# the machines I place on
+desktop=192.0.2.10:8181
+laptop=192.0.2.11:8181
+```
+
+One `LABEL=HOST[:PORT]` per line, exactly the syntax `--node` takes, parsed by the same code. Blank
+lines and whole-line `#` comments are ignored, so a machine can be taken out for an hour by
+commenting it rather than deleting what you wrote. It conflicts with `--node`, because "which
+machines am I placing on" must have one answer.
+
+Adding a line makes that machine placeable; removing one stops it being placed on. Neither restarts
+the proxy, and **neither disturbs a request already running on the machine you removed** — that
+request is answered by the node it was placed on, and only new placements see the new set.
+
+The file is re-read at most once a second, and only when its size or modification time has changed.
+A change is only acted on if the set actually differs, so saving the file without editing it costs
+nothing.
+
+Two refusals are deliberate. A file that cannot be read, is not valid, or **names no nodes at all**
+is refused at startup rather than started on an empty fabric, which would answer every request with
+503 while looking as though it had started. And if a *re-read* fails the same way, the previous set
+stays in force and a warning is logged — a file is usually replaced by writing a new one and renaming
+it over the old, so a read landing mid-swap sees a partial or empty file, and emptying the fabric on
+that would turn an ordinary edit into a total outage. The cost is that a change written into a broken
+file has not taken effect, and the warning is the only thing that says so.
+
 The proxy re-probes its nodes at most once per `--observation-max-age-ms` (500 by default) rather
 than once per request. Inside that window its view can be wrong, so a request placed on a node that
 has gone since is placed again on another node serving the same model, up to `--max-forward-attempts`
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -213,10 +213,24 @@ nothing.
 Two refusals are deliberate. A file that cannot be read, is not valid, or **names no nodes at all**
 is refused at startup rather than started on an empty fabric, which would answer every request with
 503 while looking as though it had started. And if a *re-read* fails the same way, the previous set
-stays in force and a warning is logged — a file is usually replaced by writing a new one and renaming
-it over the old, so a read landing mid-swap sees a partial or empty file, and emptying the fabric on
-that would turn an ordinary edit into a total outage. The cost is that a change written into a broken
-file has not taken effect, and the warning is the only thing that says so.
+stays in force — a file is usually replaced by writing a new one and renaming it over the old, so a
+read landing mid-swap sees a partial or empty file, and emptying the fabric on that would turn an
+ordinary edit into a total outage.
+
+That fallback has a cost worth stating plainly: **a change written into a file that does not parse,
+or deleting the node file altogether, does not change anything.** The previously loaded set goes on
+being placed on until the file is readable again or the proxy is restarted — and a restart refuses to
+come up at all while the file is unusable. A machine you meant to take out is still taking requests.
+So the proxy prints to stderr when a re-read starts failing, and again when it recovers, rather than
+relying on `RUST_LOG` being set:
+
+```
+fabric: could not reload the node file: <why>. The previous set of 2 machines is still being
+placed on, so a change written here has NOT taken effect.
+```
+
+It is printed once each way, not once per re-read, so a file left broken does not emit a line a
+second for as long as it stays that way.
 
 The proxy re-probes its nodes at most once per `--observation-max-age-ms` (500 by default) rather
 than once per request. Inside that window its view can be wrong, so a request placed on a node that
--- a/src/fabric/mod.rs
+++ b/src/fabric/mod.rs
@@ -25,6 +25,7 @@ pub(crate) mod client_keys;
 pub mod forward;
 pub(crate) mod http;
 pub mod node;
+pub(crate) mod nodes;
 pub mod policy;
 pub mod probe;
 pub mod server;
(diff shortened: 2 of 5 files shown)


karan68:
Second commit, `22d09162a`, after reviewing my own first one against the built binary rather than the tests.

## The fallback was silent, and silence was the whole risk

Keeping the previous set in force when a re-read fails is right, and I did not change it — emptying the fabric because a file was caught mid-rewrite refuses *every* request, which is worse than acting a second late. But it rests on the operator being told, and this PR said so itself: *"the warning is the only thing that says so."*

That warning was `tracing::warn!`, and `RUST_LOG` is unset on a stock proxy. Measured on the built binary, using `/v1/health`'s `node_detail` to name the live set — same situation twice, only an environment variable different:

| | |
| --- | --- |
| `RUST_LOG` unset | start `a,b` → broken file `a,b` → **delete the node file entirely** `a,b`. **stdout: nothing about it. stderr: nothing at all.** |
| `RUST_LOG=camelid=warn` | identical, and the WARN appears — **once per re-read, so once a second** for as long as the file stays broken |

So a machine an operator meant to take out goes on being placed on, and on a default proxy nothing anywhere says so. Silent when you need it, and a line a second when you don't.

A failed re-read now prints to stderr as well, once on the way into the degraded state and once on the way out, from a pure `stale_node_set_notice` that names the consequence rather than only the error:

```
fabric: could not reload the node file: <why>. The previous set of 2 machines is still
being placed on, so a change written here has NOT taken effect.
fabric: node file <path> is readable again; placing on 1 machine
```

Re-measured after, no `RUST_LOG`: a broken file *and then* a deleted file, across eight seconds of one-second re-reads, produced **exactly one line**; recovery produced exactly one more.

**The periodic `tracing::warn!` is deliberately left as it is.** It describes a state, not an event, so a log stream should carry it every interval — someone who starts tailing after the transition would otherwise see nothing at all. The stderr notice is for the person at the terminal, and that one has to be edge-triggered or it fills the screen. Different consumers, different cadence.

## One more claim that was not quite true

`FileStamp`'s doc said length plus modification time covers a rewrite inside one timestamp tick. It narrows that window; it does not close it — as this PR's own removed reorder test discovered. Closing it would mean reading the file every interval, which is the cost the stamp exists to avoid, so the doc now says what it actually buys.

`docs/CONFIGURATION.md` gains the deletion case explicitly, because "removing a line stops that machine being placed on" invites "removing the file stops everything", and the opposite is true.

- **The generation design is right, and the check is in the right place** — before the freshness clock, not after it, so an observation of a set that has since changed is never reused even while it is still "fresh".
- **The generation is bumped only when the parsed set actually differs**, so touching the file costs nothing, and a *failed* re-read correctly does not bump it — the set did not change, so the observation is still about the right set.
- **Not dropping reservations on removal is correct**, and better than the plan it came from. They are counted by label and released by the `Placement` whatever the set says; dropping them would lose track of work still running.
- **Lock order holds.** `nodes.current()` releases before `observed` is taken; `reserved` is separate. No two are held together.
- **One transient I traced and satisfied myself is not a bug**: a caller holding generation *N* can write back an observation tagged *N* after a concurrent caller already stored *N+1*. It costs one extra probe round and is rejected by the very next caller, because each reads the generation fresh. Self-correcting, bounded to the instant of an edit.
- `Fabric::specs()` and `is_empty()` now touch the filesystem, which is unusual for things shaped like getters, but neither has a non-test caller and the doc says "as they stand right now".

## Two things for whoever merges

- **This conflicts with #672.** `git merge-tree --write-tree 70f3a8e8e 6716bf6da` reports conflicts in `docs/CONFIGURATION.md` and `src/main.rs` — both PRs add a flag to the same `FabricAction::Serve` block and a section to the same doc. `src/fabric/mod.rs` and `tests/fabric_serve.rs` are clean. Both were cut from `d080367b4` and neither carries the other, so: merge one, rebase the other, re-run the gates. Neither is stale yet.
- **After both land, `client_keys.rs` and `nodes.rs` are two near-identical file-watchers** — same `FileStamp`, same interval, same keep-previous-on-failure, same stale flag, same notice shape. They even disagree on poison policy: `nodes.rs` recovers with `PoisonError::into_inner`, `client_keys.rs` uses `.expect()`. Worth one shared mechanism rather than a third copy next time.

## Gates

Re-measured on the commit with a clean tree: `--lib fabric::` **184** (183 + the notice test), `--test fabric_serve` 63, `--test fabric_end_to_end` 14, `--bin camelid` 41, `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` **0**, `check-public-scrub.sh` 0, `git diff --check` clean, touched files `i/lf w/lf`.

PR #675, opened 2026-08-18
feat(fabric): stop a request when the client that asked for it has gone

karan68:
## The defect

`src/fabric/server.rs` documented it plainly on `main`:

> A non-streaming dispatch runs on a blocking thread and blocking socket I/O is not cancellable, so a
> client that hangs up leaves its dispatch running until the node answers or `forward_timeout`
> expires.

That is worse than a wasted thread. `src/api/engine.rs:1` describes the engine as *"one dedicated OS
thread ... executes at most one compute step at a time"*, so an abandoned request holds the **node's
only generation slot**, not this proxy's. On a two-machine fabric that is half the operator's
hardware, spent on an answer nobody will read, for up to `--forward-timeout-s` (300s by default).

A blocking thread cannot be aborted from outside — dropping its `JoinHandle` detaches it rather than
stopping it — so the only fix is to tell the work it is no longer wanted, and have it agree to look.

## What this does

Every handler frame holds a `Cancel`, fired when that frame is dropped, which is what a client
leaving looks like to axum. Every blocking loop that can wait on a node consults it beside its own
deadline, and the socket to the node is dropped on the way out. A node that sees its caller go is
already the case the engine handles: `src/api/engine.rs:14` says *"a dropped handler (client
disconnect) stops the running job within one step"*.

Both directions of the gap are covered:

| | before | after |
|---|---|---|
| buffered dispatch | nothing to relay, so nothing whose failure could reveal the client had gone | stops at the next socket-operation boundary |
| streaming, before the first event | the relay's own mechanism cannot fire — nothing has been sent yet, and a head is a whole prefill away | same |
| streaming, mid-relay | already stopped, via a failing send to the client | unchanged, plus the idle gap between two events is now covered |
| probe round | bounded by `--timeout-ms` | **unchanged, deliberately** — see Honest gaps |

## Design calls worth arguing with

**A cancellation is not a node failure, and that distinction is behavioural.**
`ForwardError::Cancelled` is its own variant rather than a `Transport`. `forget_observation_if_node_vanished`
drops the observation on `Transport`, which is what keeps #656's freshness bound honest — but folding
cancellation into it would mean *every client that hangs up makes the next request pay a whole probe
round*, on a fabric that is working perfectly. It also must not be `node_never_received_it()`: nobody
is left to hand a second node's answer to. Ablation A2 exists precisely to prove both.

**`Cancel` is an atomic flag, not a `tokio_util::CancellationToken`.** Nothing here ever awaits it —
the async side is the side that *fires* it — so importing an async notification primitive to answer a
synchronous question would be the wrong kind of reuse. `Cancel::never()` exists so the callers with
no client to lose (`fabric status|route|run`, and a health probe) say so at the call site rather than
pass an anonymous unfired one.

**The streaming guard is moved into the response body, not created inside it.** A body can be dropped
before it is ever polled — a client that leaves between the head and the first read — and a guard
declared inside the generator would never exist to fire. That distinction is invisible in the source,
so `a_response_body_dropped_before_it_is_read_still_gives_up_on_the_request` pins it.

**Cancellation is checked *before* the deadline in every loop**, so an abandoned request is reported
as abandoned rather than as a node that was too slow. The answer is 499, which nothing on the network
reads — the client it describes has gone — but which keeps a log or a test from being told a healthy
node failed.

## What I measured before building it

The whole design rests on axum dropping the handler future when a client half-closes while a response
is pending. I did not assume it. A throwaway probe (an axum handler holding a drop guard, a raw
socket that writes a complete request and then closes) reported the guard firing **15.9 ms** after the
close. The probe is not part of this PR; what is part of it is
`a_client_hanging_up_gives_the_node_back_its_generation_slot`, which proves the same thing end to end
over real sockets against a node that would otherwise have held the request for 30s, with the proxy's
forward budget set to 60s so that neither of those can be what ended it.

## Gates

Rebased onto `a825682d5` (main, after #672 and #674). Every leg re-run on the rebased commit
`17de63f5f`, with the tree verified clean at the end of the run, so the gates describe the commit
that is being merged rather than a working tree.

```
cargo fmt --all -- --check                     0
cargo clippy --all-targets -- -D warnings      0 findings
cargo test --lib fabric::                    206   (195 + 11)
cargo test --test fabric_serve                68   (65 + 3)
cargo test --test fabric_end_to_end           15   (14 + 1)
cargo test --bin camelid                      42   (unchanged)
scripts/check-public-scrub.sh                  0
git diff --check                               0
git ls-files --eol                             all i/lf w/lf
```

The four counts were predicted before the rebase was run, by adding this PR's contribution to the
counts main reached after #674, and each landed exactly: 206 / 68 / 15 / 42. The rebase resolved one
conflict, `src/fabric/mod.rs`'s module list, by keeping both neighbours' entries; this PR's own diff
is byte-identical across the rebase (10 files, +957/-43 before and after).

## Ablations

Each one breaks a single mechanism, predicts exactly which tests fail, and is restored
byte-identically (verified by SHA-256, not by `git status` — the new file is untracked during the
run).

| ablation | predicted | measured |
|---|---|---|
| **A1** nothing ever reads the cancellation | 9 unit, 1 e2e, 2 integration | **exactly that**, and the pre-existing streaming hang-up test still passed |
| **A2** a cancellation is reported as a `Transport` failure | exactly 1 unit + 1 e2e | **exactly that** — `fabric_serve` stayed 63/0, so the hang-up still reaches the node; what breaks is only the attribution and the observation |
| **A3** the buffered handler frame holds no guard | exactly 1 integration | **exactly that** — `a_client_hanging_up_gives_the_node_back_its_generation_slot`, while the streaming one kept passing |

Control on the restored tree: 181 / 15 / 63, all green.

The ablations above were measured on the pre-rebase base `d080367b4`; they exercise the cancellation
mechanism, which the rebase did not touch.

There is also a negative control in the suite itself,
`a_client_that_waits_is_served_and_the_node_is_never_told_to_stop`: a proxy that cancelled everything
the moment it started would pass the two hang-up tests and fail this one.

## A stale claim this PR had to fix

`a_client_hanging_up_stops_the_node_generating` carried the sentence *"The buffered path cannot do
this — blocking socket I/O is not cancellable — so this is a property the streaming path adds rather
than inherits."* That is now false, and the comment says what is actually true instead: the relay
notices because a send fails, which is why the buffered path needed a mechanism of its own.

## Honest gaps

* **Live receipt, on real binaries.** A real `camelid serve` node (Llama-3.2-1B-Instruct-Q8_0) behind
  `fabric serve --forward-timeout-s 120`, twice, with the node's own `/v1/health`
  `engine_queue_depth` as the instrument — that gauge is the generation slot this is about:

  | arm | took the job | slot free again | node busy | curl |
  | --- | --- | --- | --- | --- |
  | client waits | 200 ms | 4466 ms | **4266 ms** | 200 |
  | client hangs up at 2 s | 227 ms | 2273 ms | **2046 ms** | 000 |

  The slot came back **~46 ms after the client's socket closed**, not at 4.3 s when the answer would
  have been ready and nowhere near the 120 s budget, so neither of those is what ended it.
* **A probe round already in flight is not interrupted.** It is bounded by `--timeout-ms` (2s by
  default) and costs a node a health read rather than its generation slot, so threading cancellation
  through `probe.rs` would be noise for no gain. `probe.rs` passes `Cancel::never()` and says why.
* **A client that hangs up still produces no access-log line.** The middleware's own future is
  dropped with the handler's. That is pre-existing behaviour and unchanged here; it is worth knowing
  before reading `RUST_LOG=camelid=info` as a complete record.
* **The stop path is unchanged.** A graceful shutdown still drains in-flight work bounded by
  `forward_timeout`; this is about a client leaving, not about the proxy being asked to stop.
* Per the campaign's own standing rule, this is a third open fabric PR while #672 and #674 are still
  unmerged. Measured rather than guessed: #672 conflicts with #674 (`docs/CONFIGURATION.md`,
  `src/main.rs`) and with this one (`src/fabric/mod.rs`), while #674 and #675 are clean against each
  other — so #672 is the one to merge first or rebase last.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -242,6 +242,15 @@ part-way through a stream, ends the request with 502 rather than risking a secon
 `x-camelid-fabric-node`, `x-camelid-fabric-reason` and `x-camelid-fabric-attempts`, so a client can
 tell a first-choice placement from a failover.
 
+A client that hangs up takes its request's work with it. The proxy notices the connection going and
+hangs up on the node in turn, which that node reads as its own client leaving — it stops generating
+within one decode step. So a request nobody is waiting for gives back the node's generation slot
+instead of holding it for the rest of `--forward-timeout-s`, which matters because a node runs one
+generation at a time. It is not instantaneous: the check happens between socket operations, so it
+takes effect within about 100 ms while the proxy is waiting on a node. A probe round already under
+way is not interrupted, but it is bounded by `--timeout-ms` and costs a node a health read rather
+than its generation slot.
+
 Request bodies are bounded at the same 16 MiB default the node itself uses. A streaming request is
 relayed as it arrives; `fabric run`, which returns one complete answer, refuses `stream: true` with
 400 instead.
--- a/src/fabric/cancel.rs
+++ b/src/fabric/cancel.rs
@@ -1,0 +1,101 @@
+//! Whether the work a request asked for is still wanted.
+//!
+//! Placement and forwarding are synchronous socket I/O, and the proxy runs them
+//! on a blocking thread. A blocking thread cannot be aborted from outside —
+//! dropping its `JoinHandle` detaches it rather than stopping it — so the only
+//! way to end that work early is to tell it, and have it agree to look.
+//!
+//! That matters more here than the shape of the code suggests. A Camelid node
+//! executes at most one decode step at a time, so a request nobody is waiting
+//! for does not merely waste a thread in this process: it holds the *node's*
+//! generation slot until the node finishes or `--forward-timeout-s` expires,
+//! and on a two-machine fabric that is half the operator's hardware, spent on
+//! an answer that will be thrown away.
+//!
+//! # What is promised
+//!
+//! A cancelled request stops at the next socket-operation boundary, not
+//! instantly. In practice that is within one read timeout (100 ms) while
+//! waiting for a node to answer, and within one dial attempt while connecting.
+//! Nothing here interrupts a syscall already in progress.
+//!
+//! Cancelling is not a failure of the node it was placed on, and the rest of
+//! the fabric treats it accordingly: the observation the request was placed
+//! from is kept, and the request is never sent to a second node, because there
+//! is nobody left to send an answer to.
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+/// One request's cancellation, shared between whoever waits for the answer and
+/// whoever is producing it.
+///
+/// Cloning shares the same signal rather than copying its state, so a caller
+/// hands a clone to the work and keeps one to fire.
+#[derive(Clone, Debug)]
+pub struct Cancel(Arc<AtomicBool>);
+
+impl Cancel {
+    /// A cancellation for one request, not yet fired.
+    pub fn new() -> Self {
+        Self(Arc::new(AtomicBool::new(false)))
+    }
+
+    /// A cancellation nothing will ever fire.
+    ///
+    /// For callers with no client to lose: `fabric status|route|run` send one
+    /// request per process and are ended by ending the process, and a health
+    /// probe is already bounded by its own timeout. Saying so at those call
+    /// sites keeps it a decision rather than an omission.
+    pub fn never() -> Self {
+        Self::new()
+    }
+
+    /// Stop wanting the answer. Idempotent, and safe to call from any thread.
+    pub fn cancel(&self) {
+        self.0.store(true, Ordering::SeqCst);
+    }
+
+    /// Whether the answer has been given up on.
+    pub fn is_cancelled(&self) -> bool {
+        self.0.load(Ordering::SeqCst)
+    }
+}
+
+impl Default for Cancel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn a_fresh_cancellation_has_not_been_fired() {
+        assert!(!Cancel::new().is_cancelled());
+        assert!(!Cancel::never().is_cancelled());
+    }
+
+    #[test]
+    fn a_clone_carries_the_same_signal_rather_than_a_copy_of_it() {
+        let held_by_the_caller = Cancel::new();
+        let handed_to_the_work = held_by_the_caller.clone();
+
+        held_by_the_caller.cancel();
+
+        assert!(
+            handed_to_the_work.is_cancelled(),
+            "the work was not told, so it would run to its own deadline"
+        );
+    }
+
+    #[test]
+    fn firing_twice_is_the_same_as_firing_once() {
+        let cancel = Cancel::new();
+        cancel.cancel();
+        cancel.cancel();
+        assert!(cancel.is_cancelled());
+    }
+}
--- a/src/fabric/cancel.rs
+++ b/src/fabric/cancel.rs
@@ -19,6 +19,13 @@
 //! waiting for a node to answer, and within one dial attempt while connecting.
 //! Nothing here interrupts a syscall already in progress.
 //!
+//! One window is deliberately not covered: a *write* to the node uses the
+//! whole forward budget as its timeout, so a peer that accepts the connection
+//! and then stops reading can hold a send for far longer than either bound
+//! above. It takes a pathological node to reach — request bodies are capped at
+//! 16 MiB and a node reads its request promptly — and shortening it would mean
+//! chunking every write to poll a flag between the pieces.
+//!
 //! Cancelling is not a failure of the node it was placed on, and the rest of
 //! the fabric treats it accordingly: the observation the request was placed
 //! from is kept, and the request is never sent to a second node, because there
--- a/src/fabric/forward.rs
+++ b/src/fabric/forward.rs
@@ -17,6 +17,7 @@ use serde_json::{json, Value};
 
 use super::http::{self, HttpError};
 use super::node::NodeSpec;
+use super::Cancel;
 
 /// Generation can legitimately take minutes, so this is far longer than a probe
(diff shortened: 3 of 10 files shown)


karan68:
Second commit, `b2375677c`, plus the receipt this PR said it did not have.

## The live receipt

The first Honest gap was *"No live receipt on built binaries ... the claim that the node stops generating is read from the engine's own source and doc."* That is the load-bearing claim of the whole PR, so I took it.

A real `camelid serve` node (Llama-3.2-1B-Instruct-Q8_0) behind `fabric serve --forward-timeout-s 120`, same binary, same prompt, same node, twice — the only difference is whether the client stayed. The instrument is the **node's own** `/v1/health` `engine_queue_depth`, polled every 150 ms: that gauge *is* the generation slot this PR claims to give back.

| arm | node took the job | slot free again | node busy for | curl |
| --- | --- | --- | --- | --- |
| **A** client waits for the answer | 200 ms | 4466 ms | **4266 ms** | 200 |
| **B** client hangs up at 2 s | 227 ms | 2273 ms | **2046 ms** | 000 |

The node's generation slot came back **~46 ms after the client's socket closed** — not at 4.3 s when the answer would have been ready, and nowhere near the 120 s forward budget, so neither of those can be what ended it. That is consistent with the ~16 ms axum drop measured at design time plus one 100 ms read poll plus a decode step.

Arm A is the control that makes arm B mean something: same everything, and the node holds its slot for the full generation when somebody is still waiting.

## Three claims that were not true

**The Limits list still carried the bullet this branch exists to delete.**

> A non-streaming dispatch runs on a blocking thread and blocking socket I/O is not cancellable, so a client that hangs up leaves its dispatch running until the node answers or `forward_timeout` expires.

That is the text the PR body quotes as *The defect*, and it was sitting one line below the new bullet saying the opposite. This repo puts its contracts in module docs, so a reader got the false half. Removed, with the part that stays true folded into the new bullet: mid-relay a stream already noticed on its own, so what cancellation adds *there* is the idle gap between two events, and everywhere else it is the only mechanism. Exactly the same class of stale claim this PR already fixed in a test comment — it just missed the one in its own docs.

**The 499's rationale said it protects "the reader of a log or a test".** There is no log line — a client that hangs up takes the access-log middleware's future with the handler's, which this PR states in its own Honest gaps. Only a test can read it, and now the doc says only that.

**`cancel.rs` named two bounds and there are three.** It promises one read timeout (100 ms) while waiting and one dial attempt while connecting. But `set_write_timeout` is the *whole* forward budget, so a peer that accepts a connection and then stops reading can hold a send far longer than either. It takes a pathological node — bodies are capped at 16 MiB and a node reads its request promptly — and shortening it would mean chunking every write to poll a flag between the pieces. Stated rather than fixed.

This is good code and the review should say so.

- **The 100 ms read timeout the design rests on is pre-existing on `main`** — verified with `git show upstream/main:src/fabric/http.rs`, not assumed. So the poll granularity is free, and "within one read timeout" is exact rather than aspirational.
- **Cancellation is checked before the deadline in every loop**, so an abandoned request is never attributed to a slow node.
- **`ForwardError::Cancelled` is correctly isolated**: `peer_never_received_it()` is false so it is never re-placed; it is not in the `forget_observation_if_node_vanished` match so the observation survives; `after_the_head` keeps it out of `Transport` once a head is in.
- **The engine claim is real and quoted accurately** — `src/api/engine.rs:14-17` does say a dropped handler stops the running job within one decode step, and `:442` explains why the job otherwise runs to completion.
- **The streaming guard being *moved into* the body rather than created inside it** is the subtle part, and `a_response_body_dropped_before_it_is_read_still_gives_up_on_the_request` pins exactly the distinction that is invisible in the source.
- `CancelOnDrop` is bound to a named local, not `_` — which would have dropped it immediately and cancelled every request before it started. The comment says so, which is the right place to say it.

## Gates

Docs only, so nothing moved: `--lib fabric::` 181, `--test fabric_serve` 63, `--test fabric_end_to_end` 15, `--bin camelid` 40, `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` **0**, `check-public-scrub.sh` 0, `git diff --check` clean, touched files `i/lf w/lf`.

## Still true from the original Honest gaps

A probe round already in flight is not interrupted; a client that hangs up produces no access-log line; the stop path is unchanged.

## Merge order, measured rather than assumed

This is a third open fabric PR, so I ran the pairs (`git merge-tree --write-tree`):

| pair | result |
| --- | --- |
| #672 × #674 | **conflict** — `docs/CONFIGURATION.md`, `src/main.rs` |
| #672 × #675 | **conflict** — `src/fabric/mod.rs` (both add a line to the module list, adjacent) |
| #674 × #675 | clean |

So #672 conflicts with both of the others and is the one to merge first or rebase last. Every conflict is a single mechanical hunk; none of the three is stale against `main` yet.

karan68:
Rebased onto `a825682d5` (main, after #672 and #674) and re-gated. Force-push was
`b2375677c` -> `17de63f5f`.

The rebase hit one conflict, `src/fabric/mod.rs`'s module list, where this branch adds
`pub mod cancel;` and main had gained `client_keys` and `nodes`. Resolved by keeping all of
them, so the list is main's eight modules plus `cancel`.

Two things I checked rather than assumed:

* **This PR's own diff survived the rebase byte-identically.** `git diff <base>...<head>` is
  10 files, +957/-43 both before and after, so the rebase absorbed the two merged PRs without
  quietly changing what this one does.
* **The re-gated counts were predicted first, then measured.** Adding this PR's contribution to
  the counts main reached after #674 predicts 206 / 68 / 15 / 42; the run produced exactly
  206 / 68 / 15 / 42, with `clippy --all-targets -- -D warnings` at 0 and the tree verified
  clean at the end of the run, so the gates describe `17de63f5f` and not a working tree.

Also fixed 16 em-dashes in the description above that an earlier `--body-file` round-trip had
double-encoded into `Î“Ã‡Ã¶`.

PR #681, opened 2026-08-20
feat(api): add authenticated LAN browser chat

karan68:
## Summary

- add an authenticated `--lan-chat-only` API surface for using Camelid Chat from a mobile browser on a trusted LAN
- enforce API-key authentication before an exact method/path capability allowlist, with startup validation for remote listeners
- let authenticated mobile clients switch only among direct regular `.gguf` files under `--models-dir`
- add `camelid lan-key` for private 256-bit key provisioning and explicit rotation
- adapt the WebUI for auth bootstrap, restricted mobile navigation, authenticated requests, and safe model selection

## Security boundary

- API authentication is enforced by Axum middleware; hiding UI routes is not the authorization boundary
- LAN Chat mode requires a key even on loopback and rejects non-allowlisted routes after authentication
- model loading accepts a validated filename, rejects traversal/absolute paths/symlinks/non-GGUF files, and validates before unloading the current model
- telemetry EventSource is disabled for this restricted surface because browser EventSource cannot attach the API-key header

## Validation

- `cargo test --lib lan_key::tests -- --test-threads=1`
- `cargo test --lib api::server::tests -- --test-threads=1`
- `cargo clippy --all-targets -- --deny warnings`
- `npm run build`
- `npm run smoke:lan-auth` (`LAN_AUTH_SMOKE_PASS`, 33 protected requests, 390px viewport, no horizontal overflow)
- physical Android 16 Chrome flow: real 401 credential gate, authenticated Chat, restricted drawer, 1B to 3B Q4 model switch, and streamed exact response `CAMELID_MOBILE_3B_OK`

## Physical evidence

The complete redacted sequence and SHA-256 manifest are in [`frontend/design-evidence/lan-chat-e2e`](https://github.com/karan68/Camelid/tree/feat/lan-chat-sync/frontend/design-evidence/lan-chat-e2e).

The phone used an ADB USB reverse to the Windows loopback listener because the available guest Wi-Fi had client isolation. This validates the physical mobile browser and authenticated HTTP flow without claiming peer-to-peer guest Wi-Fi connectivity.

A server-terminal screenshot is intentionally omitted: automated Windows console capture was not reliable enough to qualify as evidence. Readiness is instead demonstrated by the authenticated browser flow, successful model switch, completed streamed inference, and visible receipt.

No API key, device serial, private LAN address, or user-specific filesystem path is included in the evidence bundle.


karan68:
Stacked responsive-layout child: https://github.com/karan68/Camelid/pull/6 (base `feat/lan-chat-sync`, one child commit `9aaa86cc`). It keeps the review diff limited to the mobile/orientation work while this parent remains open. After #681 lands, the child can be retargeted to upstream `main`.

karan68:
Correction to the stacked-child pointer above: the final child head is `c9074267f22c5b2bad35ef84f7f0458c0543b8cd` after replacing the scripted visual captures with notice-free, internally coherent evidence. The stack base remains `32c182005363bfee63d2838a244547b186b6201b`, and the child remains one commit / nine files: https://github.com/karan68/Camelid/pull/6

karan68:
Responsive child PR location corrected: upstream https://github.com/timtoole02/Camelid/pull/682 replaces the closed fork-local PR. Final child head remains `c9074267f22c5b2bad35ef84f7f0458c0543b8cd`. Because GitHub cannot base an upstream PR on an unmerged fork branch, #682 temporarily targets `main` and includes this parent stack; merge #681 first, after which #682 reduces to the single responsive commit.

timtoole02 (commented):
Read the full diff, checked the security-critical Rust paths against the surrounding code, verified the evidence bundle, and reproduced the CI failure locally.

## The design holds up

The allowlist is default-deny on exact `(method, path)` and runs *after* `auth.accepts`, so an unknown or newly added route stays refused until it is named — the `/api/not-yet-invented` case pins that. `resolve_lan_chat_model_path` is genuinely tight: `valid_local_model_filename` rejects `/`, `\`, `:`, multi-component paths and non-`.gguf`, then `symlink_metadata` rejects symlinks and non-files before any unload. The key file is 0600, created via `create_new` + hard-link so concurrent first use converges on one key, and rotation is explicit. The negative tests cover the cases that matter.

`84aa66e1` closing the pre-auth `executable`/`listen_addr` disclosure was the sharpest hole here, and the reasoning in that commit message is right: a loopback bind stops meaning "the reader is at the keyboard" the moment this flag exists.

## Blocking — `validation-scripts` is red

`84aa66e1` changed `src/api/mod.rs` again but did not re-update the SmolLM3 pins that `7f05d5cc` had just bumped.

- pinned `RENDERER_GIT_BLOB_SHA1` / fixture `source_git_blob_sha1`: `306571bc1e721de8a9d01e388bbb181038244f20`
- actual `git hash-object src/api/mod.rs` at head: `31782fff3a7be3d58b6931da7e70d6c006221464`

`inspectGroundings` shells out to `git hash-object` and throws `chat_parity_grounding_invalid`. Reproduced in a clean worktree at `84aa66e1`. Updating the blob sha in `qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json`, `scripts/hf-qualification-smollm3-chat-parity.mjs` and `scripts/test-hf-qualification-smollm3-chat-parity.mjs`, then recomputing the fixture digest to `db7e4bd479eb0ebd40d5db61abd80aa4345214a1957a5dead406ff2db7ac2d45`, makes it pass locally.

Worth doing as the very last step before merge — any rebase that moves `mod.rs` invalidates the pin again.

## Should fix

**The evidence manifest does not verify off Windows.** `shasum -a 256 -c SHA256SUMS` fails for `01-terminal-lan-key-redacted.txt` and `README.md` on a fresh checkout. Those two hashes match the files with CRLF endings — computed against the Windows working tree, but the blobs are committed LF and nothing in `.gitattributes` pins them. The other seven (all PNGs) verify fine. Recomputing from the committed bytes fixes it.

Related: `frontend/design-evidence/` is not covered by CI at all — `audit-evidence-bundle-privacy.mjs` and `check-evidence-bundle-checksums.sh` both default to `qa/evidence-bundles`. So the bundle's privacy and checksum claims are self-attested. Running both with an explicit root would make them real. (I did read the screenshots: no key, no LAN address, no user path — the redaction claim is accurate.)

**Telemetry gating regresses full mode.** `App.jsx` now passes `runtime?.status === 'online' && !isLanChatOnly(apiSurface) ? apiBase : null`, and `ensureInferenceTelemetryConnected(null)` calls `sharedStore.disconnect()`. Only the surface check is needed here.

Gating on `runtime.status` means the stream no longer connects until the first successful health poll — which is exactly the "Phase 6.1 DEFECT 1" the comment three lines above says it fixed — and any health blip now tears down the EventSource mid-run. Worse, `disconnect()` closes the source without touching `connection` or calling `notify()`, so the Observatory keeps reporting `live` against a closed stream. This PR is the first caller of `disconnect()` from the app shell, so that latent wart becomes visible here.

**Absolute host paths still reach the phone.** `/api/models/local` returns `models_dir` as an absolute path and `/api/models/current` returns the resident model's absolute `path`; both are allowlisted. That is the same class of disclosure `84aa66e1` just removed from `/health`, and the same thing the evidence README scrubs to `%LOCALAPPDATA%`. It is behind auth here, so not the pre-auth leak — but the frontend now sends `filename` and no longer needs `models_dir` on this surface, so redacting it would make the boundary consistent with its own stated rule.

## Minor

`FirstRunCard` is not gated on `apiSurface`. It calls `/api/models/catalog`, `/catalog/ack`, `/catalog/cancel` and `/catalog/install`, none of them allowlisted. Reachable when `--lan-chat-only` starts without `--model` against an empty models directory — nothing enforces `--model` — which hands the phone a download flow where every action 403s.

`frontend/src/lib/apiSurface.js` has no trailing newline. Cosmetic, no ESLint in this repo.

`CONFIGURATION.md`: "It conflicts with the anonymous remote override" holds for the clap path but not `ServerPolicyArgs::from_env`, which sets both from env with no conflict check. Harmless in practice, since `ServerPolicy::resolve` demands a key before it looks at the override.

karan68:
@timtoole02 thanks - that was a genuinely useful read, and the `84aa66e1` catch was mine to make and I missed it. All of it is addressed in `0fc8f0e2`.

**The pin.** You are right, and my rule was wrong rather than my execution: I had "re-pin when the feature touches `api/mod.rs`" in my head instead of "re-pin on any commit that changes a byte of it". Note the values are not the ones in your comment, because this commit touches `mod.rs` again for the `models_dir` change below - so I recomputed rather than copying: blob `01d52b40b873ab59488ed0eb56e146f87b07520a`, fixture digest `52ee0c2613bf222fe55af35b504915b47a12dfc3303a985c1988f04a07a1bb8c`, canonical length still 5,578 (which is the check that the edit replaced 40 hex characters and nothing else). All 42 `scripts/test-*.mjs` pass, `git hash-object src/api/mod.rs` equals the fixture, and `git grep` for both old values is empty. Flagging the obvious: if this gets rebased over anything that moves `mod.rs`, it needs doing again.

**The manifest.** Reproduced exactly: the seven PNGs match, `01-terminal-lan-key-redacted.txt` is `096b059f` recorded against `b297bb44` actual, and `README.md` is `5661596b` against `558608e5`. Worktree bytes equal the committed blobs, so I recomputed from those - two lines changed. I also wired both gates into CI over `frontend/design-evidence`, since they take an explicit root already and it is two lines; `check-evidence-bundle-checksums.sh` exits 0 and the privacy audit reports `finding_count: 0` in strict mode. Agreed that self-attested claims are not claims.

**Telemetry - your fix as written does not hold, and I would rather say so than quietly ship it.** Dropping the `runtime.status` term makes the effect run on the very first render, where `apiSurface` is still the `full` default because `runtime` is null. A LAN browser then opens `/api/telemetry/stream` with no credential and EventSource retries the 401 forever. `smoke:lan-auth` fails on exactly that assertion - I made the change you asked for, watched it go red, and went back. The surface genuinely is unknown until health answers once.

Both of your actual complaints are real though, so I kept those and dropped only the unsafe part. The connect now **latches**: it waits for one health answer, and after that a blip cannot tear down a live run. And `disconnect()` now sets `UNAVAILABLE` and calls `notify()`, so the Observatory can no longer report `live` against a closed stream - you were right that this PR is what makes that latent wart reachable.

**`models_dir` is gone from this surface**, using the same absent-not-empty shape as `executable`/`listen_addr`. Checked before changing it: load paths are built as `models/<filename>`, `modelActivation.js` documents that they are deliberately not built from `models_dir`, and the only renderer is `DownloadedModelsView`, which this surface hides. The two scripts that do consume it only ever talk to a full-mode server, where it is unchanged.

**`/api/models/current` keeps its path, deliberately.** I measured it against a real 1B load rather than reasoning about it, and the answer surprised me:

| | full | `--lan-chat-only` |
|---|---|---|
| `local.models_dir` | `C:\camelid-fork\models` | absent |
| `current.id` | `Llama 3.2 1B Instruct` | `Llama 3.2 1B Instruct` |
| `health.active_model_id` | `Llama 3.2 1B Instruct` | `Llama 3.2 1B Instruct` |

A model loaded at startup reports its **display name**, not its filename - the filename identity only holds for models loaded through this surface. That path is what `activeFilename` in `useDashboardData` uses to match the resident model to its local record, and it has no fallback, so redacting it would leave the phone unable to tell which model is loaded. That is the model selector this mode exists for. Happy to do it as a follow-up if the resident id is made filename-shaped first, but not as a blind redaction here. Worth noting the disclosure is real even though it did not show on my box - my models live at `C:\camelid-fork\models`, so no username appears; a normal install would.

**Minors.** `FirstRunCard` is now hidden on this surface - you are right that nothing enforces `--model`, and all four of its routes 403. `apiSurface.js` ends with a newline. The configuration note no longer claims a conflict that `ServerPolicyArgs::from_env` does not enforce; it now says the command line refuses the pair, the environment does not, and it does not matter because the mode has no anonymous form to fall back to.

**Gates on the commit:** fmt clean; `clippy --all-targets -D warnings` clean; `api::server::tests` 11, `executable_disclosure_tests` 3, `models_dir_resolution_tests` 8, `lan_key::` 4, all 0 failed; 42/42 validation scripts; `smoke:lan-auth` (32 protected requests, down from 33 - that difference is the `/api/runtime/gpu` probe the earlier commit removed), plus `smoke:ui`, `smoke:observatory` and `smoke:first-run`; both evidence gates; public scrub; `git diff --check` clean; no line-ending churn.

PR #682, opened 2026-08-20
feat(web): adapt LAN chat to compact viewports

karan68:
## Stack

Responsive child of [#681](https://github.com/timtoole02/Camelid/pull/681). **Merge #681 first.**

This branch is rebased onto #681's current green head, `0fc8f0e2`, so it includes all four parent commits, including `84aa66e1` (`discloses_host_details`) and the follow-up provenance/evidence fixes. After #681 merges, GitHub will reduce this PR to the responsive commits `95c6f417` and `0f8a54d2`.

Replaces the incorrectly located fork PR `karan68/Camelid#6`.

## Summary

- apply the compact toolbar and hidden horizontal tool strip globally at viewports 860px wide and below, including tablets and narrow desktop windows, not only LAN mode
- keep the model selector sticky and visible while secondary composer tools scroll; the send action remains outside the scroller
- adapt short landscape layouts by compacting the top bar, notice, message spacing, receipt spacing, composer input, and status row
- register window and `visualViewport` resize listeners once while still resizing the textarea on draft and thread changes
- add safe-area-aware inline and bottom padding for constrained landscape screens
- use a two-line notice clamp in portrait and a real one-line ellipsis in narrow landscape

## Measured layout

The authenticated browser smoke switches models, completes a streamed reply, then checks seven cells:

| Cell | Viewport | Toolbar | Notice | Message area |
| --- | ---: | ---: | ---: | ---: |
| Compact portrait | 360 x 800 | 40px | 64px | 521px |
| Phone portrait | 390 x 844 | 40px | 64px | 565px |
| Narrow phone landscape | 480 x 320 | 40px | 35px | 129px |
| Phone landscape | 844 x 390 | 40px | 35px | 199px |
| Short landscape | 740 x 320 | 40px | 35px | 129px |
| Tablet portrait | 768 x 1024 | 40px | 55px | 754px |
| Desktop control | 1440 x 900 | 68px | 55px | 602px |

At maximum horizontal tool scroll, the smoke now asserts that the sticky model selector and final generation-control button are both visible and do not overlap. Drawer open and close use Puppeteer's hit-tested clicks, including a measured visible point on the scrim. The tail foreign-origin, overflow, and reload/key-reuse checks are explicitly restored to 390 x 844 portrait first.

The 480 x 320 cell also pins the computed notice cascade to `display: block`, no line clamp, `white-space: nowrap`, and `text-overflow: ellipsis`.

## Validation

- #681 current head `0fc8f0e2`: 9/9 GitHub checks green
- frontend build: passed
- `smoke:lan-auth`: passed; seven viewport cells, authenticated model switching and streaming, real drawer hit-testing, max-scroll model/tool reachability, final viewport 390 x 844
- `smoke:ui`: passed
- validation harness self-tests: 42/42 passed
- inherited `executable_disclosure_tests`: 3/3 passed; `84aa66e1` and `0fc8f0e2` are both ancestors
- design-evidence checksum verification: 2/2 passed
- strict design-evidence privacy audit: 0 findings
- public scrub: passed
- editor diagnostics: no errors in touched source files
- `git diff --check`: clean

The general `npm run smoke` was not counted: it targets separately running services at `127.0.0.1:4175` and `127.0.0.1:8181`; no such preview/backend pair was started for this focused review run. The CI frontend job runs the repository's normal full smoke matrix.

### Negative control

Removing only the window/`visualViewport` resize listeners makes the short-landscape invariant fail: textarea 24px -> 40px, composer 104px -> 120px, message area 129px -> 113px. Restoring the listener returns the measured geometry above.

## Visual evidence

The capture path now preserves the same notice-visible state measured by the assertions. The images were regenerated and their SHA-256 manifest was resealed.

### Compact portrait - 360 x 800

Camelid compact portrait

### Short landscape - 740 x 320

Camelid short landscape

These are dated review evidence, not cross-platform pixel-golden tests. CI gates the built CSS through deterministic geometry, computed-style, and hit-testing assertions; pixel rendering remains browser/font/platform dependent. Parent #681 contains the separate physical Android authentication, model-switch, and inference evidence.


timtoole02 (commented):
Read the delta commit `c9074267` (9 files — the rest of the PR view is #681's commits), the CSS, the JSX, and the smoke additions, and test-merged the two branches together.

The engineering quality is high. A six-cell measurement table asserted on real geometry rather than screenshot diffs, plus a documented negative control that removed only the resize listeners and confirmed the intended invariant failed, is more rigor than most responsive work gets.

## Merge order — decide this first

This PR is based on `main` and carries `7f05d5cc` + `32c18200`, which is all of #681 **except `84aa66e1`** — the commit adding `discloses_host_details`, which stops `/health` handing out `executable` and `listen_addr` before authentication on a LAN Chat listener.

- `git merge-base --is-ancestor 84aa66e1 <this branch>` → no
- `grep -c 'fn discloses_host_details'` on this branch's `src/api/mod.rs` → 0

So merging this PR on its own would ship `--lan-chat-only` with that disclosure open. The description says the PR view holds "its two parent commits" — true when written, but #681 has three now.

Worth noting: `validation-scripts` passes here *because* this branch is pinned to the older `mod.rs` blob (`306571bc`), which matches the SmolLM3 pins. On #681 the same job is red for exactly that reason. The green check here is a symptom of the staleness, not evidence against it.

Nothing to untangle, though — I test-merged both branches and it is clean, no conflicts, `mod.rs` lands at `31782fff` with `discloses_host_details` intact and both SettingsView gates present. Merging #681 first resolves it.

## Findings

**The model selector can scroll out of view.** `.cxcomposer__model` is the first child of `.cxcomposer__tools`, which at ≤860px becomes `overflow-x: auto` with `scrollbar-width: none` and the webkit scrollbar hidden.

Send is safe — it lives in `.cxcomposer__actions`, a `flex: none` sibling of the scroller — but the model picker is inside it. `assertResponsiveLayout` only measures `modelSelect` at `scrollLeft: 0`, and `assertCompactInteractions` sets `scrollLeft = scrollWidth` but then checks only the last tool, so nothing covers the model control at max scroll.

For the feature whose headline is switching models from a phone, putting the switcher in a scroller with no visible affordance is the one call I would push back on. `position: sticky; left: 0` with a background on `.cxcomposer__model`, or lifting it out of the scroller next to `.cxcomposer__actions`, would fix it.

**640px → 860px is a global change, not a LAN-scoped one.** Icon-only labels and the hidden-scrollbar strip now apply to tablet portrait and to any desktop window under 860px. The measurement table shows this is deliberate, but the title and summary read as mobile-only — worth stating plainly in the description so it is not a surprise for desktop users with narrow windows.

**The smoke leaves the viewport in landscape.** The new block runs after the palette check, and `assertCompactInteractions` last sets `phone-landscape` (844x390). The tail `geometry` overflow assertion and the reload/key-reuse check now run there instead of 390x844. Coverage is not lost — `assertResponsiveLayout` covers 390 — but #681's "390px viewport" claim stops describing this run. One `setViewport` back to 390x844 before the foreign-origin check restores it.

**Evidence capture never runs in CI.** It is gated on `CAMELID_CAPTURE_LAN_RESPONSIVE=1`, which CI does not set, so the committed PNGs can drift from the CSS with nothing detecting it. The capture path also dismisses the notice first, so the images show a state the assertions never measured. The manifest itself verifies clean, 2/2 — no CRLF problem here, unlike the `lan-chat-e2e` manifest in #681.

## Minor

The drawer toggle and close use `$eval(btn => btn.click())`, which skips puppeteer's visibility and hit-testing checks — weaker than `page.click` for an assertion specifically about reachability.

The `useLayoutEffect` deps include `composer`, so both resize listeners are torn down and re-added on every keystroke. Correct, just churn; a separate `[]`-dep effect calling a ref'd resize would avoid it.

In `ui.css`, a ≤480px-wide landscape screen matches both new blocks. The 480 block sets `display: -webkit-box` with `-webkit-line-clamp: 2`; the landscape block then adds `white-space: nowrap; text-overflow: ellipsis`, which does not apply to `-webkit-box`. Result is a hard clip with no ellipsis. Narrow band, low impact.

karan68:
Addressed in `0f8a54d2` and rebased onto #681's current green head `0fc8f0e2`.

- **Merge order/security:** `84aa66e1` and `0fc8f0e2` are now ancestors of this branch. `src/api/mod.rs` retains `discloses_host_details`; its blob is the corrected `01d52b40`. #681 remains the required first merge.
- **Model selector:** it is sticky at the left edge of the compact tool scroller. The browser smoke scrolls to the maximum and asserts the model selector and final tool are simultaneously visible and non-overlapping.
- **Global breakpoint:** the title, summary, and evidence README now say plainly that the compact treatment applies globally at <=860px, including tablets and narrow desktop windows.
- **Final viewport:** the smoke restores 390 x 844 before the foreign-origin, overflow, and reload/key-reuse tail checks.
- **Evidence:** capture now keeps the same notice-visible state the assertions measured; both images were regenerated and the manifest resealed. The README explicitly treats screenshots as dated evidence rather than unstable cross-platform pixel goldens; CI gates the built CSS through geometry, computed-style, and real hit-testing assertions.
- **Drawer hit-testing:** synthetic `$eval(...click())` calls were replaced with Puppeteer's `page.click`. The smoke waits for the responsive rail transition, verifies the hamburger owns its center point, and closes through a measured visible point on the scrim.
- **Resize listeners:** registration now uses an empty-dependency layout effect; a separate effect still resizes on draft/thread changes.
- **Narrow landscape notice:** the landscape rule resets `display` and line clamp before applying one-line ellipsis. A new 480 x 320 cell pins the computed style to `block / none / ellipsis / nowrap`.

Validation: frontend build passed; `smoke:lan-auth` passed across seven viewport cells; `smoke:ui` passed; all 42 validation-script self-tests passed; inherited host-disclosure tests passed 3/3; design-evidence checksum verification passed 2/2; strict evidence privacy audit found 0 issues; public scrub and `git diff --check` passed; touched source diagnostics are clean.

PR #683, opened 2026-08-20
feat(fabric): learn completion-time placement

karan68:
## Summary

- add an opt-in resident `completion-time` placement mode while leaving `throughput` as the default
- learn workload-scoped service durations only from clean, queue-free buffered completions and clean, queue-free streaming EOF, then rank mature candidates by `(load + 1) * estimated service time`
- preserve least-load cold fallback until every eligible candidate has five fresh samples, invalidate node-attributable failures, expire estimates after five minutes, and bound in-memory state to 1,024 entries
- keep busy placements, queue-full refusals, client cancellation, affinity-selected successes, and client-backpressured streams out of timing samples
- reject the stateful mode from one-shot `fabric route` and `fabric run`, and document the complete operational contract

## Design

The estimate key includes node label and authority, backend/version, model, API path, streaming shape, request-byte bucket, and requested-output-token bucket. Prompt and response content are never retained.

Only a successful completion placed at observed-or-reserved load zero is a duration sample. Busy completions are skipped because total in-flight work can belong to other workload classes; dividing sojourn time by that count would fabricate a per-class service duration. Non-backpressure 5xx, unreachable/transport, malformed-body, and truncated-stream failures invalidate the relevant estimate and force a fresh five-sample warmup. Queue saturation and cancellation are not node-speed evidence. Streaming is sampled only at clean EOF and never after the relay has observed client backpressure.

The mature score uses integer nanoseconds with saturating `u128` multiplication. Mature ties break by load and then label. Estimate state is process-local and intentionally resets on restart.

## Validation

- `rustup run 1.89.0 cargo test --all-targets` with `CARGO_BUILD_JOBS=1`: passed
- focused fabric coverage: 304 passed, 0 failed
  - policy: 33 passed
  - fabric library: 215 passed
  - fabric end-to-end: 15 passed
  - resident proxy: 74 passed
- CLI tests: 8 passed, 0 failed
- pinned Rust 1.89 all-target Clippy with `-D warnings`: passed with only scoped allowances for verified pre-existing `duplicated_attributes` and `needless_range_loop` warnings in untouched code
- rustdoc broken-link gate: passed
- available `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- public scrub: passed
- VS Code diagnostics: no errors in touched Rust files

A parallel Windows all-target compile first hit `rustc` `STATUS_STACK_BUFFER_OVERRUN`; the one-job pinned retry compiled the same targets and completed the full suite successfully. Windows Defender later blocked a newly downloaded pinned `rustfmt.exe` with OS error 225, so CI remains authoritative for the exact Rust 1.89 formatting binary.

## Performance evidence

This PR makes no throughput or latency improvement claim. The paired F3 Macs and their tunnels are currently unavailable, so the required interleaved physical `throughput` versus `completion-time` campaign could not be run. The policy is opt-in and the existing default is unchanged pending that measurement.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -208,9 +208,10 @@ successful requests completed during this proxy process, then minimizes:
 The estimate is scoped to the node's label, address, backend and version; the model; the API route;
 streaming versus buffered delivery; and coarse power-of-two buckets for request bytes and requested
 output tokens. That prevents a node which happened to receive a longer prompt or generation from
-being labelled intrinsically slow. A measurement is wall time, so it also carries whatever the node
-was already working through; it is divided by the work the node reported in flight when the request
-was placed, because the formula above puts the queue back. No prompt or response content is retained.
+being labelled intrinsically slow. Only a completion placed while the node's observed-or-reserved
+load was zero is sampled. A busy completion includes unknown queueing from other workload classes,
+so dividing its wall time by total in-flight work would invent a service time the proxy did not
+observe. No prompt or response content is retained.
 
 Every eligible node needs five successful completions in the same workload class before an estimate
 may decide placement. Until then the proxy uses least-load, rotating equal-load cold candidates by
@@ -218,12 +219,13 @@ least-recent selection so sequential traffic samples all of them. The response h
 `x-camelid-fabric-reason: LeastLoaded` while cold and `EstimatedCompletion` once learned timing made
 the decision.
 
-Only clean success is a speed sample. A node-attributable 5xx, unreadable answer, transport failure,
-or truncated stream invalidates that node's estimate for the workload immediately. A queue-full
-refusal does not: it says the load bound worked, not that service became slow. Client cancellation
-does not either, because it says nothing about the node. A stream is sampled only at clean EOF, and
-not if the bounded relay channel filled behind a slow client. A request carrying a sticky header is
-not sampled either: affinity, rather than this policy, chose its node.
+Only a clean, queue-free success is a speed sample. A node-attributable 5xx, unreadable answer,
+transport failure, or truncated stream invalidates that node's estimate for the workload
+immediately. A queue-full refusal does not: it says the load bound worked, not that service became
+slow. Client cancellation does not either, because it says nothing about the node. A stream is
+sampled only at clean EOF, and not if the bounded relay channel filled behind a slow client. A
+request carrying a sticky header is not sampled either: affinity, rather than this policy, chose its
+node.
 
 Estimates older than five minutes become cold and have to be sampled again. They are memory-only,
 bounded to 1,024 node/workload entries, and disappear when the proxy restarts. `fabric route` and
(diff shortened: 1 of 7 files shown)


karan68:
@timtoole02 self-review before you spend time on this one - I found a defect in my own policy and pushed the fix as `8e9dbbfc`.

The doc and the formula were already right. `CONFIGURATION.md` promises "an exponentially weighted service time" minimising `(load + 1) * learned service time`, and that is the correct model for this engine: one decode slot, jobs queue behind each other.

The sample was wrong. `forward.rs` starts the clock immediately before the request goes to the node and stops it when the response is complete, so a measurement is **sojourn** time - it already contains the queueing the node imposed at that moment. Selection then multiplied it by `(load + 1)` a second time, so load was counted twice and speed was under-weighted.

I proved it against the real types before touching anything, rather than arguing it. Scratch test, since reverted: `fast` needs 40ms of service and is carrying 3 requests, `slow` needs 400ms and is idle, so fast's honest sample is 160ms:

```
chose slow (EstimatedCompletion); true cost fast=160ms slow=400ms
```

The mode that exists to find fast nodes picked the one 2.5x worse. And it compounds, because this policy sends more traffic to fast nodes: the more work a fast node earns, the more its samples inflate, the worse it scores.

The fix is small and does not move the design. `Placement::record_success` now records `elapsed / (in-flight + 1)`, using the load the node itself reported at placement - which `Placement` already carries, so there is no new field and no policy-module change. `ahead = 0` divides by one, so anything measured against an idle node is byte-identical to before and every existing fixture is untouched. Two tests pin it: a pure one on the arithmetic, and `a_fast_busy_node_beats_a_slow_idle_one` on the ranking. One sentence added to the doc.

Counts, measured with my three files stashed and then restored so the delta is attributable: `--lib fabric::` 216 -> 218 (the two tests here), `fabric_serve` 74 and `fabric_end_to_end` 15 both unchanged. Small correction to my own PR body while I am here - it says "fabric library: 215", but the same command reports 216 at the PR head, so that number was off by one.

Gates on the commit: fmt clean, `clippy --all-targets -D warnings` clean, public scrub clean, `git diff --check` clean, all three files LF.

Two things I deliberately did **not** change, because they are judgement calls rather than defects, and I would rather you decide:

1. **The maturity condition may rarely be met in real traffic.** Every eligible node needs five fresh samples in the *same* class inside five minutes, and the class includes a power-of-two request-byte bucket. A conversation that grows walks through buckets, so the proxy may sit in least-load fallback most of the time - inert rather than wrong. There is no measurement either way yet, which the PR body already says.
2. **One node that keeps failing disables the mode fabric-wide**, since `has_complete_estimates` requires every candidate mature and a failure resets that node to zero. Conservative and documented, but worth knowing it is fabric-wide rather than per-node.

karan68:
Correction to my earlier self-review comment: the final head no longer divides a busy request's sojourn time by `(in-flight + 1)`.

That rule assumed the work ahead had the same service cost and workload class. Node health exposes only a total in-flight count, so heterogeneous work makes that division an invented per-class duration. The final, conservative rule samples only clean completions placed at observed-or-reserved load zero. Busy completions are served normally but do not change learned speed.

The final head also carries the #676 rebase and uses the same `max(observed, reserved)` load for both selection and the queue-free sampling gate. Focused negative/positive controls pass, as do 230 Fabric tests, Fabric CLI integration, strict Clippy, fmt, and diff checks. No performance claim is added; fresh CI is running on `f5fdb21b0e1c56d1e601e9a64e1d833b267e6fe9`.

Issue #692, opened 2026-08-24
Docs suggestion: point external harnesses at /v1/health rather than /v1/models for readiness


cdnwetzel:
## Summary

A docs suggestion, not a bug report. Nothing is broken.

An external harness polled `GET /v1/models` and treated any HTTP 200 as readiness. Camelid answers 200 with `{"object":"list","data":[]}` while weights are still loading, so requests issued immediately after came back:

```json
{"error":{"message":"model is not loaded","type":"invalid_request",
          "code":"model_not_loaded"}}
```

## This was the reporter's bug, not Camelid's

`/v1/health` already exposes a proper readiness contract, and it is public by design so a load balancer can probe it without a credential:

```rust
pub loaded_now: bool,
pub generation_ready: bool,
pub active_model_id: Option<String>,
pub execution_plan: Option<ExecutionPlan>,
```

The correct check is `generation_ready`, with `active_model_id` naming what is loaded, so readiness and model identity come from the same response and cannot disagree. Our harness now uses it.

## The only thing worth considering

An OpenAI-compatible client that knows nothing about `/v1/health` will reasonably read an empty `/v1/models` as "server has no models" rather than "server is still loading". Whether that is worth a line in the docs, or a distinct status while loading, is entirely your call.

Raising it because the failure mode cost real debugging time on this end, and because it produced a receipt that looked like a platform difference and was not. That misreading is the kind of thing worth writing down for the next person doing external validation.

## Not claimed

No claim that current behaviour is wrong. `v0.6.1-191-gdb283b69`.

Issue #691, opened 2026-08-24
distribute-worker started before the master reports EHOSTUNREACH and never retries (macOS worker)


cdnwetzel:
## Summary

If `distribute-worker` starts before `distribute-master`, the worker's first attempt to reach the master's feedback port hits a port nothing is listening on yet. On macOS that returned `EHOSTUNREACH` rather than a connection refusal, and the worker did not recover: it sat silent, then exited after 600 seconds with an error that points at the network.

The same ordering works between two Linux hosts.

## Reproduce

```bash
# worker first, on a macOS host
camelid distribute-worker --layers "14..28" --addr 0.0.0.0:5005 \
  --master-addr <master>:5006 model.gguf

# then master, on a Linux host
camelid distribute-master --worker-addr <worker>:5005 --layers "0..14" \
  --addr 0.0.0.0:5006 model.gguf
```

## Observed

Worker, after ten minutes of silence:

```
Failed to connect to <master>:5006 after 600 seconds: No route to host (os error 65)
```

Master, meanwhile, reported success connecting downstream and then hung indefinitely waiting for an upstream connection that never arrived:

```
Connecting to downstream <worker>:5005...
Connected successfully to <worker>:5005!
```

It never reaches `Accepted connection from upstream/client`.

## What it is not

Reachability was verified independently in both directions while the pair was stalled: `ping` at 0% loss, and the master's port confirmed open from the worker host. Both hosts had `nftables` and `iptables` stopped. The identical worker-first ordering succeeds between two Linux hosts.

## Workaround

Start the **master** first. The ring then forms immediately and the run completes normally:

```
Cluster worker execution loop active!
[distributed] decode: 47 tokens in 3.42s = 13.76 tok/s
```

## Environment

Master: Linux x86_64, CUDA (Blackwell). Worker: macOS arm64, Metal. `v0.6.1-191-gdb283b69`, `Qwen3-0.6B-Q8_0`, 28 layers split 0..14 / 14..28.

## Not claimed

Observed on one macOS host against one Linux master. No claim the retry logic is wrong, only that a first-attempt `EHOSTUNREACH` is not recovered from while the equivalent Linux path is. It is possible this is specific to how macOS reports a connection to a closed port on a live host, in which case the useful change may be documentation (start the master first) rather than code. That is a maintainer call.

Found while running the distributed lane across a six-machine fleet; the successful cross-ISA run it led to is in the companion PR.

Issue #690, opened 2026-08-24
Two tests fail above ~16 test threads and pass serialised; CI's 2-4 core runners never see it


cdnwetzel:
## Summary

Two tests fail when `cargo test` runs with many test threads, and pass when serialised. CI never sees this because GitHub-hosted runners are 2 to 4 cores.

Affected:

- `api::tests::smollm3_exact_renderer_is_architecture_scoped_and_not_generic_qwen3`
- `gait::tests::memory_measurement_is_sane`

## Reproduce, one flag

```bash
# fails on a many-core host
cargo test --all-targets --all-features

# same host, same tree, passes
cargo test --all-targets --all-features -- --test-threads=1
```

## Evidence that this is parallelism, not hardware

Six runs across three machines, in both directions:

| Host | Logical cores | `--test-threads` | Result |
|---|---|---|---|
| Ryzen 9 5950X | 32 | default | `1951 passed; 1 failed` |
| Ryzen 9 5950X | 32 | **1** | `1952 passed; 0 failed` |
| Xeon E5-2699 v4 | 44 | default | `1950 passed; 2 failed` |
| Xeon E5-2699 v4 | 44 | **1** | `1952 passed; 0 failed` |
| i5-1135G7 | 8 | default | `1952 passed; 0 failed` |
| i5-1135G7 | 8 | **32** | `1951 passed; 1 failed` |

The last row is decisive: an 8-core host reproduces the failure when forced to 32 test threads. The failure follows the thread count, not the machine.

Either test also passes when run alone on a host that fails it in the full suite:

```
test api::tests::smollm3_exact_renderer_is_architecture_scoped_and_not_generic_qwen3 ... ok
test result: ok. 3 passed; 0 failed; 2035 filtered out
```

## The two appear to fail for different reasons

`smollm3_exact_renderer_...` panics at `src/api/mod.rs:28755`:

```
template dispatch must fail before generic Qwen3 rendering: RenderedPrompt {
  text: "<|im_start|> :: Knowledge Cutoff Date: June 2025 :: strftime_now ::
  Reasoning Mode: :: /system_override", add_special: true, parse_special: true }
```

The renderer resolved to the generic Qwen3 path where the test asserts dispatch must refuse first. That reads like shared state rather than a timing effect, since it passes deterministically alone and fails deterministically under load.

[...]

Issue #689, opened 2026-08-24
cargo test --all-targets --all-features fails with 226 errors on macOS: --all-features enables cuda


cdnwetzel:
## Summary

On macOS arm64, `cargo test --all-targets --all-features` fails on a clean tree with 226 compile errors, so the suite never runs. Both `CONTRIBUTING.md` and the PR template instruct contributors to run this exact command, with no platform caveat.

## Reproduce

```bash
# macOS arm64, clean tree
cargo test --all-targets --all-features
```

## Observed

```
error[E0433]: cannot find module or crate `cudarc` in this scope
error[E0599]: no method named `is_prism_low_bit` found for struct `RawMat`
error: could not compile `camelid` (lib) due to 226 previous errors
error: could not compile `camelid` (lib test) due to 229 previous errors
```

## Cause

`--all-features` enables the `cuda` feature regardless of target, because Cargo does not filter features by platform. `Cargo.toml` states the intent plainly:

> "macOS never has it (Metal is its GPU lane)."

`build.rs` correctly declines to set the `cuda` cfg on macOS, but an explicit `--all-features` overrides that.

## Evidence it is the flag, not the code

Same host, same tree, dropping one flag:

```
cargo test --all-targets                 ->  exit 0, suite green
cargo test --all-targets --all-features  ->  226 compile errors
```

## Environment

Mac Mini M4, macOS, arm64, `v0.6.1-191-gdb283b69`, built from source.

## Not claimed

No fix is proposed. Whether the right answer is cfg-guarding the CUDA code with `not(target_os = "macos")`, making the documented gate platform-aware, or something else depends on how the feature is meant to interact with the Windows and Linux default-on paths, which is a maintainer call. Only Apple silicon macOS was tested; Intel macOS was not.

Found while running the documented gates across a six-machine fleet. Full context and sealed evidence in the companion PR.

PR #693, opened 2026-08-24
qa: external platform receipts from a six-machine heterogeneous fleet

cdnwetzel:
## Summary

- External platform validation of `v0.6.1-191-gdb283b69` on six machines outside the CI matrix: four x86_64 Linux hosts (8 to 44 logical cores, AVX-512 and AVX2, Ampere and Blackwell CUDA) and two Apple silicon hosts.
- Adds `FLEET-RECEIPTS.md` with seven evidence-gated receipts, plus 19 platform bundles and 18 run-to-run repeat bundles under `qa/evidence-bundles/fleet-2026-08/`.
- Adds four portable harness scripts that run unmodified on macOS and Linux and seal their own `SHA256SUMS`.
- Promotes no support row and changes no compatibility wording.

## Why this change exists

CI runs `ubuntu-latest`, `macos-14` and `windows`, all GitHub-hosted runners with no CUDA GPU, no Blackwell, no AVX-512 guarantee, and 2 to 4 cores. This contributes measurements from hardware that matrix cannot reach, in the repo's existing evidence-bundle format, with every non-claim stated.

The receipts credit existing work rather than re-claiming it. F-001 opens by quoting `qa/determinism/determinism-baseline-20260614T063455Z.md`, which already establishes the thread-count invariant and already predicts the ISA divergence. Only one of its three parts is new, and it says so.

## What is actually new

1. **AVX-512 and AVX2 produce byte-identical output** for the tested row. The determinism baseline's model expects hosts on different kernels to give "different-but-internally-deterministic" results; two AVX-512 hosts and two AVX2 hosts do not differ. 66 independent runs, two digests, split exactly on instruction set, with the AVX-512 lane carrying more runs (27) than AVX2 (18).
2. **First measured x86 versus ARM divergence** for this engine, located to a single near-tie token.
3. **The engine builds and runs on Blackwell** (compute capability 12.0), currently recorded as `Phase 5 (Blackwell) BLOCKED-HW`. NVFP4 was not exercised and is explicitly not claimed.
4. **A distributed pipeline across x86_64+CUDA and ARM64+Metal**, run as a controlled pair against a homogeneous x86 control. Existing `qa/distributed/` receipts are ARM plus ARM; this adds the first x86 node and the first CUDA plus Metal ring.
5. **Offload planning at 4 GB and at full residency**, bracketing the existing 6 GB record from both sides.

## Headline

Every host passes the complete 1,952-test suite once the two parallelism issues below are accounted for. **No correctness defect was found in the engine.**

## Related issues found while running the documented gates

- #689 `--all-features` enables `cuda` on macOS, 226 compile errors (why the checkbox below is unticked)
- #690 two tests fail above ~16 test threads, pass serialised
- #691 `distribute-worker` started before the master does not retry
- #692 docs suggestion on `/v1/health` versus `/v1/models` for readiness

## Validation

- [x] `git diff --check`
- [ ] `cargo fmt --all -- --check` (not applicable, no Rust changed)
- [ ] `cargo test --all-targets --all-features` (**cannot run on macOS, see #689**; the suite is green on all six hosts via the forms that do run, recorded in the receipts)
- [ ] `cd frontend && npm run build` (not applicable, no frontend changed)
- [x] Other evidence attached: 39 sealed bundles, each with `SHA256SUMS`

Change type is Docs-only plus QA/evidence-publication per `docs/VALIDATION_MATRIX.md`, so the lane run was `git diff --check`, `check-public-evidence-claims.mjs`, `check-public-scrub.sh`, and `audit-evidence-bundle-privacy.mjs --strict`.

## Public-repo privacy check

- [x] No private hostnames, SSH commands, user home paths, key paths, local validation details, raw SSH stderr, or raw infrastructure failure output
- [x] `bash scripts/check-public-scrub.sh`
- [x] `node scripts/audit-evidence-bundle-privacy.mjs --root qa/evidence-bundles/fleet-2026-08 --strict`, `finding_count: 0`

Two publication-time transforms, both documented in the receipts header rather than done silently: operator paths were redacted from `raw/` streams, and 20 oversized `inspect` artifacts were excerpted, since `inspect` dumps the full tokenizer vocabulary at 7.1 MB per model row (147 MB total, reduced to 6.0 MB). Each excerpt records the original byte count, line count, sha256, and regeneration command. `SHA256SUMS` was recomputed after both transforms and every bundle verifies as published.

## Support-contract check

- [x] This PR does not overclaim support
- [x] TinyLlama Q8_0 remains the current full-support gate; no Llama row wording is touched
- [x] `COMPATIBILITY.md`, `docs/reference/STATUS.md` and `/api/capabilities` are untouched by design

Every receipt carries an explicit "not claimed" section. No parity claim, no throughput claim, no NVFP4 claim, no support-row promotion. Two bundles produced before a defect in our own harness was found are retained under `superseded/` with the defect and its correction written up, rather than deleted, since a receipt reporting `diverged_count: 0` across seven runs that all failed is exactly the kind of artifact that should not quietly disappear.

## Docs impact

Adds `FLEET-RECEIPTS.md`. No existing doc is modified.


timtoole02 (commented):
Thank you for this — it is a serious piece of work, and the parts that are backed by the bundles are exactly the kind of contribution this repo is set up to accept. The harness is portable, the bundles are sealed and self-verifying, and the "not claimed" discipline is the right instinct.

I ran the repo's own gates plus a claim-by-claim audit of `FLEET-RECEIPTS.md` against the 934 committed files. **What passes is solid. What blocks merge is that several stated results have no committed receipt.** In a repo whose whole contract is evidence-gating, that is the one defect class I can't wave through, so I'm asking for changes rather than merging.

---

## What I verified holds

- All 39 new bundles verify: `sha256sum -c` clean in every directory, and `scripts/check-evidence-bundle-checksums.sh` passes repo-wide (203 bundles).
- `scripts/audit-evidence-bundle-privacy.mjs --strict` → `finding_count: 0`, on both the fleet root and the default root.
- `scripts/check-public-evidence-claims.mjs`, `check-ledger-schema.mjs`, `check-ledger-drift.mjs`, `check-fit-verdict-vocabulary.mjs`, `check-cuda-prefill-parity-gate.mjs`, the README contracts — all pass.
- The three `camelid pull` catalog ids resolve in `ledger/camelid-ledger.json`; every flag in F-007's reproduce block exists in `src/main.rs`.
- The `qa/distributed/` prior-art characterisation is exact.
- All 104 `req-t*.json` carry `temperature 0 / top_p 1 / seed 0 / max_tokens 48` with an identical prompt, so F-001's controlled conditions genuinely hold.
- Exactly two completion digests across all 97 completed runs, splitting cleanly x86 (67) vs arm (30) — the underlying observation is real.
- The three model `sha256`s are identical across all seven platform hosts and match the repo's pre-existing bundles.
- All 20 excerpted `inspect` artifacts carry the documented byte-count / line-count / sha256 / regenerate footer.

---

## Blockers

**B1 — F-007 is graded `Attested` but ships no evidence.** The PR commits zero distributed artifacts: `git diff --name-only db283b69..HEAD | grep -i distrib` returns nothing. The "byte for byte" result therefore cannot be checked by anyone. Separately, the digest offered as proof —

```
control (x86 + x86) sha256 ecedad706a5c0658bfabdfcc3e5605e1
```

— is 32 hex characters. A SHA-256 is 64. Please either commit the two run bundles and re-seal, or drop F-007 to an observation with the grade removed.

**B2 — F-005 and F-006 state Qwen3-4B results for the 8 GB MacBook, which never ran that row.** `apple-silicon-metal-8gb-unified-20260822T145927Z-head-db283b6/` contains only `Qwen3-0.6B` and `Qwen3-1.7B` artifacts. The F-005 table row for that host, and the conclusions drawn from it, have no receipt.

**B3 — F-006's coverage numbers don't match the tree.** The claim is `verify` "exited 2 on all three model rows on all six hosts", graded "six hosts, 19 bundles". Only 7 of the 19 platform bundles contain a `verify` artifact, and per B2 the 8 GB host has two rows, not three.

**B4 — F-004's "contended" condition has no receipt.** The table's middle row (RTX 5060 Ti contended, 2.6 GiB free, "0 resident, planner refused") appears in no bundle. Both `linux-x86_64-cuda-16311mb-*` bundles record `free 15684` — the free-GPU state — so the 258 s figure attributed to the contended run comes from a bundle that self-identifies as the free run.

**B5 — F-002 credits a host that produced no R-007 receipt, and omits the one that did.** The six committed `r007-retest-*` bundles come from: Precision T5810 (44T AVX2, A4500), Mac Mini M4 ×2 (the v0.4.4 / v0.6.1 pair), XPS 9510 (16T AVX-512, 3050 Ti), **NUC11 (8T AVX-512, no GPU)**, and MacBook Neo (6T NEON, A18 Pro). The "confirmed on four more hosts" table lists **ASRock B550 / RTX 5060 Ti Blackwell**, which produced no R-007 bundle at all, and omits the NUC11, which did.

**B6 — F-001's arm64 half was served by the Metal resident engine, not the NEON CPU kernel.** The receipt claims the "`Qwen3-0.6B-Q8_0` **CPU lane** … differs from the **NEON path** by one token" and explains the split as "f32 reduction order differs between the AVX and NEON paths". The Apple serve logs say otherwise:

```
# Mac Mini M4 — cpu-thread-determinism-20260822T141155Z-.../raw/serve-t1.err
[hw] GPU: Apple M4 | Metal compute | unified-memory yes | GPU-resident inference available
  Warming up the GPU (building the resident engine, one-time)…
[camelid] CAMELID_METAL_NOCOPY: loading Q8_0/K-quant/Prism Q1_0/Q2_0 weights as page-aligned wire pages

# Precision T5810 (x86) — same field, same run
[camelid] WARNING: CAMELID_LAZY_Q8_0_LINEAR is set; Q8_0 weights will stream from disk per token
```

`CAMELID_LAZY_Q8_0_LINEAR` — listed under the receipt's own Environment — appears in all 67 x86 logs and **zero** Apple logs. Corroborating: every Apple `resp-t*.json` has `prefill.layers_total: 0` and empty `layers` / `first_token_layers`, while every x86 response carries 28 per-layer timings; Apple `generate` is 546–1048 ms against 4427–45370 ms on x86; and Apple wall clock is flat across thread counts (553/546/551/547/552 ms for 1/2/4/8/10), which is what a GPU-resident decode looks like and not what a Rayon-partitioned CPU kernel looks like.

So the measured divergence is **x86 CPU vs Apple Metal**, not AVX vs NEON. That doesn't destroy the finding — the two-digest split is still real and still interesting — but the mechanism sentence, the `NEON` labels in the lane table, and the extension of the thread-count invariant to the Apple hosts all need restating. The same document already calls the M4 "ARM64+Metal" in F-007.

Worth fixing at the source too: `cpu-thread-determinism.sh` says in its own comment that `/v1/health`'s `execution_plan` / `backend` identify the serving lane, but `wait_for_model` only reads `generation_ready` / `active_model_id`, so nothing records the backend. Capturing the `/v1/health` body per run into `raw/` would have made this self-evident.

---

## Majors

- **`scrub-bundle.sh`'s superset claim is false.** The header says "The pattern set is a superset of Camelid's own `scripts/check-public-scrub.sh`, so a bundle that passes here also passes the repo's guard." Comparing the two arrays, ~13 guard patterns have no counterpart: `ssh: connect to host`, `Operation timed out`, `rc=255`, `StrictHostKeyChecking=accept-new`, `Documents/cert`, the four literal IPs, both `stderr … evidence bundle` phrases, `target/model-promotion-host-*`, `/Desktop/Code/backend`. Either add them or drop the claim.
- **`fleet-receipts.sh` depends on `timeout`** at lines 79, 114, 130, 146, 157. Stock macOS has no `timeout`, so on a Mac orchestrator every host is reported unreachable — while `FLEET-RECEIPTS.md` says "the scripts run unmodified on macOS and Linux".
- **`camelid-receipts.sh` emits invalid JSON when `lscpu` is absent or localized** (line 86 feeds line 274), and the bundle is still sealed and reported as written.
- **`FLEET-RECEIPTS.md:23-24` cites a privacy proof that cannot see most of this PR.** `scripts/audit-evidence-bundle-privacy.mjs:10` filters on `['.json','.md','.txt','.log','.tsv']`, so it never opens `.out` / `.err` — 502 of the 934 added files, which is exactly the `raw/` stdout/stderr class your own `scrub-bundle.sh` header names as the exposure, and where all 130 `[REDACTED]` markers sit. There is **no live leak** (zero `/Users/` or `/home/` survivors in the tree), but the stated proof does not cover what it claims. Note also that none of `check-public-scrub.sh`'s three home-path patterns match a Gentoo `/home/<user>/` path. That gap is ours, not yours — I'll take it as a follow-up issue.
- **Dead cross-references.** `FLEET-RECEIPTS.md:18` and `:500` point at `fleet-evidence-clean/`, which does not exist — the bundles landed at `qa/evidence-bundles/fleet-2026-08/`. `:464` cites `FLEET-DEFECTS.md`, which this PR does not add. `:364` cites a `qa/agent-eval/…-FAIL.json` receipt that is not in the repo or the PR. `:367` documents a `plan-offload --detected` flag that is not in the CLI.
- **No receipt cites its bundle.** None of F-001…F-007 names the committed directory it rests on, so no claim can be traced to its evidence without reverse-engineering the tree. Please add the bundle path to each receipt.
- **`pxx` / `R-007` are unresolvable here.** F-002's entire premise is "pxx receipt R-007", and neither string appears anywhere else in the repo. Please link it or restate the finding self-containedly.
- **Internal contradiction on repeats.** The document-wide boundary at `:538` says "One run per configuration. No variance repeats." — contradicted by the 18 repeat bundles this PR adds and by F-001's own n=3 section.
- **Thermal-envelope claim is over-general.** `:505` "Every bundle now records PL1, PL2, CPU governor, energy performance preference, and chassis type" holds for 5 of 39 bundles — and not for the two 13:xx bundles that supply F-004's 258 s figure. Both manifest shapes ship under one unbumped `camelid.external-platform-receipts.v1` schema id. The fleet table also lists two laptops, not one, and `chassis_kind()` hard-returns `unknown` off Linux.
- **`CROSS-PLATFORM-RECEIPTS.md` reads as a stale pre-run plan**, not a deliverable: it is not mentioned in the PR description, it describes both Macs as 8 GB when the fleet that ran has a 16 GB M4 (which collapses its "bandwidth only" experiment), section A says the procedure "adds a row to his matrix" (contradicting the no-promotion statement and the repo's support rule), and it refers to the maintainer in the third person. I would drop it, or fold the still-accurate procedure into `FLEET-RECEIPTS.md`.
- **Minor factual slip:** `:53` "Every CUDA bundle in `COMPATIBILITY.md` is … RTX 3060 Laptop 6 GB" — the matrix also records CUDA parity evidence on an RTX 4060.
- **F-001 counts:** please re-derive "identical for 21 tokens". In the committed completions the shared prefix runs through `…One is not among them because`, which is materially longer than 21 tokens. The "one decision, then natural divergence" framing itself is fine. Also `:78` "four x86 microarchitectures" — the sentence names three, and the fleet has four x86 hosts.

---

## CI

The only failing gate is `check-public-scrub.sh`, and it is self-inflicted: `scrub-bundle.sh` lists the guard's own regexes as literals. The repo guard already solves this for itself with adjacent-string concatenation — the same trick works here, with byte-identical runtime behaviour:

```diff
-  '/Users/[^/[:space:]"]+'
+  '/Users''/[^/[:space:]"]+'
...
-  'ssh -i'
+  'ssh ''-i'
```

plus rewording the bare `".pem"` in the line-22 comment. `ci-gate` fails only downstream of this. Every other step of the `public-scrub` job passes locally.

Also worth taking Copilot's two notes on that file: `--extra` with no value dies on `set -u` before the usage message, and the exit code there is 1 where the header documents 2.

---

## Placement

Seven files land at the repository root. The repo keeps harness scripts in `scripts/` and dated validation write-ups in `qa/validation-notes/`, and `DOCS.md` indexes the root docs. Suggested:

- `FLEET-RECEIPTS.md` → `qa/evidence-bundles/fleet-2026-08/FLEET-RECEIPTS.md` (next to the bundles it cites), with a pointer added to the evidence map in `qa/evidence-bundles/README.md`
- the five `.sh` → `scripts/fleet/` or `qa/evidence-bundles/fleet-2026-08/harness/`
- update the `Reproducing this` block to the new paths

One operational note: the added tree raises `check-evidence-bundle-checksums.sh` from 1834 to 2728 checksum entries, and that script spawns a `git` process per entry — measured roughly +42% wall on that gate. Not a blocker, and the fix is on our side.

---

To be clear about the shape of this: **the bundles are good and I want this in.** The blockers are all the same edit — bring the prose down to exactly what the 39 sealed bundles support, and commit the receipts for anything you want to keep claiming. If it is easier, narrowing the document is a perfectly good resolution; nothing here needs new hardware runs except F-007 and F-004's contended row.

PR #708, opened 2026-09-02
feat(agent): harden the agent harness and fix dual-stack IPv4/IPv6 web research

karan68:
# feat(agent): harden the agent harness and fix dual-stack web research

8 files, +1436/-29. The CUDA work that used to be bundled here is now #711.

## Summary

Six changes to the native agent harness, one network fix, and one classifier
widening.

1. **Multi-tier `edit_file` matching.** An `old` block that differs only in line
   endings, or by a *uniform* indentation shift, now applies instead of failing.
   Tiers are tried in order and **each requires a unique match**:
   1. exact substring
   2. exact after CRLF/LF normalization (the replacement is re-encoded to the
      file's own line endings)
   3. line-by-line equality after trimming, where every line in the window
      differs from `old` by the *same* number of leading spaces, the leading
      whitespace runs are comparable, and the shift can be applied to every line
      of the replacement

   Anything less certain is **not** applied — there is no similarity threshold
   on the mutation path. A tolerant match says so in its result
   (`edited X (matched lines 12-18 after adjusting indentation/whitespace)`).
2. **Actionable near-miss diagnostics.** When no tier matches, the error names
   the closest window with line numbers and a snippet so the model can re-read
   instead of guessing. This path is read-only, and both its window scan and its
   Levenshtein scoring are budget-bounded, so a failed edit on a large file
   cannot pin a core.
3. **Ambiguity reports line numbers** — `old text is not unique (3 occurrences
   at lines 12, 48, 96)` instead of a bare count.
4. **Fuzzy sandbox path suggestions.** A missing path suggests the nearest file
   in the sandbox (`Did you mean 'src/chat/tools.rs'?`). The walk is bounded to
   64 directories / 500 files.
5. **`search` gains `path_filter`.** Exactly three forms are supported —
   `*.ext`, `dir/**`, and a plain file name or path fragment. Anything else is
   an explicit error, because a silently empty result set reads to the model as
   "the symbol is not there". Patterns and paths are normalized to `/` so a
   `dir/**` filter behaves identically on Windows.
6. **Working-set ledger in compaction.** The compaction summary now leads with
   the files modified, the files inspected, and the last command with its
   outcome. The lists are capped, because this summary exists to *save* context.
7. **Premature-completion guard.** On the benchmark profile, if the model
   answers after editing a file without having run any command, the loop asks
   for one — **once** — then accepts whatever it says next. A command that ran
   counts whatever it reported.
8. **IPv4/IPv6 dual-stack web research.** A dual-stack DNS answer on a
   single-stack host handed curl records that could never connect, and the hop
   could die with `couldn't connect to server` instead of using the usable
   record beside it. The families the host cannot route are no longer pinned,
   decided from the host's own reachability rather than from the shape of the
   DNS answer, so it holds in both directions. The set is narrowed, never
   emptied, so the SSRF pin still binds every address curl may reach.
9. **Conversational weather queries** are recognized by both the frontend and
   backend classifiers.

## Key changes

- `src/chat/tools.rs` — tiered `edit_file`, bounded near-miss diagnostics,
  `suggest_path_in_sandbox`, `PathFilter` parsing and matching.
- `src/chat/agent.rs` — working-set ledger in `compact`, richer tool-call
  digests, one-shot premature-completion guard in `run_loop`.
- `src/api/web_research.rs` — `host_reachable_families`, `addresses_to_pin`,
  weather needles.
- `src/fabric/http.rs` — **unrelated to the agent harness, and listed here after
  arriving undeclared in an earlier revision.** The TLS test stub did not drain
  the request body before closing, which is an abortive close (RST) and made a
  fabric TLS test fail intermittently on macOS. Same fix `1afd0f030` already
  applied to the plaintext stub. Happy to split it out.
- `frontend/src/lib/webResearch.js` — matching weather patterns.
- `tests/fixtures/websearch/classifier_cases.json` — shared cases pinning the
  Rust and JS classifiers to the same answers.
- `qa/benchmarks/agent/native-contracts-v1.json` +
  `tools/bench/system/test-native-contracts.mjs` — 5 new contracts, 21 total.

## Review round 2 — `9ce7f13d`

Both blocking findings were real bugs on the mutation path, and both are fixed
where the review suggested.

- **Tier 3 silently rewrote indentation in tab-indented files.** Indentation was
  measured in leading *spaces*, so two tab runs of different depth both measured
  zero, the delta computed to zero, and the replacement was written verbatim
  with the model's tab count. `find_tolerant_line_matches` now rejects a window
  unless the two leading-whitespace runs are byte-identical or both space-only;
  anything else goes to the near-miss diagnostic. The exact reproduction from
  the review is now
  `a_tab_indented_file_is_not_rewritten_by_a_shallower_tab_count`.
- **A negative shift flattened relative indentation inside `new`.**
  `adjust_indentation` clamps per line, so a line with less indent than the
  shift landed at column 0 while its siblings moved fully. New
  `indent_shift_applies` gates tier 3: every non-blank line of the replacement
  must be able to absorb the shift, otherwise the edit is refused. The clamp is
  now a floor, not a fallback. Pinned by
  `an_indent_shift_that_would_flatten_the_replacement_is_refused`, which also
  asserts a *valid* negative shift still applies.
- **`addresses_to_pin` was the mirror of the bug it fixed.** Rather than mirror
  it back, `host_reachable_families()` asks the host: a UDP `connect` per
  family, which sends no packet and only needs a route. Narrowing now happens in
  whichever direction the host needs; both-reachable or neither means don't
  narrow; the set is still never emptied, so the SSRF pin keeps binding.

The three minor items are addressed too: a `search` whose single-file root is
excluded by the filter now says so instead of reporting no matches; the
Levenshtein cell budget carries a comment about its top-down bias; and the
compaction ledger renders paths through `normalize_workspace_path`, matching
`mutated_files`.

**One thing the full suite caught that targeted runs did not.**
`write_and_edit_through_the_tool_path_are_checkpointed` clears the process-wide
checkpoint log and asserts an exact count, but no agent test took `cp_lock` or
disabled checkpoints. Pre-existing race — the two new mutating tests just made
it reproduce. Benchmark-profile agent tests now mirror what production already
does on that path. It passes in isolation either way, so only the full suite
reveals it.

## Review round 1

The first four commits were reviewed and corrected in place before that.

**Blockers**

- **clippy `unnecessary_map_or`** in `compact` failed all three rust CI legs, so
  nothing after it ever ran.
- **The verification guard could trap the agent.** `run_shell` returns
  `ToolOutcome::Err` for any non-zero exit, and only a non-error cleared the
  flag — so an edit followed by a legitimately *failing* test could never be
  reported. Every text step was rejected until `max_steps`, ending the run
  `StepCapped`/inconclusive. It now asks once; a command that ran clears the
  guard whatever it reported; a command refused by the approval policy does not.
- **The guard was armed by `user_text.to_lowercase().contains("test")`,** which
  also matches "latest", "contest", "greatest", and never matches a non-English
  prompt. Removed. The guard is scoped to the benchmark profile, like the
  evidence gate directly above it, so the shipped interactive agent is unchanged.
- **`generate_near_miss_diagnostics` was unbounded** — it scored every window of
  the file with an O(a·b) Levenshtein per line pair, on files up to 8 MiB, on
  every failed edit. Both the window scan and the Levenshtein work are budgeted.

**Correctness**

- **`path_filter` `dir/**` silently matched nothing on Windows.** `Sandbox::rel`
  renders with the platform separator, so the comparison never matched. Patterns
  and paths are normalized to `/`. CI would have caught this on the Windows leg
  had clippy not failed first.
- `src/**` matched `src-generated/lib.rs` — a prefix test without the separator.
- An unsupported glob such as `src/*.rs` fell through to a `contains` test and
  returned zero hits **with a success status**. Now an explicit error.
- The compaction ledger read the result at `idx + 1` for every call in a batch,
  so one failure marked the whole batch. It reads the nth result now.
- The ledger listed every inspected file, inside the summary that exists to save
  context. Capped.
- `weather` is now required for the new forecast needles in both classifiers. A
  bare "forecast for" also matches "revenue forecast for next year", and a false
  positive here sends the user's prompt to a search engine.

## CUDA

Split out to **#711**, cut fresh from `main`. `src/cuda.rs` is restored
byte-for-byte here; `git diff main -- src/cuda.rs` is empty.

Worth knowing before reviewing it: the `__dp4a` → inline-PTX half is weaker than
its original comment claimed, and testing is what showed that. Reverting the
call site to `main`'s spelling and re-running the block-kernel parity test on the
RTX 4060 passes **either way** — NVRTC 12.9 resolves `__dp4a` on its own, so the
stated justification does not hold for the CUDA version this runs against. It is
kept as commit 2 of 2 over there, relabelled as portability insurance for an
older NVRTC, and drops cleanly if you'd rather not carry it. The lane diagnostic
is the half with a real symptom behind it.

## Tests

New regressions, each of which fails without its fix:

- `a_tab_indented_file_is_not_rewritten_by_a_shallower_tab_count`
- `an_indent_shift_that_would_flatten_the_replacement_is_refused`
- `a_failing_verification_command_can_still_be_reported`
- `the_verification_guard_asks_once_and_then_accepts_the_answer`
- `a_directory_path_filter_does_not_match_a_sibling_with_the_same_prefix`
- `an_unsupported_path_filter_is_refused_rather_than_matching_nothing`
- `dual_stack_answers_pin_only_the_reachable_family`
- three cases in the shared `classifier_cases.json`, including the negative
  "revenue forecast" case
- `benchmark_requires_file_evidence_before_mutation_or_answer` updated to record
  the step the guard now spends

## Verification

Run on the exact pushed tree (`f4d33988e`):

- `cargo fmt --all -- --check` → 0
- `cargo clippy --all-targets -- -D warnings` → 0
- `cargo test --lib` (the whole suite, not only the touched modules) →
  **2230 passed / 0 failed / 111 ignored**
- `node tools/bench/system/test-native-contracts.mjs` → 21/21
- `git diff --check` → 0
- frontend classifier driven directly against all 31 shared fixture cases → 0
  mismatches
- Live run on Windows: release binary, WebUI, real `/api/web/research` query
  returning two sources, and the negative "revenue forecast" prompt answering
  locally with no fetch.

## Honest gaps

- The premature-completion guard is deliberately scoped to the benchmark
  profile. Widening it to the interactive agent should be a separate, measured
  change.
- Tier 3 is now conservative in the tab case rather than correct in it: a
  tab-indented file with a space-indented `old` (or a differing tab count) is
  refused and sent to the near-miss diagnostic. That is the intended outcome for
  an uncertain match, but it does mean a class of edit that used to "succeed"
  now costs the model a retry.
- `host_reachable_families` probes with a UDP `connect`, which tests for a
  *route*, not for a working path to the internet. A host with a v6 route and a
  broken v6 upstream still gets v6 addresses pinned. Narrowing is best-effort
  and never empties the set, so this degrades to today's behaviour rather than
  to a failure.
- The live web-research evidence is one Windows host on one network.

### DIFF
--- a/frontend/src/lib/webResearch.js
+++ b/frontend/src/lib/webResearch.js
@@ -150,7 +150,9 @@ const CURRENT_INFO_PATTERNS = [
   /\b(?:news|weather|price|schedule|score)\s+(?:today|right now|now)\b/i,
   /\bweather\b[^.!?\n]{0,80}\b(?:today|tomorrow|now|tonight|week|weekend|days?|forecast)\b/i,
   /\bweather\s+(?:in|for|at|around|report|forecast)\b/i,
-  /\b(?:weather\s+)?forecast\s+(?:for|in|today|this|next)\b/i,
+  // `weather` is required: a bare "forecast for next quarter" is not a
+  // real-time question, and a false positive sends the prompt to a search engine.
+  /\bweather\s+forecast\s+(?:for|in|today|this|next)\b/i,
   /\brecent\s+(?:news|events|developments|changes|updates|releases)\b/i,
   /\bwho is (?:the )?current\b/i,
 ]
--- a/frontend/src/lib/webResearch.js
+++ b/frontend/src/lib/webResearch.js
@@ -148,6 +148,9 @@ const CURRENT_INFO_PATTERNS = [
   /\bcurrently available\b/i,
   /\b(?:today's|today’s)\s+(?:news|weather|price|schedule|score)\b/i,
   /\b(?:news|weather|price|schedule|score)\s+(?:today|right now|now)\b/i,
+  /\bweather\b[^.!?\n]{0,80}\b(?:today|tomorrow|now|tonight|week|weekend|days?|forecast)\b/i,
+  /\bweather\s+(?:in|for|at|around|report|forecast)\b/i,
+  /\b(?:weather\s+)?forecast\s+(?:for|in|today|this|next)\b/i,
   /\brecent\s+(?:news|events|developments|changes|updates|releases)\b/i,
   /\bwho is (?:the )?current\b/i,
 ]
--- a/qa/benchmarks/agent/native-contracts-v1.json
+++ b/qa/benchmarks/agent/native-contracts-v1.json
@@ -102,6 +102,37 @@
       "evidence": [
         { "kind": "js_test", "file": "tools/bench/system/test-native-adapter.mjs", "marker": "outside_task_root", "platforms": ["windows", "linux"] }
       ]
+    },
+    {
+      "id": "native_edit_self_healing",
+      "evidence": [
+        { "kind": "rust_test", "file": "src/chat/tools.rs", "test": "edit_file_tolerates_uniform_indentation_shift", "platforms": ["windows", "macos", "linux"] },
+        { "kind": "rust_test", "file": "src/chat/tools.rs", "test": "edit_file_provides_actionable_near_miss_diagnostics", "platforms": ["windows", "macos", "linux"] }
+      ]
+    },
+    {
+      "id": "native_path_suggestion",
+      "evidence": [
+        { "kind": "rust_test", "file": "src/chat/tools.rs", "test": "sandbox_resolve_suggests_nearest_file_path", "platforms": ["windows", "macos", "linux"] }
+      ]
+    },
+    {
+      "id": "native_compaction_working_set",
+      "evidence": [
+        { "kind": "rust_test", "file": "src/chat/agent.rs", "test": "compaction_records_active_working_set_ledger", "platforms": ["windows", "macos", "linux"] }
+      ]
+    },
+    {
+      "id": "native_search_path_filter",
+      "evidence": [
+        { "kind": "rust_test", "file": "src/chat/tools.rs", "test": "search_respects_path_filter", "platforms": ["windows", "macos", "linux"] }
+      ]
+    },
+    {
+      "id": "native_verification_gate",
+      "evidence": [
+        { "kind": "rust_test", "file": "src/chat/agent.rs", "test": "verification_gate_requires_test_run_after_mutating_files", "platforms": ["windows", "macos", "linux"] }
+      ]
     }
   ]
 }
--- a/src/api/web_research.rs
+++ b/src/api/web_research.rs
@@ -17,7 +17,7 @@
 use std::{
     collections::{HashMap, HashSet},
     io::{Read, Write},
-    net::{IpAddr, Ipv4Addr, Ipv6Addr, ToSocketAddrs},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, ToSocketAddrs, UdpSocket},
     process::{Command, Stdio},
     sync::{Arc, Mutex, OnceLock},
     thread,
@@ -3934,19 +3934,56 @@ fn validated_redirect_target(current: &Url, location: &str) -> Result<Url, Strin
     Ok(next)
 }
 
+/// Which IP families this host has a route for.
+///
+/// `to_socket_addrs` does not set `AI_ADDRCONFIG`, so a dual-stack DNS answer
+/// comes back whatever the host can actually reach. Deciding from the shape of
+/// that answer would only ever be right for one kind of host; this asks the host
+/// instead. A UDP `connect` sends no packet — it just needs a route to exist —
+/// so this costs a syscall, blocks on nothing, and is probed once per process.
+///
+/// Both flags false means the question could not be answered (no network, or a
+/// sandbox that refuses the bind), and the caller pins the answer unchanged.
+fn host_reachable_families() -> (bool, bool) {
+    static FAMILIES: OnceLock<(bool, bool)> = OnceLock::new();
+    *FAMILIES.get_or_init(|| {
+        // Documentation prefixes: only a default route can match them, which is
+        // exactly the property being tested.
+        let v4 = UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0))
+            .and_then(|socket| socket.connect((Ipv4Addr::new(192, 0, 2, 1), 9)))
+            .is_ok();
+        let v6 = UdpSocket::bind((Ipv6Addr::UNSPECIFIED, 0))
+            .and_then(|socket| socket.connect((Ipv6Addr::new(0x2001, 0xdb8, 0, 0, 0, 0, 0, 1), 9)))
+            .is_ok();
+        (v4, v6)
+    })
+}
+
 /// Choose which of the already-validated addresses to pin with `--resolve`.
 ///
 /// A dual-stack DNS answer on a single-stack host (a v4-only GCP Compute Engine
-/// VM, say) hands curl AAAA records that can never connect, and the hop dies
-/// with `couldn't connect to server` instead of using the A record beside it.
-/// When both families are present the IPv6 addresses are dropped. A host whose
-/// only answer is IPv6 keeps it — this narrows the address set, it never empties
-/// it, so the SSRF pin still binds every address curl may reach.
-fn addresses_to_pin(addresses: &[IpAddr]) -> Vec<IpAddr> {
-    if addresses.iter().any(IpAddr::is_ipv4) {
-        addresses.iter().copied().filter(IpAddr::is_ipv4).collect()
-    } else {
+/// VM, say) hands curl records that can never connect, and the hop can die with
+/// `couldn't connect to server` instead of using the usable record beside it.
+/// The families the host cannot reach are dropped — in either direction, so an
+/// IPv6-only host is not handed A records for the same reason.
+///
+/// This only ever narrows, and never to nothing: an unreachable pin is still a
+/// bound pin, whereas pinning nothing would hand the hostname back to curl's own
+/// resolver and unbind the SSRF check.
+fn addresses_to_pin(addresses: &[IpAddr], families: (bool, bool)) -> Vec<IpAddr> {
+    let (v4_reachable, v6_reachable) = families;
+    if v4_reachable == v6_reachable {
+        return addresses.to_vec();
+    }
+    let reachable: Vec<IpAddr> = addresses
+        .iter()
+        .copied()
+        .filter(|address| address.is_ipv4() == v4_reachable)
+        .collect();
+    if reachable.is_empty() {
         addresses.to_vec()
+    } else {
+        reachable
     }
 }
 
@@ -4032,11 +4069,16 @@ fn curl_single_hop(
         .host()
         .is_some_and(|host| matches!(host, Host::Domain(_)))
     {
-        let pinned = addresses_to_pin(addresses);
+        let pinned = addresses_to_pin(addresses, host_reachable_families());
         if pinned.len() < addresses.len() {
-            // Every address we still pin is IPv4; tell curl so too, so a cache
-            // miss cannot send it back to the resolver and out over IPv6.
-            command.arg("--ipv4");
+            // Every address still pinned is one family; tell curl so too, so a
+            // cache miss cannot send it back to the resolver and out over the
+            // family this host cannot reach.
+            command.arg(if pinned.iter().all(IpAddr::is_ipv4) {
+                "--ipv4"
+            } else {
+                "--ipv6"
+            });
         }
         for address in pinned {
             let pinned = match address {
@@ -4247,14 +4289,23 @@ mod tests {
         let v4: IpAddr = "93.184.216.34".parse().unwrap();
         let v4b: IpAddr = "93.184.216.35".parse().unwrap();
         let v6: IpAddr = "2606:2800:220:1:248:1893:25c8:1946".parse().unwrap();
+        let v4_only = (true, false);
+        let v6_only = (false, true);
+        let dual = (true, true);
+        let unknown = (false, false);
 
-        // Mixed answer: the IPv6 addresses are dropped so a v4-only host cannot
-        // spend the hop budget dialling an unreachable address.
-        assert_eq!(addresses_to_pin(&[v6, v4, v4b]), vec![v4, v4b]);
-        // IPv6-only answer is kept verbatim; narrowing must never empty the set,
-        // or the SSRF pin would stop binding and curl would resolve for itself.
-        assert_eq!(addresses_to_pin(&[v6]), vec![v6]);
-        assert_eq!(addresses_to_pin(&[v4]), vec![v4]);
+        // A v4-only host drops the AAAA records it could never dial.
+        assert_eq!(addresses_to_pin(&[v6, v4, v4b], v4_only), vec![v4, v4b]);
+        // And the mirror image: a v6-only host drops the A records.
+        assert_eq!(addresses_to_pin(&[v6, v4, v4b], v6_only), vec![v6]);
+        // Narrowing must never empty the set, or the SSRF pin would stop binding
+        // and curl would resolve for itself.
+        assert_eq!(addresses_to_pin(&[v6], v4_only), vec![v6]);
+        assert_eq!(addresses_to_pin(&[v4], v6_only), vec![v4]);
+        // A dual-stack host keeps both and lets curl choose; so does a host whose
+        // families could not be probed.
+        assert_eq!(addresses_to_pin(&[v6, v4], dual), vec![v6, v4]);
+        assert_eq!(addresses_to_pin(&[v6, v4], unknown), vec![v6, v4]);
     }
 
     #[test]
--- a/src/api/web_research.rs
+++ b/src/api/web_research.rs
@@ -1222,9 +1222,10 @@ fn classify_prompt(prompt: &str) -> ResearchDecision {
         "weather in ",
         "weather for ",
         "weather report",
+        // Deliberately weather-anchored. A bare "forecast for" also matches
+        // "revenue forecast for next year", and a false positive here sends the
+        // user's prompt to a search engine.
         "weather forecast",
-        "forecast for ",
-        "forecast in ",
         "price now",
         "schedule now",
         "score now",
@@ -3933,6 +3934,22 @@ fn validated_redirect_target(current: &Url, location: &str) -> Result<Url, Strin
     Ok(next)
 }
 
+/// Choose which of the already-validated addresses to pin with `--resolve`.
+///
+/// A dual-stack DNS answer on a single-stack host (a v4-only GCP Compute Engine
+/// VM, say) hands curl AAAA records that can never connect, and the hop dies
+/// with `couldn't connect to server` instead of using the A record beside it.
+/// When both families are present the IPv6 addresses are dropped. A host whose
+/// only answer is IPv6 keeps it — this narrows the address set, it never empties
+/// it, so the SSRF pin still binds every address curl may reach.
+fn addresses_to_pin(addresses: &[IpAddr]) -> Vec<IpAddr> {
+    if addresses.iter().any(IpAddr::is_ipv4) {
+        addresses.iter().copied().filter(IpAddr::is_ipv4).collect()
+    } else {
+        addresses.to_vec()
+    }
+}
+
 // This is the final SSRF/auth boundary before spawning curl. Keeping every
 // validated input explicit makes accidental credential or redirect widening
 // visible at each call site.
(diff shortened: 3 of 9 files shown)


karan68:
Review pass pushed as `54b4212b` + `651155bc`. Summary of what changed and why, since several were behaviour bugs rather than style. Full detail is in the updated description.

**Why CI was red:** exactly one thing — clippy `unnecessary_map_or` in `compact`. It failed all three rust legs *before tests ran*, which is why none of the below surfaced.

**Behaviour bugs**

1. **The verification guard could trap the agent.** `run_shell` returns `ToolOutcome::Err` for any non-zero exit, and only a non-error cleared the flag — so an edit followed by a legitimately *failing* test could never be reported. Every text step was rejected until `max_steps`, ending `StepCapped`/inconclusive. Now: asks once, a command that ran clears it whatever it reported, a policy-denied command does not.
2. **`user_text.contains("test")`** also matches "la**test**", "con**test**", "grea**test**", and never matches a non-English prompt. Removed; the guard is scoped to the benchmark profile like the evidence gate above it, so the shipped interactive agent is untouched.
3. **`path_filter` `dir/**` silently matched nothing on Windows.** `Sandbox::rel` renders with the platform separator. The Windows leg would have caught this had clippy not failed first.
4. **`src/**` matched `src-generated/`** — prefix test without the separator.
5. **Unsupported globs returned zero hits with a success status.** `src/*.rs` fell through to a `contains` test; to the model that reads as "the symbol is not there". Now an explicit error.
6. **`generate_near_miss_diagnostics` was unbounded** — O(a·b) Levenshtein per line pair over every window, on files up to 8 MiB, on every failed edit. Budgeted now.
7. **Compaction ledger** read `middle[idx + 1]` for every call in a batch, so one failure marked the whole batch; and it listed every inspected file inside the summary that exists to *save* context. Both fixed.
8. **`forecast for` is now weather-anchored.** It also matched "revenue forecast for next year" — and a false positive here sends the user's prompt to a search engine.

**Description accuracy**

The old body advertised "Levenshtein fuzzy block anchor search (85% similarity threshold)". No such thing exists on the mutation path — tier 3 is exact-trimmed-line matching with a uniform indent delta, and every tier requires a unique match. Levenshtein is used only for path suggestions and read-only diagnostics. Rewritten to match the code.

**Shared classifier fixture**

`tests/fixtures/websearch/classifier_cases.json` is read by both `src/api/web_research.rs` and `frontend/scripts/web-research-smoke.mjs`; it exists to stop the two implementations drifting. This PR changed both classifiers and added no cases. Added three, including the negative "revenue forecast" case. Both implementations pass all 31.

**CUDA scope**

`extern "C" { __global__ }` is identical to `extern "C" __global__`, so that commit was a no-op with a misleading message — reverted.

The `eprintln!` on backend init is **kept but scoped**. As written it fired on every init failure including the ordinary no-NVIDIA-driver host, undoing the deliberate quiet degradation. It now fires only after the device has been announced, i.e. a usable GPU exists but the lane could not start. Measured on a driver-only Windows host:

```
[cuda] selected device 0 of 1: "NVIDIA GeForce RTX 4060 Laptop GPU" (compute 8.9) | VRAM 7099 MiB free / 8187 MiB total
[cuda] device found but the GPU lane could not start (CUDA NVRTC library not available); running on the CPU path instead
```

Before this, that host silently ran every token on the CPU with nothing explaining why a healthy 4060 sat idle.

The `__dp4a` → inline-PTX change is kept and commented, and the stale `compile_options` comment naming the intrinsic is corrected. Flagging honestly: it is unrelated to this PR and **untested by CI** (no GPU on the runners). On the review machine NVRTC is absent, so the kernel source is never compiled — that change is neither exercised nor validated here. Happy to split it out if you'd prefer.

**Gates** (local, on `651155bc`): fmt; `clippy --lib --tests -D warnings`; `chat::tools` 68/0; `chat::agent` 90/0; `api::web_research` 61/0; `cuda::` 15/0 (5 device tests skipped); native contracts 21/21; frontend classifier 0 mismatches across all 31 shared cases; `git diff --check` clean.

karan68:
### macOS CI was red — root-caused and fixed (`15371ef6`)

The macOS Rust leg failed on `651155bc0`. It was **not** one of this PR's tests, and not a
compile or lint problem — format and clippy both passed on macOS. The result was
`2301 passed; 1 failed`, and the one failure was in a file this PR does not touch:

```
thread 'fabric::http::tests::a_tls_stream_cut_off_before_its_declared_end_is_not_a_clean_eof'
panicked at src/fabric/http.rs:1521:10:
TLS stream opens before the truncation: Io("Connection reset by peer (os error 54)")
```

**Root cause.** The reset happens while the stream is being *opened*, before the truncation the
test is actually asserting. `serve_tls_once` read only until `\r\n\r\n`, i.e. to the end of the
request head. The failing test POSTs a body, and the client writes head and body separately, so
those body bytes can still be sitting unread when the stub thread returns. Closing a socket that
holds unread bytes is an abortive close (RST) rather than a FIN, so the client observes
`ECONNRESET` instead of the response the stub had already written.

Whether the body lands in the same read as the head is pure timing, which is why the suite is
usually green — both TLS tests that POST a body are exposed to this, and only one lost the race
in that run.

**Fix.** `1afd0f030` ("drain the whole request in the canned-node test helper") already fixed
exactly this for the plaintext path and introduced `request_complete` for it. The TLS stub was
never moved over. This commit does that, so both stubs now wait for the head *and* the declared
body before answering. One line of behaviour plus a comment; test-only, no production path.

Happy to split this into its own PR if you would rather keep this one to the agent harness — it
is independent of everything else here. I left it in because it is what is standing between this
branch and a green `ci-gate`.

**Everything else on `651155bc0` was already green:** ubuntu (`--all-features`), windows,
frontend, desktop, validation-scripts, public-scrub, setup. macOS was the only red leg, and
`ci-gate` only failed because it aggregates it.

timtoole02 (changes requested):
Reviewed at `15371ef6`. I pulled the head, read the full diff, and lifted the new
`edit_file` tier logic into a standalone harness so I could exercise it directly
rather than reason about it. Two of the tolerant-match behaviours write a wrong
file under an `Ok` status, which is the one thing this PR promises the mutation
path will not do. Everything else is either sound or minor.

## Verified

Recording what I actually checked, so it doesn't get re-litigated:

- **The near-miss budgets hold.** Against a 9.9 MB / 200k-line file:
  `generate_near_miss_diagnostics` runs 15 ms (m=1) to 37 ms (m=10), then falls
  through to `NEAR_MISS_GENERIC` at m=11 when `file_lines * m` crosses
  `MAX_NEAR_MISS_WINDOW_COMPARISONS`. `find_tolerant_line_matches` is ~4 ms
  throughout. The "a failed edit on a large file cannot pin a core" claim is
  correct.
- **The compaction `idx + 1 + nth` fix is right.** Every path inside
  `for call in calls` in `run_loop` — the benchmark evidence gate, the
  `validate_for` error, and each of the four `Decision` arms — pushes exactly one
  `ToolResult`, and the single `ToolCalls` push precedes the batch. The
  adjacency the ledger assumes genuinely holds, and a compaction window that
  cuts a batch degrades to "not recorded" rather than to a wrong attribution.
- **Both classifiers are pinned by the same fixture in CI**, not only by the
  manual run in the description: `src/api/web_research.rs:4457` and the
  `smoke:web-research` step in the `frontend` job both read
  `classifier_cases.json`.
- The `executed` flag correctly separates "ran and reported a failure" from
  "refused by the approval policy". That was the right correction — the guard as
  originally written could have trapped the agent, and the two regressions
  covering it earn their place.

## Blocking

### 1. Tier 3 silently rewrites indentation in tab-indented files

`src/chat/tools.rs:2625-2626`

Indentation is measured as `take_while(|c| *c == ' ')`, so two tab-indented
lines that differ by tab *count* both measure zero. `delta` computes to 0, the
trimmed lines compare equal, the window matches, and `adjust_indentation` writes
`new` verbatim — carrying the model's tab count into the file in place of the
file's own:

```
file: "def f():\n\t\tif x:\n\t\t\treturn 1\n"
old:  "\tif x:\n\t\treturn 1"          (one tab shallower than the file)
new:  "\tif x:\n\t\treturn 2"

result: "def f():\n\tif x:\n\t\treturn 2\n"     <- the block lost an indent level
status: edited <path> (matched lines 2-3 after adjusting indentation/whitespace)
```

In Python, a Makefile, or YAML that is a semantic change reported as a success,
and the model has no way to detect it — the result line says indentation was
adjusted, which is exactly what it asked for.

The "honest gaps" section covers a space-indented `old` against a tab-indented
file. This is the same root cause but the likelier shape: a model that has read
the file will send tabs, and the thing it gets wrong is the count.

Suggested fix, in `find_tolerant_line_matches` rather than in
`adjust_indentation` — reject the window unless the two leading-whitespace runs
are either both space-only or byte-identical. That keeps the tier honest
(space-indented files, the case it was built for) and sends everything else to
the near-miss diagnostic, which is the correct outcome for an uncertain match.

### 2. A negative `indent_delta` flattens relative indentation inside `new`

`src/chat/tools.rs:2557-2558`

`to_remove = leading_spaces.min(trim_count)` clamps per line, so any line in
`new` with fewer leading spaces than the shift lands at column 0 while its
siblings shift uniformly:

```
file:  "if x:\n  a = 1\n  b = 2\n"
old:   "        a = 1\n        b = 2"          (delta -6)
new:   "        a = 1\n    if y:\n        b = 2"

result: "if x:\n  a = 1\nif y:\n  b = 2\n"      <- `if y:` flattened to column 0
```

Same class as #1: an uncertain transform applied and reported green. The clamp is
the tell — a line that cannot absorb the shift is evidence the uniform-delta
assumption does not hold for this replacement.

Suggested fix: before applying, check that every non-blank line of `new` has at
least `trim_count` leading spaces; if not, fall through to
`generate_near_miss_diagnostics` and let the model re-read. That costs a retry
in a rare case and removes a silent corruption.

## Worth settling before merge

### 3. `addresses_to_pin` hard-codes "IPv4 is the reachable family"

`src/api/web_research.rs:3945`

This is the mirror image of the bug it fixes. Rust's `to_socket_addrs` does not
set `AI_ADDRCONFIG` — which is precisely why a v4-only host was handed AAAA
records in the first place — so an IPv6-only glibc host is handed A records just
as unconditionally. Every hop is then pinned to addresses it cannot reach, and
the added `--ipv4` removes curl's own fallback, so that host fails on every
request rather than intermittently.

To be clear about the security half: this is not a regression there. The set is
only ever narrowed, an IPv6-only answer is kept, and the test covering that is
the right test to have written.

The options I see are to keep both families and shorten the happy-eyeballs
window instead of dropping one, or to decide from the host's own interface
addresses rather than from the shape of the DNS answer. Either is a real fix for
both directions; the current one trades a v4-only failure for a v6-only failure.
Worth knowing which hosts you actually care about before picking.

## Minor

- `src/chat/tools.rs:2155` — when `search`'s root is a single file and the filter
  excludes it, the result is `Ok("no matches for ...")`. That is the same
  silently-empty-result-set failure the PR argues against for unsupported
  filters, and the argument applies here too: say the filter excluded the path.
- The near-miss Levenshtein budget is a running total consumed top-down, so on a
  large file only the first ~250 windows get fuzzy scoring (4M cells over roughly
  16k per window at 40-char lines). The free exact/`trim`/`trim_end` tiers still
  cover the whole file so it degrades gracefully, but "closest match" is biased
  toward the top of the file. Fine as-is; worth a comment at the constant.
- The compaction ledger records `call.args["path"]` as the model typed it, while
  `run_loop`'s `mutated_files` records `normalize_workspace_path(sandbox.rel(..))`.
  Two renderings of the same path in two summaries the model reads.

## Scope

`src/fabric/http.rs` is a ninth file, not listed under "Key changes" (the
description says 8 files, +1329/-59; GitHub reports 9, +1337/-60). The change
itself is a good catch and the reasoning about the abortive close is right — it
just arrived undeclared.

The CUDA work is still bundled. You have offered twice to split it, and now that
the RTX 4060 evidence is attached, the evidence survives the split — it belongs
on the PR that carries the kernel change. Splitting would let this one be
reviewed as one thing.

---

Findings 1 and 2 are both narrow, local fixes. With those in, this is a good
change: the tier ordering is sound, the uniqueness requirement at every tier is
the right call, and the diagnostics are a genuine improvement over
`` `old` text not found in file ``.

karan68:
Both blocking findings were real. Fixed in `9ce7f13d`, along with the three minor
ones and the scope point. Thank you for lifting the tier logic into a harness —
finding 2 in particular is not something the tests I wrote would ever have caught.

## 1. Tab-indented files — fixed

Fixed in `find_tolerant_line_matches`, as you suggested, rather than in
`adjust_indentation`. A window is rejected unless the two leading-whitespace runs
are byte-identical or both space-only:

```rust
let comparable = fl_ws == ol_ws
    || (fl_ws.bytes().all(|b| b == b' ') && ol_ws.bytes().all(|b| b == b' '));
```

The delta then falls out of the run lengths — zero for a byte-identical run, a
real column count for space-only runs — so both comparable cases keep working and
everything else goes to the near-miss diagnostic.

Your reproduction is now `a_tab_indented_file_is_not_rewritten_by_a_shallower_tab_count`:
the edit is refused and the file is byte-identical afterwards.

## 2. Negative delta flattening — fixed

Also before applying, not inside the clamp:

```rust
fn indent_shift_applies(new_text: &str, indent_delta: isize) -> bool
```

Every non-blank line of `new` must carry at least `trim_count` leading spaces,
otherwise tier 3 is skipped and the model gets the diagnostic. The clamp in
`adjust_indentation` is now a floor rather than a fallback, and its doc says so.

`an_indent_shift_that_would_flatten_the_replacement_is_refused` pins your
reproduction and also asserts that a negative shift every line *can* absorb still
applies, so the tier did not quietly become positive-only.

## 3. `addresses_to_pin` — changed direction rather than mirrored

You were right that it was the mirror image of the bug it fixed, and I did not
want to trade a v4-only failure for a v6-only one. It now asks the host instead of
the DNS answer:

```rust
fn host_reachable_families() -> (bool, bool)
```

A UDP `connect` sends no packet — it only needs a route to exist — so this answers
the `AI_ADDRCONFIG` question directly for one syscall, probed once per process.
Both-or-neither means do not narrow, so a dual-stack host keeps both families and
a host where the probe cannot answer (no network, restrictive sandbox) pins the
answer unchanged. `--ipv4` became `--ipv4`/`--ipv6` to match.

`addresses_to_pin` takes the families as an argument so it stays pure and
testable; the test now covers v4-only, v6-only, dual-stack, unprobeable, and both
never-empty cases.

## Minor

- **`search` on a single file the filter excludes** now says so. I made it an
  error rather than a clearer `Ok`, on the same reasoning the unsupported-filter
  refusal uses: that request cannot ever match, so it is a request to fix rather
  than a result to report. Say the word if you would rather it stayed `Ok`.
- **Near-miss budget bias** — noted at the constant, including why it degrades the
  ranking rather than the correctness of a reported match.
- **Ledger path rendering** — `summarize_tool_call` now runs the path through
  `normalize_workspace_path`, the same rendering `mutated_files` uses.

## One more thing the full suite caught

Running the whole lib suite rather than the touched modules surfaced
`chat::checkpoint::tests::write_and_edit_through_the_tool_path_are_checkpointed`
failing: it clears the process-wide checkpoint log and asserts an exact count of
2, but no agent test takes `cp_lock` or disables checkpoints, so any concurrent
mutation breaks it. The race is pre-existing — adding two more mutating tests just
made it reproduce.

The benchmark-profile agent tests now mirror what production already does on that
path (`with_checkpoints(benchmark_events.is_none())`) and leave the log alone.
It passes in isolation either way; the full suite is what shows it.

## Scope

- **fabric** is declared now, in Key changes, with the reasoning and the offer to
  split. Counts in the description corrected to 9 files, +1492/-62 — you were
  right that they were stale.
- **CUDA** — I have asked the author to confirm before I move it; I would rather
  not restructure someone else's commits unilaterally. If you would prefer it out
  regardless, say so and I will split it with the RTX 4060 evidence attached.

Gates on `9ce7f13d`: 2230 lib tests pass / 0 failed, rustfmt clean, clippy
`--lib --tests -D warnings` clean.

karan68:
Split done — the CUDA work is now [#711](https://github.com/timtoole02/Camelid/pull/711), cut fresh from `main` (`49ac4126`), and `src/cuda.rs` is restored byte-for-byte here (`git diff main -- src/cuda.rs` is empty). This PR is now the **8 files** it claims to be.

Two things worth knowing before you look at #711.

**The `__dp4a` change is weaker than its comment claimed, and I only found that out by testing it.** I reverted the call site to `main`'s spelling and ran the block-kernel parity test on the RTX 4060 both ways:

| spelling | result |
|---|---|
| `camelid_dp4a` (inline PTX) | 1 passed |
| `__dp4a` (main) | **1 passed** |

NVRTC 12.9 resolves `__dp4a` on its own, so the justification in the diff — "NVRTC compiles this source without the CUDA headers that declare `__dp4a`" — is not true for the CUDA version this actually runs against. I kept the change but rewrote the comment to say it is portability insurance for an older NVRTC, not a fix for an observed failure. It is commit 2 of 2 over there and drops cleanly if you'd rather not carry it.

**The lane diagnostic is the half with a real symptom behind it.** A host with a working NVIDIA device but no NVRTC compiles nothing and degrades to the CPU silently — I hit exactly that here. It is scoped so it can only fire *after* the device announcement, so driver-less machines stay quiet as they do today.

## Re-gated after the revert, on the exact pushed tree

| gate | result |
|---|---|
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy --all-targets -- -D warnings` | 0 |
| `cargo test --lib` (whole suite, not just the touched modules) | **2230 passed / 0 failed / 111 ignored** |
| `git diff --check` | 0 |
| native contracts | 21 / 21 |

New head is `f4d33988e`, pushed as a fast-forward from `9ce7f13d1`. I'll confirm CI green on it before asking for another look.

PR #717, opened 2026-09-04
fix(metal): take the attention K/V element type from the caller, not process globals

timtoole02:
## The bug

`encode_attention` chose between the `device const float*`, `device const half*` and
Q8-block attention kernels by reading process-global state — `kvq8_enabled()`,
`kv16_enabled()`, and inferring "f16 primary" from `kv16_mirrors.is_none()`.

That last inference is sound for `ResidentDecodeState`, which allocates mirrors exactly
when the primary is f32. It is **local to that lane**. `encode_attention` is shared, and
`encode_gemma4_attention`, `encode_qwen35_full_layer`, `encode_lfm2_attn_layer` and
`Lfm2MetalDecode::forward_prefill_chunk` all own unconditionally-f32 KV caches
(`... * head_dim * 4`; gemma4 sizes `kv_bytes` with `size_of::<f32>()`) and pass `None`
for mirrors. They took the F16 arm and bound f32 bytes to a half-typed kernel.

**LFM2 is live on the shipped CLI with no debug env.** `apply_default_fast_stack`
(`src/main.rs:8749`) sets `CAMELID_METAL_ATTN2=1`, `splitk_attention_enabled()` defaults
ON, and `splitk_kv_format_admits(false, _) = !false` admits the f32 lane.
LFM2.5-2.6B-Q8_0 is head_dim 64 (embedding_length 2048 / head_count 32) with group
32/8 = 4 — the top of the admitted `(1..=4)` range. Every gate opens at
`position_count >= 128`.

The blast radius differs per lane, and the difference matters:

| lane | head_dim | exposure |
|---|---|---|
| **LFM2** | 64 | **live today**, single model, default flags, >=128 positions |
| qwen35 (ornith-9B) | 256 | fails the `head_dim <= 128` v2 gate, so never split-K. Corrupts only via the non-split `match` after an **in-process model switch** from a K-quant Llama flips the global |
| gemma4 | 256/512 | same switch-dependent path, plus the panic below |

`set_resident_kquant_lane` is called only from the Llama resident lane, so a
single-model qwen35 process keeps `kv16_enabled() == false` and took the correct f32
kernel — which is why the existing ornith-9B agent-eval PASS is real and is unaffected
by this change.

## Evidence

LFM2.5-2.6B-Q8_0, Apple M4 / 16 GiB, `serve` + `/v1/chat/completions`, temperature 0,
identical prompt, release builds either side of this change:

| binary | prompt tokens | result |
|---|---|---|
| before | 20 | coherent output |
| **before** | **221** | **panics** |
| after | 221 | coherent output |
| after, `CAMELID_METAL_ATTN_SPLITK=0` | 221 | byte-identical to the split-K run |

The pre-fix panic is `range start index 9345848833920 out of range for slice of length
278528000` at `src/runnable/model.rs:365` (`dequant_row`). That index divides exactly:
`9345848833920 / 2176 row-bytes = 4294967295 = u32::MAX`. Reinterpreting f32 bytes as
f16 pairs lands on f16 exponent-all-ones often enough to produce NaN logits; argmax then
returned its sentinel, and the sentinel became a row index.

So the failure mode was not subtle drift. Below the threshold it is invisible; above it,
the forward pass is poisoned.

## Why it was never caught

Token-identity parity cannot see this, and the fixtures cannot reach it.
`tests/fixtures/lfm2_parity/lfm2.json` holds four fixtures of 4-7 prompt tokens with
`max_new: 24` — about 31 positions against a 128-position threshold. **The LFM2 parity
suite is structurally incapable of exercising the path it is meant to certify.**

This is the same blind spot as #714 on CUDA, where a mis-derived group id moved a logit
by 8.6 while greedy output still looked comparable to f16.

## The fix

A `KvOperand` witness supplied by the buffer's owner, replacing all three process-global
reads, plus the operand choice extracted into the pure `splitk_kv_source` so routing has
a GPU-free gate the way `splitk_kv_format_admits` already did.

Two further defects fall out of the same change:

* `unreachable!("Q8 KV requires the v2 geometry")` was reachable, and panicked, on any
  head_dim > 128 lane under `CAMELID_METAL_KV_DTYPE=q8`. With a witness it is unreachable
  by construction: a Q8 witness can only come from `ResidentDecodeState::new`, which
  already refuses `kvq8` unless `head_dim.is_multiple_of(32) && head_dim <= 128` — exactly
  the `v2` predicate.
* `encode_attention_tree` gated split-K on `!kv16_enabled()`, a global that session does
  not own, then `expect`ed the mirrors that gate does not guarantee. It now gates on the
  witness plus `kv16_mirrors.is_some()`, matching its own "mirror-only" comment and
  keeping tree verify in lockstep with the linear path whose byte-identity is the verify
  lane's correctness anchor.

## Tests

`splitk_kv_source_follows_the_caller_witness_not_the_absence_of_mirrors` — verified in
**both** directions, not asserted once. Against the pre-fix logic it fails with
`left: PrimaryF16, right: PrimaryF32`, the mis-typed operand itself.

`metal_attention_decode_splitk_f32_matches_cpu_reference_across_the_threshold` — the f32
split-K kernel had **no correctness test**, and this change sends it real traffic for the
first time from those lanes. Sweeps (8,4,64,128), (8,4,64,200), (6,2,64,131),
(6,2,96,131), (6,2,128,131): head_dim 64 is LFM2's width, 96 gives `dpl = 3` so lane dim 3
is inactive (the under-subscribed shape that made the CUDA sibling race), and positions
cross the admission threshold with chunk tails that do not divide `PT = 8`. All five match
the CPU f32 reference within 1e-4 — so the kernel the corrected routing selects is sound,
and the defect was purely the operand choice.

## Gates

Run on the mini2 M4 rig, cargo/rustc 1.94.1:

```
cargo fmt --all -- --check           clean
cargo clippy --all-targets -- -D warnings
                                     clean for this change; one PRE-EXISTING
                                     `unnecessary_cast` at src/metal.rs:47057 from
                                     ee76a22f fires on this toolchain
cargo test --lib                     2341 passed
cargo test --all-targets             2 failures, both `fabric::http` TLS loopback tests
                                     (`Connect("[::1]:PORT: Connection refused")`).
                                     Intermittent on this host and reproduce identically
                                     on pristine origin/main with these changes removed.
```

## Not fixed here

`attention_decode_splitk_kv16_direct{,_tree}` hardcode the partials row stride at
`(128 + 2)` while the host sizes `partials` by `(head_dim + 2)`. That is #714's exact
shape. It is held together only by three separate `if head_dim == 128` call-site guards
(`src/metal.rs:18242`, `:18256`, and the tree twin); one edit to any of them makes it
live. Pre-existing and independent of this fix, so it belongs in its own PR — flagged
here so it is not lost.

### DIFF
--- a/src/metal.rs
+++ b/src/metal.rs
@@ -15435,6 +15435,9 @@ fn encode_gemma4_attention(
         &qn_buf,
         cache_k_buf,
         cache_v_buf,
+        // gemma4's KV cache is unconditionally f32 (`kv_bytes` is sized with
+        // `size_of::<f32>()`), and it stages no f16 mirrors.
+        KvOperand::F32,
         None,
         &scores_buf,
         &ctx_buf,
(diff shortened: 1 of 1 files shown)


timtoole02:
Follow-up: #718 stacks on this branch, hardening the one remaining `head_dim == 128` assumption in the same split-K arms (`attention_decode_splitk_kv16_direct` writes partials at a hardcoded `(128 + 2)` stride while the host sizes them by `head_dim + 2`). It is based on this branch and should be retargeted to `main` once this merges.

timtoole02:
Closing: **the core fix landed independently on main**, and merging this as written would now regress it.

## What superseded it

`encode_attention` on main already takes the primary KV format **from the caller**, as a `kv16: bool, kvq8: bool` pair rather than this PR's `KvOperand`, with the same rationale this PR argues:

> Primary KV format of the BOUND `keys`/`values` buffers, threaded from the caller (an engine's latched `self.kv16`/`self.kvq8`, or `false, false` for the family paths that always keep an f32 cache). Never re-read the process-global gates here: two resident engines with different formats coexist in one process … a mismatch makes these kernels read the cache in the wrong element width and decode garbage.

And all three lanes this PR names now pass `false, false`: `encode_lfm2_attn_layer`, `encode_qwen35_full_layer`, `encode_gemma4_attention`. So **the live LFM2 panic at 221 prompt tokens is fixed**, and `b9dade03 keep standalone KV caches f32` is covered too.

## Why not merge it anyway

I rebased it onto main to find out (253 commits, 11 conflict hunks in `src/metal.rs`), and two resolutions would have *undone* newer work:

- main added **kv16-primary support to tree split-K** — binding `keys`/`values` directly when the primary is already half, with a `TREE_KV16_SPLITK_ENCODES` counter. This PR makes that path mirror-only and would remove it.
- main added a **kv16 tree pipeline** (`attention_decode_v2_kv16_tree_pipeline`). This PR's `debug_assert_eq!(kv, KvOperand::F32)` — "tree attention has no compressed-primary kernel" — is no longer true and would fire.

Reworking it against main would mean rewriting most of the diff around a fix that already exists.

## Worth salvaging separately

Two things here are **absent from main** and still valuable, independent of the routing change:

- `metal_attention_decode_splitk_f32_matches_cpu_reference_across_the_threshold` — the f32 split-K decode kernel still has no correctness test of its own, and the sweep (head_dim 64/96/128, positions across the admission threshold, chunk tails that do not divide `PT = 8`) is the right shape.
- `splitk_kv_source` as a pure, GPU-free predicate plus `splitk_kv_source_follows_the_caller_witness_not_the_absence_of_mirrors`, which is verified in **both** directions rather than asserted once.

The `KvOperand` enum is also a genuine improvement over the `(bool, bool)` pair — it makes the invalid F16-and-Q8 state unrepresentable — but that is a refactor to land on its own terms, not a bug fix.

## #718

Its hazard is **still live on main** and is being reopened against `main` separately. Its own warning aged well: it described three call sites and said "a fourth call site was one edit away" — main now has four, a `_batch` pair having been added since. Its commit `b376c0be` touches `KvOperand` zero times, so it never actually depended on this PR.

PR #724, opened 2026-09-06
feat(frontend): token inspector — per-token probabilities on an inspected reply

timtoole02:
Camelid's engine has reported per-token logprobs since #334, advertised on `/api/capabilities` as the `rich_logprobs` row. Until now no frontend file mentioned them. This adds the surface.

A **Tokens** toggle in the composer. Flip it, send, and the reply carries a collapsible panel showing the model's per-token distribution:

```
Position 1 — emitted  7  at 58.5% (−0.54 log)
  1  7                       58.5%
  2  8                       17.4%
  3  6                        7.8%
  4  9                        6.7%
  5  5                        4.0%
  ·  rest of the vocabulary    5.6%

TOKENS 3 · ALTERNATIVES PER POSITION 5 · MEAN LOG-PROBABILITY −0.31 · PERPLEXITY 1.36
```

## The decision everything follows from: capture, don't re-run

The scores come from the reply's own decode, not a replay. The obvious alternative — an "Inspect" button on a finished message — would decode a **second** time and report numbers describing a different generation. On an experimental row (which samples at 0.7, `useDashboardData.js:1603`) it would not even reproduce the text.

That is why the toggle is pre-send: the cost of the panel being *unable* to misstate provenance, rather than merely disclosing that it might. It also means the request composes with the non-streaming branch `receiptMode` already uses, inheriting its 400-avoidance instead of restating it.

## Engine behaviour, verified live

Everything below was measured against a loaded `Llama-3.2-1B-Instruct-Q8_0` on the Windows CUDA-resident lane (`cuda_resident_q8_runtime`), not read off the source.

| Behaviour | Result |
|---|---|
| logprobs on the CUDA-resident lane | **Works** — matches the committed macOS Metal receipt to ~0.007 nats, same ranking |
| `logprobs` + `stream:true` | typed 400 |
| `logprobs` + `n>1` | typed 400 — mutually exclusive |
| `top_logprobs` without `logprobs` | typed 400 |
| `n=3` at `temperature:0` | **three byte-identical choices**; at 0.9 they differ |
| Payload size | **414 B/token** at depth 5, **1456 B/token** at depth 20 (~100× the reply text) |

The CUDA-resident result is new — the repo had a logprobs receipt for macOS Metal only.

## What ships guarded

The side-by-side candidates mode is **guarded, not live**, for two independent reasons: the engine refuses `logprobs` with `n>1`, and at `temperature:0` — where every parity-locked row sits — repeated candidates are byte-identical. The panel says both in typed copy. This is the `samplingContract.js` pattern, and the surface becomes real work once sampling rows exist.

## Storage

Records are held in session state keyed by message id, **never on the message object**. At 414 B/token the alternative is a data-loss bug: `persistConversations` swallows quota errors, so an overflow would silently stop saving the entire conversation rather than just the scores. Asserted structurally in the smoke, not left to a comment.

## Fails open, and the panel knows it

The gemma4, runnable (qwen35/Ornith, lfm2, gemma2, bitnet, command-r) and DiffusionGemma lanes short-circuit at `src/api/mod.rs:18011/18063/18077` **before** logprobs validation, so `logprobs:true` returns 200 with the key simply absent. Routing is per HOST, not per architecture: a Q8_0 gemma3 takes the runnable bridge on Windows/CUDA but the Metal resident lane on macOS, where it DOES report logprobs (measured — see the macOS verification comment). The panel reports that as an unmeasured position — never as a flat or certain distribution — and distinguishes it from a proxy forcing SSE. Filed as an engine defect in `BACKEND_ASKS.md` §6; this PR does not attempt the Rust fix.

## Encoding

No new colour tokens. Copper and amber stay reserved for claim colours, and red stays for errors — a low-probability token is a normal state. Probability rides an **underline weight** rather than text colour, so the audited body-text contrast pair is untouched and the encoding survives greyscale; the exact value is always present as text. The salience budget is inverted on purpose: settled tokens are unmarked, so ink goes to the contested minority.

## Drive-by fixes, found by a new gate

A generic tab-registry assertion added here caught real drift, none of it visible in a screenshot:

- `arena` missing from `TopBar` `TITLES` → header rendered "Camelid"
- `arena` and `downloads` missing from `CommandPalette` `VIEW_LABELS` → unreachable from Cmd+K
- `observatory` missing from `VALID_TABS` → tab dropped on reload

Fixed here because the gate cannot ship red. Also wires `npm run smoke:contrast` into CI — it has existed since the design-token work and **never ran there**, so palette changes have been shipping unaudited.

## Verification

- `smoke:token-inspector` — 33 checks, new, wired into the `frontend` job
- `smoke:contrast` — 258 token/fill pairs pass WCAG AA in both themes
- `smoke:ui`, `smoke:streaming`, `smoke:integration`, `smoke:workspace`, `smoke:model-state`, `smoke:model-lanes`, `smoke:first-run`, `smoke:first-run-card`, `smoke:capability-readiness`, `smoke:arena` — all exit 0
- `check-public-scrub.sh` — clean
- Driven end-to-end in the browser in both themes; storage claim confirmed live (5 inspections captured, 0 logprobs keys in `localStorage`)
- Re-verified on macOS/Metal (M4) against a release build of this head — engine contract, panel, both themes, storage claim, and the fails-open lane all reproduce; full record in the comment below

The smoke normalizes against the repo's own committed wire capture (`qa/capability/mac_logprobs_out/llama-3.2-1b/chat_logprobs.json`) rather than a hand-written fixture, so it fails if the engine's shape drifts. It also asserts that a `=== null` presence test *would* misread the key-absent body, that no rendered state uses certainty vocabulary, and that the vendor name in the `rich_logprobs` notes never reaches the screen.

## Not included

- Streaming logprobs — filed as `BACKEND_ASKS.md` §5. Would collapse the pre-send decision into an ordinary disclosure.
- The Rust fails-open fix — reported in §6, left to the engine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -559,6 +559,18 @@ jobs:
         if: ${{ !cancelled() && steps.build.outcome == 'success' }}
         run: npm run smoke:ui
 
+      - name: Token inspector smoke
+        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+        run: npm run smoke:token-inspector
+
+      # Wired here because the token inspector adds the first value-encoding
+      # colors to the UI. The script has existed since the design-token work but
+      # was never added to this job, so every palette change since then shipped
+      # unaudited — the same gap this file already records for other lanes.
+      - name: Design token contrast audit
+        if: ${{ !cancelled() }}
+        run: npm run smoke:contrast
+
   # ---------------------------------------------------------------------------
   # Single aggregate gate. Make THIS the only required status check in branch
   # protection (see the note in the PR/commit). It runs on every event and
--- a/frontend/design-evidence/BACKEND_ASKS.md
+++ b/frontend/design-evidence/BACKEND_ASKS.md
@@ -96,3 +96,46 @@ frontend cannot — and will not fake — a chat session against the runnable la
 
 Until then the Compatible rows stay receipt-only with an explicit "CLI only — no HTTP serve
 yet" note; membership and evidence remain fully derived, nothing is invented.
+
+## 5. Logprobs on the streaming path (Token Inspector, 2026-09-06)
+
+**Surface waiting on it:** the Token Inspector composer toggle
+(`lib/tokenInspection.js`, `components/chat/render/TokenInspector.jsx`). Because
+`rich_logprobs.unsupported_modes` contains `streaming`, an inspected turn must be
+sent with `stream: false`, so the user gives up token-by-token rendering for that
+reply and must decide to inspect BEFORE sending. Inspecting afterwards is not an
+option we are willing to take: it would mean decoding a second time, and those
+numbers would describe a different generation than the one on screen.
+
+**Ask:** carry the per-token record on the SSE path — either a `logprobs` object on
+each `choices[0].delta` (one entry per emitted token), or one terminal chunk
+carrying the whole `content[]` array in the same shape the non-streaming response
+already uses. Either shape collapses the pre-send decision into an ordinary
+disclosure of data already in hand, and the frontend keeps streaming.
+
+**Not asked:** logprobs together with `n > 1`. Those are mutually exclusive by
+contract and the UI treats the candidates surface as guarded, not pending.
+
+## 6. Logprobs fail OPEN on four serve lanes (engine defect, 2026-09-06)
+
+**Not a frontend ask — a backend correctness report.**
+
+`chat_completions` short-circuits into the gemma4 (`src/api/mod.rs:18011`), runnable
+(`:18063`) and DiffusionGemma (`:18077`) lanes BEFORE `validate_choice_and_logprob_fields`
+runs. On those lanes a request carrying `logprobs: true` returns **HTTP 200 with the
+`logprobs` key simply absent** — no error, no typed refusal. Every other invalid
+logprobs combination on the dense lane is a typed 400, so this one path fails open
+while its neighbours fail closed.
+
+That matters because the missing key is indistinguishable, to a client, from a lane
+that genuinely reported nothing — and the honest reading of "no distribution" is
+easy to confuse with "a flat or certain distribution". The Token Inspector handles
+it defensively (`inspectionAbsenceReason`, `code: 'lane_absent'`) and says the
+measurement is missing rather than flat, but the engine should refuse rather than
+leave every client to detect this independently.
+
+**Ask:** move the logprobs validation ahead of lane dispatch, and return the same
+typed 400 these lanes already return for other unsupported combinations — or add a
+lane axis to `api_conformance` so the refusal is at least machine-readable. Note
+that `unsupported_modes` today covers only the ROUTE axis (`streaming`,
+`multi_choice`) and says nothing about which serve lanes can honour the request.
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -21,6 +21,7 @@
     "smoke:catalog-companion": "node scripts/catalog-companion-smoke.mjs",
     "smoke:model-deletion": "node scripts/model-deletion-smoke.mjs",
     "smoke:3b-closure": "node scripts/frontend-3b-closure-smoke.mjs",
+    "smoke:token-inspector": "node scripts/token-inspector-smoke.mjs",
     "smoke:ui": "node scripts/ui-regression-smoke.mjs",
     "smoke:ghost-moe-execution": "node scripts/ghost-moe-execution-plan-smoke.mjs",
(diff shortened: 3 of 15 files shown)


timtoole02:
## macOS verification — Metal, live

Re-ran this on an M4 Mac against a release build of the PR head (`a8168a5d`), because everything in the description was measured on the Windows CUDA-resident lane. Every claim reproduces on Metal, and one of them is narrower than the description says.

### Engine contract, `Llama-3.2-1B-Instruct-Q8_0` on `metal_resident`

| Behaviour | Windows CUDA (as filed) | macOS Metal (measured here) |
|---|---|---|
| `rich_logprobs` row in `/api/capabilities` | supported | `supported_current_gate_nonstreaming`, `supported_modes: [chat_nonstreaming, completions_nonstreaming]` — so the toggle resolves **live**, not guarded |
| logprobs, non-streaming | works | works — `'Red'` −0.3584, `'Blue'` −1.2964, `'Green'` −4.3737 |
| `logprobs` + `stream:true` | typed 400 | typed 400 |
| `logprobs` + `n>1` | typed 400 | typed 400 |
| `top_logprobs` without `logprobs` | typed 400 | typed 400 |
| `top_logprobs: 21` / `: 20` | — | 400 / 200 — the cap mirrored as `MAX_TOP_LOGPROBS` is exact |
| no logprobs requested | — | key **absent**, not null — which is what makes the `hasOwnProperty` presence test in `chatCompletionStream.js` the correct guard rather than a stylistic one |

### The description overstates the fails-open set on macOS

The body lists `gemma3` among the runnable lanes that answer 200 with the key absent. That is true on Windows and CUDA hosts, and **false on macOS for Q8_0**: `arch_requires_runnable_bridge_given` (`src/model.rs:244`) sends a Q8_0 gemma3 to the Metal resident lane on a resident-capable host, so it takes the dense path and the logprobs handler runs.

Measured, same build: `gemma-3-4b-it-Q8_0` → **HTTP 200 with logprobs present**, first position `'Red'` at −1.42e-05.

`BACKEND_ASKS.md` §6 is unaffected — it is worded by *lane* (`gemma4` / runnable / DiffusionGemma), not by architecture, so it stays accurate on both hosts. Only the PR description's parenthetical was host-specific; corrected above.

The fails-open path itself does reproduce here. `LFM2.5-1.2B-Instruct-Q8_0` is runnable-only on every host: HTTP 200, `logprobs` key absent, and the panel renders

> **THIS EXECUTION LANE DID NOT REPORT PROBABILITIES** — The reply completed normally and the engine returned no probability record for it. […] This says nothing about how the model ranked these tokens — the measurement is missing, not flat.

which is the honest reading, and it offers no control that would fire a second generation.

### Panel, driven in the browser

Toggle reads "Tokens on"; the reply carries the collapsible panel with the token strip, the ranked alternatives plus *rest of the vocabulary*, the TOKENS / ALTERNATIVES PER POSITION / MEAN LOG-PROBABILITY / PERPLEXITY summary, the guarded side-by-side block naming both blockers, and Download JSON.

On a 19-token reply the trigger read `19 tokens · 1 closely contested` and the panel opened on position 17 — emitted `·horizon` at 26.7% (−1.32 log) against `·canvas` at 18.6%, `·that` at 11.5%, and 35.2% spread over the rest of the vocabulary. That is the salience-budget argument in the description working as designed: the eighteen settled positions are unmarked and the one contested position is where the ink went.

Renders correctly in both themes. No console errors.

**Storage claim confirmed live on this host:** after two inspections, `localStorage` held 8 keys totalling ~5.5 KB with **0 occurrences of `logprob`, `top_logprobs`, or a `bytes` array**. Reloading the page drops the records while the conversation survives — exactly what "held for this session only" should look like from the outside.

### Smoke suites on macOS

All 38 run: **31 pass, 7 fail**. All 7 fail *identically* on a clean `origin/main` worktree (`20ad6b58`) with the same node\_modules, so **this PR introduces no regressions**. Five of them need a live server or browser that I did not start; `smoke:ghost-moe-execution` and `smoke:gemma4-mtp12-target-render` are genuine pre-existing assertion failures on main, and neither is wired into `ci.yml` — worth a separate issue, not this PR's problem.

Everything this PR claims passes: `smoke:token-inspector` (33 checks, not the 29 the description says — the adversarial-review commit added four), `smoke:contrast` (258 pairs), and ui / streaming / integration / workspace / model-state / model-lanes / first-run / first-run-card / capability-readiness / arena.

One flake worth recording rather than chasing: the first `smoke:token-inspector` run exited 138 (SIGBUS) *after* printing "33 checks passed". Six consecutive reruns exited 0. The volume the worktree lives on is 98% full, so I am reading that as host environment, not the script.

### Verdict

Works on macOS. The gemma3 correction makes the surface *wider* here than the description claims, not narrower, and the fails-open handling is correct on the lanes that genuinely fail open.

PR #730, opened 2026-09-06
feat(frontend): render tool calls on the chat surface

timtoole02:
> **Overlaps #726** (structured outputs) on `App.jsx`, `MessageTurn.jsx`, `useDashboardData.js`, `styles.css` and `ChatWorkspace.jsx` — the same composer and message-panel wiring points. Either order works; whichever merges second needs a rebase.

`/api/capabilities` advertises `streaming_tool_calls` as `supported_current_gate`. The chat surface had zero handling for it — `chatCompletionStream.js` read only `delta.content`, so a tool-capable model ending a turn with `finish_reason: "tool_calls"` and empty content rendered as a **blank bubble**.

Verified live against `Llama-3.2-3B-Instruct-Q8_0`, a `tool_capable` row:

```
finish_reason: tool_calls   content: ""
tool_calls: [{ id: "call_072ec0…", type: "function",
               function: { name: "get_weather", arguments: "{\"city\":\"Paris\"}" } }]
```

Streaming carries the whole call in **one delta** — id, name and the complete argument string together — not fragmented arguments. The reader accumulates by index anyway, so a future build that *does* fragment them won't silently truncate.

## Three things this surface refuses to get wrong

**Nothing ran.** A tool call is a *request*; the model asked and the turn ended. The card never says a tool executed and offers no way to run one — a browser executing model-chosen calls against the user's machine is a different and much larger proposition. Asserted against rendered output **and** at the source: no `eval`, no `Function`, no `exec` anywhere in the lane.

**The gate is stricter than the engine's, and says so.** `POST /v1/chat/completions` gates on the chat **template** and never reads `tool_capable` — a row with `tool_capable: false` whose template happens to carry tools is accepted with a 200 and returns unchecked calls. `tool_capable` is the receipt that those calls were actually checked, so this lane requires it, *as a product choice*. An earlier draft of the copy claimed the engine enforced it; the smoke now asserts that claim is absent.

**Models loop.** Given the tool result, the 3B model re-issued the identical call instead of using it — verified live. Detection deliberately ignores the call `id`, because the runnable lane mints non-unique ids (`call_0`, `call_1`) under a constant response id; matching on id would miss every repeat there.

## Lane differences handled

The runnable/Ornith lane lifts structured calls out of the reply but **keeps the raw `<function=…>` envelope in `content`**, so one turn carries both. The capability row's "without leaking the raw envelope" holds on the dense lane only. Detected rather than assumed, and explained in the card so two views of one request don't read as two requests.

Model-written arguments are parsed defensively: a malformed argument object is reported with the raw text preserved, because a model emitting broken JSON is a real thing to know.

## Verification

18-check smoke wired into `ci.yml`, normalizing against the engine's own committed capture (`qa/capability/mac_tools_out/llama-3.2-3b/p1.json`). Build, `tool-calling`, `token-inspector`, `ui`, `integration`, `streaming`, `contrast`, `model-state`, `first-run-card` and the scrub gate all pass. Driven end-to-end in the browser.

Rebased onto `main` after #724 merged; both features verified together after resolving the overlap in `MessageTurn`, `useDashboardData`, `chatCompletionStream` and `ChatWorkspace`.

## MCP is deliberately not included

`serve` accepts no `--allow-mcp` / `--trust-mcp-server` flags, and `src/api/` contains **no MCP surface at all** — so the engine behind the browser has no MCP capability. A browser manager would mean adding MCP to the HTTP server, which changes the threat model: today an MCP server can only be trusted at the command line by someone on that machine, and `serve` can listen on a LAN. `mcp.rs` states that "off unless explicitly trusted" is *enforced, not described*. That decision belongs to you, not to a UI PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -558,6 +558,9 @@ jobs:
       - name: Retrieval rerank smoke
         if: ${{ !cancelled() && steps.build.outcome == 'success' }}
         run: npm run smoke:rerank
+      - name: Engine metrics smoke
+        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+        run: npm run smoke:engine-metrics
 
       - name: Frontend UI regression smoke
         if: ${{ !cancelled() && steps.build.outcome == 'success' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -561,6 +561,9 @@ jobs:
       - name: Engine metrics smoke
         if: ${{ !cancelled() && steps.build.outcome == 'success' }}
         run: npm run smoke:engine-metrics
+      - name: Tool calling smoke
+        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+        run: npm run smoke:tool-calling
 
       - name: Frontend UI regression smoke
         if: ${{ !cancelled() && steps.build.outcome == 'success' }}
--- a/frontend/design-evidence/BACKEND_ASKS.md
+++ b/frontend/design-evidence/BACKEND_ASKS.md
@@ -139,3 +139,62 @@ typed 400 these lanes already return for other unsupported combinations — or a
 lane axis to `api_conformance` so the refusal is at least machine-readable. Note
 that `unsupported_modes` today covers only the ROUTE axis (`streaming`,
 `multi_choice`) and says nothing about which serve lanes can honour the request.
+## 7. Host RAM total/available on an HTTP surface (Wave C, 2026-09-06)
+
+**Surface waiting on it:** Settings → Response length still renders its memory
+ceiling and projected-memory gauge ABSENT (see ask 3, 2026-06-12), and the new
+System → Engine metrics panel can show resident memory and VRAM but cannot say how
+close either is to the host's ceiling.
+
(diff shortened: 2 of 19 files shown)


timtoole02:
Rebased. This branch now sits on top of #729's two commits so CI validates the combined tree — the diff shown here will collapse to this PR's own commit once #729 merges. Merge order is #729 then #730.

Seven conflicts, four of them non-trivial:

- `useDashboardData.js` and `MessageTurn.jsx` both import `tokenInspection` / `TokenInspectorCard` on both sides. Keeping both would have re-declared the binding, so the superset import wins and the narrower copy is dropped.
- `useDashboardData.js` — the structured-record `if` and the tool-signature `if` share one closing brace in the context below the conflict. Keeping both without an extra `}` nests the second inside the first.
- `ChatWorkspace.jsx` — same shape in JSX: the conflict cuts mid-element and the trailing `</p></div>)}` closes whichever block lands last. The structured editor is now closed explicitly before the tool editor opens.

`package.json` and `MessageTurn.jsx` are mixed-EOL files, so the resolution was byte-preserving — MessageTurn's CRLF count went 212 → 211, which is the one deleted duplicate import and nothing else.

All 25 frontend smokes CI runs pass locally, including `workspace-visual` against a preview server.

PR #731, opened 2026-09-06
fix(api): stop the dense streaming lane duplicating a reply on a straddling stop

timtoole02:
Fixes a pre-existing defect on the **dense** streaming lane: a stop string that straddles a token boundary makes both streaming jobs re-emit the entire reply as one delta, on top of everything already streamed.

> Reuses the `stop_safe_stream_len` helper introduced by #725, now merged. Based on `main`.

## The defect

`streamed_text` after step n-1 is the UNTRUNCATED decode, which by construction contains no stop match. So a match found at step n must END past `len(text_{n-1})` but may START before it. `truncate_at_stop_sequence` cuts at the match START, so the truncated text is strictly SHORTER than what the client already holds, `strip_prefix` returns `None`, and the `unwrap_or_else(|| text.clone())` fallback sends the whole reply again. `chatCompletionStream.js` concatenates deltas verbatim, so the user sees the prefix twice.

Both jobs carry it — `run_stream_decode_job` and `CooperativeStreamDecodeJob::step` — and both are production-reachable (cooperative when `continuous_batch_slots() > 1` and CUDA resident decode is inactive, exclusive otherwise).

## Captured on the wire

Tiny fixture, `stop: "><"`, `max_tokens: 2`. The sequence is absent from `"<unk>"` and present in `"<unk><unk>"` starting at byte 4 — behind the end of what was already sent:

```
delta 1: {"text":"<unk>"}    <- full, untruncated
delta 2: {"text":"<unk"}     <- the entire truncated reply, re-emitted
client:  "<unk><unk"         vs non-streaming "<unk"
```

That is the actual SSE body from the parent commit, printed by the new test's failure.

## The fix

Reuses `stop_safe_stream_len` from #725. While generation continues the visible text is bounded at the stop-safe length, so bytes a later cut would remove are never streamed. On the terminating step the cut is settled and the whole truncated text is released. `is_final` is computed **before** the emit, from the same condition the loop tail already uses, so the release lands in the same delta rather than a step later.

Three supporting pieces, each closing a way the tail could still be lost or duplicated:

- **`stream_delta`** replaces the `unwrap_or_else(|| text.clone())` fallback. With stop sequences live a failed `strip_prefix` now yields nothing rather than the whole reply — re-sending *is* the bug, and the diff rests on detokenization being append-only, so a violation should degrade to silence rather than duplication. With no stop sequences the original fallback is preserved byte-for-byte.
- **`stop_hold_back_flush`** releases a held tail at every terminal exit that does not re-decode: `finish_clean` (the cooperative job's single clean-termination funnel, also reachable from the top of `step`) and both timeout paths. Without it, a run ending while a proper prefix of a stop string is held — `max_tokens` reached, or a timeout — silently drops those bytes.
- **The shared streaming prologue** seeds `streamed_text` from a prompt-cache hit and needed the same bound, or the first post-prologue step diffs against an untruncated seed and reproduces the defect.

## Why it shipped

`streaming_completion_honors_stop_sequence_finish_reason` asserts only `finish_reason` and `[DONE]`. It never concatenates the deltas, so the duplication was invisible to it.

The new `streaming_completion_stop_straddling_a_token_matches_the_nonstreaming_answer` reassembles the SSE stream exactly as a client does and pins it to the non-streamed answer for the same request. **Verified to fail on the parent commit and pass here** — the failure output above is from that run.

## No-op without `stop`

`stop_safe_stream_len` returns `text.len()` on its first line for an empty set, `truncate_at_stop_sequence` is identity, `stop_hold_back_flush` returns empty, and `stream_delta` keeps the original fallback. No certified no-stop request can take a different branch, which is what keeps the hash-pinned evidence on this lane untouched.

## Gates

3054 tests pass, 0 failures. `clippy --all-targets -D warnings` clean, `fmt --check` clean, 47/47 validation scripts, ledger drift clean, public scrub clean. Renderer seal re-pinned after fmt; `render_smollm3_production_chat_prompt` is byte-identical to the sealed blob and the SmolLM3 mention count is unchanged, so the churn is mechanical.

### DIFF
--- a/qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json
+++ b/qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json
@@ -27,7 +27,7 @@
   },
   "implementation": {
     "source_file": "src/api/mod.rs",
-    "source_git_blob_sha1": "a255d6081d1a12d8e876ed5db95ee4413618d4fb",
+    "source_git_blob_sha1": "7648cce1e54090ffa0baac909a58f2ce3cca798b",
     "architecture": "smollm3",
     "renderer": "render_smollm3_production_chat_prompt",
     "public_chokepoints": [
--- a/scripts/hf-qualification-smollm3-chat-parity.mjs
+++ b/scripts/hf-qualification-smollm3-chat-parity.mjs
@@ -93,11 +93,11 @@ const GROUNDING_FILES = Object.freeze({
   }),
   runtime_envelope: Object.freeze({
     path: 'qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json',
-    sha256: 'c0ce4c09a7e74e3710a22944c2485290bf69d602fadcc788666695c53e232689',
+    sha256: 'df191140b362b1c02946d54eeef977d6a91673654748a60061a3cf5872151ccf',
   }),
 })
 
-const RENDERER_GIT_BLOB_SHA1 = 'a255d6081d1a12d8e876ed5db95ee4413618d4fb'
+const RENDERER_GIT_BLOB_SHA1 = '7648cce1e54090ffa0baac909a58f2ce3cca798b'
 const SHAPE_CASE_ID = 'default_think_single_user_generation_prompt'
 const NORMALIZED_PROMPT_UTF8_BYTES = 1_392
 const NORMALIZED_PROMPT_SHA256 = '7619416ae94ba9a00378d976bfa944f5ba726747f9b67ba4e862d9a7fe20e4f1'
--- a/scripts/test-hf-qualification-smollm3-chat-parity.mjs
+++ b/scripts/test-hf-qualification-smollm3-chat-parity.mjs
@@ -114,7 +114,7 @@ assert.deepEqual(TEMPLATE_IDENTITY, {
 assert.equal(GROUNDING_FILES.shape_pack.sha256,
   'd46794448b7c2585d0aa83dfd7bb17d4904c2dcbc048ade1ef68cd3863166de6')
 assert.equal(GROUNDING_FILES.runtime_envelope.sha256,
-  'c0ce4c09a7e74e3710a22944c2485290bf69d602fadcc788666695c53e232689')
+  'df191140b362b1c02946d54eeef977d6a91673654748a60061a3cf5872151ccf')
 assert.equal(LLAMA_PIN.executable_sha256,
   '6c787bf07ac1d7e1bbaa1ee176c3ef0df58ea86494c8c1b1d2d9f4a9176b19ae')
 assert.equal(LLAMA_PIN.server_impl_sha256,
@@ -331,7 +331,7 @@ assert.equal(classifySmolLM3ChatParityError(sharedSmolError).error_code,
 
 const committedGrounding = await inspectGroundings(root)
 assert.equal(committedGrounding.renderer_git_blob_sha1,
-  'a255d6081d1a12d8e876ed5db95ee4413618d4fb')
+  '7648cce1e54090ffa0baac909a58f2ce3cca798b')
 assert.equal(committedGrounding.shape_case.normalized_prompt_sha256,
   '7619416ae94ba9a00378d976bfa944f5ba726747f9b67ba4e862d9a7fe20e4f1')
 const shapePack = JSON.parse(await readFile(resolve(GROUNDING_FILES.shape_pack.path), 'utf8'))
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -23345,7 +23345,7 @@ fn stream_prompt_cache_prologue(
     // with correct usage but no content delta. Longer cached streams also must
     // establish `streamed_text` before subsequent deltas are diffed.
     if !generated.is_empty() {
-        let mut text = match prepared.tokenizer.decode(generated, true) {
+        let text = match prepared.tokenizer.decode(generated, true) {
             Ok(text) => text,
             Err(err) => {
                 send(StreamDecodeEvent::Failed {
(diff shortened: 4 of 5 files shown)


timtoole02:
Reviewed. The fix is sound and the reasoning behind `stop_safe_stream_len` holds up under proof. One correctness gap, one coverage gap, one question.

## Blocker: the exclusive job never releases the hold-back on its clean exit

`CooperativeStreamDecodeJob::finish_clean` flushes on every clean termination. `run_stream_decode_job` has no equivalent — it releases by computing `is_final` *inside* the decode loop, so the release only happens if the loop body runs. It does not always run:

```rust
for _ in generated.len() as u32..prepared.max_tokens {
```

`stream_prompt_cache_prologue` samples one token **before** that loop on an exact prompt-cache hit. With `max_tokens: 1` the range is `1..1` — empty. The job falls straight through to `Finished` with the held bytes still held, and the client never receives them.

Verified, not argued. A unit test driving `run_stream_decode_job` directly, with the stop sequence derived from the sampled token's own text (its last character plus one the vocabulary cannot produce, so it is a proper prefix that is withheld and never a complete match that would legitimately cut):

```
thread 'api::tests::exclusive_stream_job_releases_the_stop_hold_back_on_a_cache_filled_budget'
panicked at src\api\mod.rs:31358:9:
assertion `left == right` failed: the exclusive job reached Finished still holding the stop hold-back
  left: "t"
 right: "t1"
```

One byte, withheld and never released.

### Reachability

Narrower than it first looks, and worth stating precisely. `stream_completion` picks the exclusive lane when `continuous_batch_slots() <= 1` **or** the CUDA resident engine is driving decode — but the prologue skips the cache lookup entirely on that CUDA path (`!crate::inference::resident_decode_cuda_active()`). So the two conditions can only meet through `CAMELID_CONTINUOUS_BATCH_SLOTS=1` on a CPU or Metal host. That is a supported setting, bounded `1..MAX` in `runtime_config.rs`, and the cooperative lane already guards the case — so the two lanes disagree about a request they should answer identically.

### The fix

Placed after the loop and before the terminal bookkeeping, matching where `finish_clean` puts its release:

```rust
// The exclusive twin of `CooperativeStreamDecodeJob::finish_clean`'s release.
// `is_final` only reaches the emit when the loop BODY runs, and it does not
// always run: `stream_prompt_cache_prologue` samples one token on an exact
// prompt-cache hit, so with `max_tokens = 1` the loop is `1..1` and the job
// arrives here with the hold-back still held. A no-op when the emit already
// released it, and empty when no stop sequences were requested.
let pending = stop_hold_back_flush(
    &prepared.tokenizer,
    &generated,
    &streamed_text,
    &prepared.stop_sequences,
);
if !pending.is_empty() {
    streamed_text.push_str(&pending);
    send(StreamDecodeEvent::Delta(pending));
}
```

I checked the other exits so this is neither over- nor under-scoped. The two early `return`s inside the loop are client hangup and cancellation, where withholding is correct because nobody is listening. The timeout paths this PR already handles. The top-of-loop `break` on `finish_reason != "length"` can only follow a prologue that already ran the full truncate. The clean fall-through was the only uncovered one.

Gates on the patched branch: `--lib` 2347 passed / 0 failed / 112 ignored, `clippy --lib --all-features -D warnings` clean, `fmt --check` clean. Same test RED before the patch and GREEN after, same build.

## The exclusive lane has no stop-sequence coverage at all

`streaming_completion_stop_straddling_a_token_matches_the_nonstreaming_answer` goes through `/v1/completions`. With `DEFAULT_CONTINUOUS_BATCH_SLOTS = 2` and no GPU on CI, that dispatches to the **cooperative** job every time. Nothing under `tests/` sets `CONTINUOUS_BATCH_SLOTS`, and nothing anywhere drove `run_stream_decode_job` with a non-empty `stop_sequences` before this.

So neither the duplication this PR fixes nor the fix itself is exercised on the exclusive lane. That is how the gap above survived a change that touched both jobs deliberately. A unit test that calls the job by name — rather than an integration test that would have to mutate a process-global env var and race the rest of the suite — seems like the right shape for that lane.

## Question: `first_content_ms` on a flush-only reply

I deliberately did **not** set `first_content_ms` in the flush, to keep the lanes symmetric — `finish_clean` does not set it either. But both lanes can now deliver a reply whose only content arrives via the flush: if the prologue's visible text is entirely withheld, `if !text.is_empty()` skips the send and leaves `first_content_ms` at `None`, and the flush is the first and only delta. Telemetry then reports `ttft_ms: None` for a request that did return content.

Fixing that on one lane only would create a fresh asymmetry, which is the same class of problem as the finding above, so I left it alone. Worth deciding for both lanes together.

## What holds up

- **The monotonicity invariant is provable.** If `T' = T + suffix` had an earlier complete match at `j < L(T)`, that match either fits inside `T` — so `complete(T) <= j` — or straddles its end, making `T[j..]` a non-empty proper prefix of the stop string so `held(T) >= T.len() - j`. Both give `L(T) <= j`, contradiction. So `stop_safe_stream_len` is non-decreasing under appending, the emitter is never asked to un-send a byte, and the final `truncate_at_stop_sequence` is never shorter than what was already streamed — which means `stream_delta` cannot silently drop on the append-only path either.
- **Char-boundary safety.** Both limits land on boundaries — `str::find` by construction, and `text.len() - held` because a byte-equal tail of valid UTF-8 starts on one — so the `String::truncate` in `stop_visible_text` cannot panic.
- **Widening the truncate to `is_final` is behavior-neutral.** `contains_stop_sequence` uses `contains` and `truncate_at_stop_sequence` uses `find`; they agree, so `finish_reason == "length"` implies no complete match and the truncate is identity. No request changes branch.
- **The renderer re-pin is mechanical.** I extracted `render_smollm3_production_chat_prompt` from both blobs: byte-identical, same hash. Only the whole-file pointer moved, and `NORMALIZED_PROMPT_SHA256` is untouched.

PR #756, opened 2026-09-11
feat(fabric): read foreign local engines honestly, and compare them

karan68:
## What this is

Camelid can already place work across its own nodes. This teaches the fabric to **read** the other engines a developer already runs — Ollama and LM Studio — without pretending it can do things it cannot, and adds the view that turns the resulting ambiguity into evidence.

**Default routing is unchanged.** The fabric still places only on Camelid nodes. Foreign engines are read and reported, never routed to.

## The engine seam

A node is written `LABEL=[ENGINE://]HOST[:PORT]`. No scheme means `camelid`, so every existing specification means exactly what it did.

| Engine | Scheme | Default port | Placed on |
|---|---|---:|---|
| Camelid | `camelid://` (or omitted) | 8181 | yes |
| Ollama | `ollama://` | 11434 | **no — read and reported only** |
| LM Studio | `lmstudio://` | 1234 | **no — read and reported only** |

The engine is **declared, never detected**. Probing an address and inferring the engine from what answered would make a typo look like a different product, and would present this fabric's bearer token to whatever was listening.

Placement ranks nodes on reported load, re-places a typed queue-full refusal on a sibling, and honours session affinity. Neither foreign API answers those three, so routing to one would mean ranking a node whose load was never measured, relaying a refusal that might have been retried, and claiming a warm prefix nobody attested.

## Unknown is a value

Ollama exposes no queue depth anywhere in its API, so such a node reports **no load at all** rather than a zero — `fabric status` prints `?`, and the health payload omits the fields entirely rather than sending `0`. LM Studio publishes no version endpoint, so its version is reported as unknown rather than guessed from a runtime name that describes a llama.cpp build instead of the application.

Models only an unplaceable node holds are deliberately absent from `GET /v1/models`: advertising a model the fabric would then refuse to route is worse than not advertising it.

## Capability matrix with provenance

Every capability answer carries how it was obtained:

| Provenance | Meaning |
|---|---|
| `measured` | Exercised against **this exact engine version** and recorded |
| `declared` | Follows from the documented API surface |
| `not_probed` | Nobody has checked; `supported` is `null` |

`not_probed` is never rendered as a no. A measurement is bound to the version string the node reported, so a neighbouring version inherits nothing from it — an Ollama node reporting `0.33.3` shows `not_probed` for tool calls even though `0.33.1` was measured.

`placement_blockers` is derived from the same entries the matrix renders, so the reason shown can never drift from the verdict it explains.

## `fabric compare` — the divergence view

The same prompt on two named nodes, via CLI or `POST /v1/fabric/compare`. It is **measurement, not placement**: both nodes are named rather than chosen, nothing fails over, and a foreign engine may be compared even though the fabric will not route to it.

Four rules decide what may be concluded:

- **Each side runs more than once.** A difference between two nodes means nothing until each node has been shown to agree with itself. If one does not, the verdict is `not_attributable` and **no diff is rendered** — a diff beside that verdict invites exactly the reading it refused.
- **Model identity is checked before stability**, so a model swap is never reported as divergence.
- **Asked is not applied.** LM Studio's documented completion API has no seed parameter, so a comparison against it is unseeded however the flag was set, and the uncontrolled parameter is named in the result.
- **Correctness is never awarded.** No verdict, reason or rendered line ranks the two sides.

Where obtainable, the chat template each backend applied is captured — `GET /props` for Camelid, `POST /api/show` for Ollama — which is usually the actual explanation. LM Studio exposes none, and says so.

## Model identity, and why a human has to assert it

Ollama always suffixes `:latest`; LM Studio never does; neither publishes a digest this fabric could compare. Exact-match identity therefore refuses **every** cross-engine comparison on a real machine.

Loosening it to strip suffixes was rejected: inferring that two names mean the same weights is precisely the guess this change refuses to make. Instead `--left-model` / `--right-model` let an operator state the equivalence. The comparison then carries `model_identity: asserted_by_operator`, adds `model identity` to the uncontrolled list, and prints `MODEL IDENTITY ASSERTED, NOT VERIFIED`. Without it, differing ids are still refused. The UI does the same: it offers the pairing and keeps **Compare** disabled until a person ticks the claim.

## An honest Cluster view

The previous one generated a sample fabric, drew nodes from browser storage, and offered worker start/stop controls that controlled nothing. All deleted. The replacement renders only fields present in a real response, and distinguishes **"this proxy withholds its node detail"** from **"this fabric has no nodes"** — the two used to look identical.

## Evidence

| Gate | Result |
|---|---|
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo test --lib fabric::` | **301 / 0** |
| `cargo test --test fabric_serve` | 74 / 0 |
| `cargo test --test fabric_end_to_end` | 17 / 0 |
| `cargo test --test fabric_engines` | 15 / 0 |
| `cargo test --bin camelid` | 55 / 0 |
| `npm run build` | clean |
| 4 smokes (2 unit, 2 puppeteer) | **89 checks**, all wired into CI |

**Ablations.** Each honesty rule was broken deliberately and had to be caught by a named test before it was believed — eight ablations, seven guarded ones caught. The eighth (a node that does not hold the requested model) is only observable against a live node, and is **declared unguarded offline** rather than quietly counted.

**Live receipts** against real Ollama 0.33.3 and LM Studio:

- *Negative control:* one server under two labels answers `IDENTICAL`. This has to hold before any divergence claim means anything.
- A node that does not hold the model is refused by name, listing what it does hold.
- A cross-engine `DIVERGENT` verdict: both sides self-consistent across two runs, distinct sha256, rendered diff, one template captured and one honestly reported as unobtainable.

## What this does not claim

The cross-engine receipt is **not** a reproduction of the previously measured template divergence (Camelid answering `12` where other backends answer `7`). That measurement used one pinned GGUF on every arm; here the weights are only *asserted* equivalent, both engines answered `12`, and the difference is in surrounding prose. Reproducing it needs the same pinned file on both engines.

Two defects these receipts caught, fixed here rather than shipped:
1. The comparison view was reading a **different proxy address** than the Cluster view — two screens would have reported on different fabrics while looking like one product.
2. LM Studio's llama.cpp **runtime string was briefly reported as its engine version**, re-creating exactly the conflation the capability work forbids.

Both are now pinned by tests.

## Review notes

- No new dependencies. The line diff is owned rather than pulled in; `sha2` was already a dependency.
- `src/fabric/policy.rs` **names no engine at all** — that is the seam claim, and it is verified by inspection each phase.
- Readiness is `any(is_placeable)`, not `ready > 0`: a proxy with only foreign nodes answers `503`, because otherwise a load balancer would keep it in rotation while every request fails.
- `POST /v1/fabric/compare` is behind the client key and caps prompt size, repetitions and tokens — unlike the read routes, it *causes work* on two machines.


karan68:
@timtoole02 — would appreciate your eyes on this one, and I could use help finishing it.

## What this PR is for

**Making Camelid the honest control plane for every local engine on a developer's machines, including the ones we didn't write.**

A developer already running Ollama or LM Studio should be able to point Camelid at them and get truthful answers to three questions: what have I got, what can each of them actually do, and **do they agree with each other?**

The differentiator isn't routing — anyone can proxy to Ollama. It's that this tells you **when your backends disagree and why**, and refuses to state anything it hasn't established, including about itself. Default routing is unchanged: the fabric still places on Camelid nodes only unless explicitly told otherwise.

## What's committed, one line each

| | |
|---|---|
| **P0** | Baseline locked at `6618ffb7` before any change, so later numbers have something to be compared against |
| **P1** | Deleted the sample-fabric generator, the storage-derived `live` chip and worker buttons that controlled nothing; the Cluster view now renders only fields present in a real response, and tells *"this proxy withholds node detail"* apart from *"this fabric has no nodes"* |
| **P2** | Engine seam — `LABEL=[ENGINE://]HOST[:PORT]`, engine **declared never detected**, Ollama read and reported but not placed on |
| **P3** | LM Studio adapter + capability matrix where every answer carries `measured` / `declared` / `not_probed`, keyed on the exact engine version |
| **P6** | `fabric compare` + `POST /v1/fabric/compare` + GUI Screen D — same prompt on two nodes, withholds a verdict unless both sides agree with themselves, never names a winner |
| **O6** | Operator-declared model identity (`alias CANONICAL=LABEL:LOCAL`), because Ollama suffixes `:latest`, LM Studio doesn't, and neither publishes a comparable digest |
| **P5** | Mixed-engine placement behind `--allow-mixed-engines`, with four guards — **core only, see below** |
| | plus a merge of `upstream/main` (270 commits, through v0.7.2) and a fix for a nav registration CI caught |

## Where I'd like help

Three phases remain and each needs verification I couldn't do well here:

- **P4 discovery** — LAN scan + confirm-before-join. Highest-risk surface in the lane: it sends traffic to machines the user didn't name.
- **P5 remainder** — GUI Screen E, relay-don't-re-place, and a live receipt of mixed placement on two real backends.
- **P7 background lifetime** — the desktop app kills the sidecar on window close, which contradicts the premise of the whole feature. Small change, but I can't verify tray/window behaviour headlessly and I won't ship it on "it compiles".

Also open: **R1**, reproducing the original template divergence by loading the *same GGUF file* into both engines rather than two files asserted to be equivalent.

## 👉 `docs/fabric-backends/AGENT.md`

I've committed a handover doc written for whoever — or whatever — picks this up. **If you put an agent on this, point it there first.** It contains the end goal, the success parameters, the invariants, exactly what's done and what isn't, the remaining work with precise exit criteria, and the testing standard.

Two sections in it matter more than the rest:

- **The testing standard.** Passing tests aren't the bar; a test that *would have failed if the code were wrong* is. Every honesty rule here was deliberately broken and had to be caught by a **named** check before I believed it — 8 ablations, 7 caught, and the 8th is declared unguarded rather than quietly counted.
- **The traps.** Eight ways the apparatus produced a confident, green, wrong answer during this work — a 1 MiB `execFileSync` buffer that made a compile failure look like "the rule is unguarded", a killed ablation run that left the tree sabotaged, a log filter that ate every added line of a diff. Worth a read before trusting any harness in this repo.

## What to build and check

```bash
cargo fmt --all -- --check                  # 0
cargo clippy --all-targets -- -D warnings   # 0
cargo test --lib fabric::                   # 322 / 0
cargo test --test fabric_serve              # 74 / 0
cargo test --test fabric_end_to_end         # 17 / 0
cargo test --test fabric_engines            # 15 / 0
cargo test --bin camelid                    # 61 / 0

cd frontend && npm ci && npm run build
npm run smoke:fabric-model                  # 27 checks
npm run smoke:fabric-view                   # 26 checks
npm run smoke:divergence-model              # 18 checks
npm run smoke:divergence-view               # 18 checks
```

To see it end to end with an Ollama you already have:

```bash
cargo run --bin camelid -- fabric status --node studio=ollama://127.0.0.1:11434
# reports it, prints `?` for load rather than 0, and refuses to route to it

cargo run --bin camelid -- fabric compare \
  --node a=ollama://127.0.0.1:11434 --node b=ollama://127.0.0.1:11434 \
  --left a --right b --model <one-you-have> --prompt 'What is 7 plus 5?' --repeat 2
# negative control: one server under two labels must answer IDENTICAL
```

The seam claim is mechanically checkable — this should print nothing:

```bash
sed -n '1,/^#\[cfg(test)\]/p' src/fabric/policy.rs | grep -nE 'Ollama|LmStudio'
```

Placement contains no engine name anywhere; every eligibility rule is written against capabilities, so a fourth backend is a new row rather than a new argument.

## What this PR does *not* claim

Stated up front so nobody has to find it by reading code:

1. **The template divergence is not reproduced.** The original measurement (we answer `12` where llama.cpp/Ollama/LM Studio answer `7`) used one pinned GGUF on every arm. Here the weights are only *asserted* equivalent, both engines answered `12`, and the difference was in surrounding prose.
2. **Mixed routing has no live receipt.** The policy is proven offline only.
3. **One ablation (D5) is unguarded offline** — it needs a live node — and the harness prints it as unguarded instead of counting it.
4. **LM Studio's undocumented per-model `capabilities: ["tool_use"]` is deliberately unused.** It's a vendor declaration about a model, not a measurement of the engine.

Two defects the live receipts caught in my own work, fixed here rather than shipped: the Compare view was reading a *different proxy address* than the Cluster view, and LM Studio's llama.cpp runtime string was briefly reported as its engine version. Both now pinned by tests.

timtoole02:
## Review of #756, and the fixes on top of it

@karan68 — the honesty machinery here is well built, and the handover doc is what made this reviewable: unknown load really is absent on the wire rather than zero, capability answers carry their provenance keyed to the exact engine version, the verdict order is right, and the `policy.rs` grep comes back clean.

We reviewed it, then **built the PR head and ran it against real engines**. Two blockers and six majors reproduced live. All of them are fixed on top of your branch, each with a regression test that fails on the old code, plus the ablation you'd ask for. Live receipts throughout; where something is still unproven it is listed at the end rather than implied.

### What the live runs found (all fixed)

1. **The fabric's credential reached a foreign engine.** With `--allow-mixed-engines`, a stand-in Ollama recorded `Authorization: Bearer <token>` on `POST /v1/chat/completions` — and with no `--bearer` at all it received the operator's exported `CAMELID_API_KEY`. Fixed at the lowest chokepoint in `forward.rs`, so no send path can do it; the test stub now records headers, which is why this slipped through before.
2. **The proxy sent no CORS, so neither new view could read it.** The UI is served from :8181 and the proxy listens on :8282; `GET /v1/health` was unreadable cross-origin and the Compare preflight got a 405. The smokes passed only because their fake proxies added headers the real one never sent. Fixed with `fabric serve --cors-origin` (off by default, reusing the engine's validated origin list), and the views now show the exact command when an origin is refused.
3. **Compare died after 2 s through the proxy**, because generations were given the health-probe budget. Fixed: generations use the forward budget.
4. **Compare only worked on loopback** — its probe ignored the configured transport and refused a LAN node while telling the operator to pass the flag they had already passed.
5. **A DIVERGENT verdict could show a diff with no changed line**: the verdict hashes bytes, the diff split with `str::lines()`, so a trailing-newline or CRLF difference vanished. Diff lines now carry their terminator.
6. **`--allow-mixed-engines` switched off the readiness check**, so a node that was down became eligible and a retryable failure was reported as "model unavailable".
7. **The model each node reported was never read**, so the `DifferentModels` verdict could not fire in production.
8. **The Compare view** rendered failures as "no nodes", dropped the operator's identity claim, sent no client key, and never disclosed the 64-token cap.

### What the same runs taught us about the product

- **R1 is reproduced.** With one pinned GGUF, Camelid answers `Hi!` where llama.cpp (jinja) and Ollama answer `Hi. How can I assist you today?`. llama.cpp with `--no-jinja` renders exactly what Camelid renders and answers `Hi!`, so **the template is the cause**: Camelid drops the system block the GGUF template always emits. That is an engine defect, tracked separately, not a fabric one.
- **Compare showed the *advertised* template as though it were the one applied.** Both sides advertised byte-identical templates while rendering differently. It now captures Camelid's rendered prompt via `POST /apply-template` and labels advertised versus applied, and says so when identical templates cannot explain a divergence.
- **Ollama's answers depend on its request history.** Same weights, temperature 0, seed 42: a fresh server answered one way, and after unrelated traffic it answered another — so "each side agreed with itself across N back-to-back runs" can be passed by luck. Measured: 10 runs per arm gave **8 confident-but-wrong DIVERGENT verdicts** before, **0** after a short perturbation between repetitions (8 honest "not attributable", 2 identical).
- **Ollama re-serialises a GGUF on `ollama create`**, so its blob digest never equals the file's. Identity by digest therefore verifies only where both sides publish a loaded-file digest; across engines an operator's assertion still stands, with both digests disclosed. Same-engine digest mismatches remain `different_models`.

### Also on this branch

- **P5 (mixed placement):** `fabric serve --allow-mixed-engines` (startup flag only — nothing could authenticate a runtime switch), residency- and alias-aware matching, typed backpressure trusted only from engines that declare it, a Routing section rendering only proxy-supplied strings, and the amended I12 written down.
- **P7 (desktop lifetime):** closing the window quits again — it had stopped doing so in v0.7.0, when a hidden Spotlight window began keeping the app alive with an invisible engine. Background mode is an explicit tray toggle; the tray states the observed port and model, and offers Open, Restart and Quit. Two defects were found by driving the real app: a `kill -9` left an orphaned engine (the watcher's exit message panicked on a broken stderr pipe), and the one-time notice was recorded as shown without ever being displayed. Both fixed and re-verified live.

### Not claimed

- **P4 discovery is not delivered.** A partial implementation exists on a WIP branch, ungated and unreviewed.
- **The relay-don't-re-place receipt against real Ollama is still owed.** Three attempts failed to saturate a 1B model's queue on this hardware; the stand-in receipt stands, disclosed as such.
- **P7 on Windows is unverified** (no Windows machine here), and the background-mode checkbox was verified by its rendered state, not by clicking it.
- Two `fabric::http` TLS tests fail on our second Mac from a pre-existing dual-stack defect unrelated to this PR, and one clippy lint in `src/metal.rs` fires only under an older toolchain than CI pins.

### Where the code is

Everything above is on **`fix/pr756-integration`** in `timtoole02/Camelid` — 43 commits on top of this PR's head (`f499d5d3`), with `origin/main` merged in. **Your branch is untouched**; nothing here rewrites your work, so take it however suits you:

```bash
git fetch origin fix/pr756-integration
git log --oneline f499d5d3..origin/fix/pr756-integration   # each fix is its own commit
git merge origin/fix/pr756-integration                      # or cherry-pick what you want
```

The partial P4 is on **`wip/pr756-p4-partial`** — ungated, unreviewed, no ablations, no LAN receipt. It is a starting point against your P4 spec, not working code.

Gates on the merged head: `fmt` clean, clippy clean (bar one pre-existing `metal.rs` lint that fires only under an older toolchain than CI pins), `cargo test --all-targets` 71 suites / 0 failed (lib 2704, fabric_serve 94, fabric_engines 39, end-to-end 18), desktop 48+10+14, frontend build plus every CI smoke passing. `docs/fabric-backends/AGENT.md` has the measured counts and the full "what is NOT claimed" list.

PR #757, opened 2026-09-11
feat(cuda): true continuous batching through eight requests

karan68:
## Goal

Camelid already had a fair scheduler, but CUDA requests still ran one model forward at a time. The goal of this work was to make one CUDA forward process several independent requests together without mixing their KV history, output, cancellation state, or memory ownership.

The product gate was deliberately concrete:

- eight equal requests finish in less than twice one-request wall time;
- the slowest first-token latency stays within twice the fastest;
- a short request can finish ahead of an older long request;
- every request matches its serial result;
- cancellation, overload, and cleanup remain bounded and observable.

## What was implemented

- **Phases 0–1: baseline and semantics.** Measured the old serialized behavior, froze thresholds, and added a stable row/result contract plus serial oracle.
- **Phases 2–3: safe batch two.** Added generation-safe CUDA sequence leases, isolated KV ownership, and the first true two-row CUDA forward.
- **Phases 4–5: serving and fairness.** Wired dynamic HTTP grouping, stable result scatter, per-request cancellation, and chunked prefill so a large prompt cannot monopolize the engine.
- **Phases 6–7: paged KV and batch eight.** Added transactional GPU pages with prepare/commit/abort, immediate reclamation, true batches from 2 through 8, and cross-request batched prefill.
- **Phase 8: two resident models.** Added an explicit capacity-two CUDA arena with exact artifact identity, active pins, deterministic idle LRU, and exact per-model unload.
- **Phase 9: rollout and reviewability.** Rebased onto current `main`, exposed bounded arena status in health/runtime memory/props/slots/Prometheus, updated configuration/compatibility/rollback docs, fixed the stale CI wiring guard, added packaged-artifact checks, and added normal/edge/browser evidence.

Everything remains explicit opt-in. Defaults retain one CUDA sequence and one resident main model. The support claim is limited to the documented exact Windows RTX 4060 / CUDA 12.9 / 512-position model envelope; this does not claim arbitrary models, GPUs, drivers, longer context, MoE/windowed/sharded batching, portable throughput, or default enablement.

## Measured result

On the selected Phase 7 release gate:

- eight-request wall ratio: **1.708x** (required `< 2x`);
- slowest/fastest TTFT ratio: **1.286x** (required `<= 2x`);
- **59** true size-8 decode forwards;
- **44** cross-request size-8 prefill forwards;
- exact serial output and completion-count parity;
- short request finished before the active long request;
- ninth request returned typed `503 no_available_slot` with `Retry-After: 1`;
- one cancellation released capacity while all seven siblings completed;
- 48/48 churn requests completed;
- final allocated pages, device KV bytes, and live sequences were all zero.

Phase 8's exact-model regression remained token-identical at **9.8124x aggregate batch-eight speedup** and **0.8153 single-request-equivalent wall ratio**. The two-model physical gate retained the exact 1B Q8_0 and 3B Q5_K_M artifacts with two builds and zero evictions; active unload returned typed 409, idle unload removed only its target, and the survivor generated unchanged output.

### Measured eight-request proof

Measured eight-request CUDA batching proof

### Normal and edge states

| Capacity two | Default one, mobile |
| --- | --- |
| Capacity two | Default one mobile |

| Legacy backend | One-model empty state |
| --- | --- |
| Legacy backend | One-model empty |

| Active residency | Long model ID, mobile |
| --- | --- |
| Active residency | Long ID mobile |

Runtime memory error state

The full gallery and SHA-256 manifest are in `qa/evidence-bundles/f1-phase9-rollout/`. Synthetic browser states are explicitly labeled as frontend presentation evidence; measured and live packaged-runtime evidence are identified separately.

## Validation

- [x] Rebased onto canonical `main` at `fb6245139bcdf36fa2ca3fe51a1b61d4417df5a3`
- [x] `cargo fmt --all -- --check`
- [x] strict `cargo clippy --all-targets --all-features -- -D warnings`
- [x] all-target/all-feature compile across 69 test binaries
- [x] focused Phase 8 arena and Phase 9 API/manifest tests
- [x] frontend Arena smoke, broad UI regression smoke, production build
- [x] seven-case desktop/mobile/dark/light browser matrix with overflow/overlap/ARIA assertions
- [x] public scrub, privacy audit, evidence claims, ledger schema/drift, toolchain pins, README screenshot guard
- [x] CUDA prefill CI guard updated to the actual transactional paged prefill paths
- [x] all-target/all-feature compile across 69 test binaries; the final full-suite attempt passed 2,419 tests before finding the now-fixed project-count assertion, and the corrected assertion passes focused
- [ ] extracted packaged Windows artifact static check and model-less boot/UI smoke (deferred to CI/release workflow)
- [x] retained exact-model Phase 7/8 physical release receipts, now summarized in the committed scrubbed evidence bundle

## Rollback

Stop the server, remove the Phase 7/8 environment variables, and restart. Setting `CAMELID_CUDA_SEQUENCE_SLOTS=1`, `CAMELID_CUDA_RESIDENT_MODELS=1`, `CAMELID_CUDA_PAGED_KV=0`, `CAMELID_CUDA_TRUE_BATCH=0`, and `CAMELID_CUDA_BATCHED_PREFILL=0` restores conservative behavior. Resident state is process-local; no model or conversation data migration is involved.


timtoole02 (commented):
Thanks for this. It's a big, carefully scoped piece of work. The paged-KV / sequence-lease / arena ownership model reads as sound, and the support envelope is enforced in code rather than only in docs. But the branch can't merge in its current state. CI has never compiled the Rust, and several changes alter the **default** path, which the PR says is unchanged.

The review was done by reading the code at `a92d8325`, not by running it. The two prefill blockers (3 and 4) were each traced independently more than once.

## Blockers

### 1. CI has never run `rust`, `frontend`, or `validation-scripts`
All three pushes failed `public-scrub`, so every downstream job was skipped. Two fixes are needed, in order:

- **Bundle checksums.** In `qa/evidence-bundles/f1-phase9-rollout/SHA256SUMS`, the entries for `README.md`, `manifest.json` and `measured-batch8-result.json` were computed over **CRLF** working copies, but the committed blobs are **LF**. The 9 PNG entries match. Regenerate the three text entries over the committed LF bytes, e.g. `git show HEAD:<path> | sha256sum`.
- **SmolLM3 renderer seal.** Any edit to `src/api/mod.rs` changes its git blob (`f7c4c04a` → `3cfe4ac8` here). That breaks `test-hf-qualification-smollm3-chat-parity` and `test-check-model-qualification-roster`, which both fail locally on this branch. The pin is chained, so re-pin all of these from the final `mod.rs`:
  - `RENDERER_GIT_BLOB_SHA1` in `scripts/hf-qualification-smollm3-chat-parity.mjs`
  - `source_git_blob_sha1` in `qa/model-qualification/fixtures/smollm3-default-thinking-runtime-envelope-v1.json`
  - that fixture's own LF-normalized sha256, pinned as `runtime_envelope.sha256` in the same script
  - both values again in `scripts/test-hf-qualification-smollm3-chat-parity.mjs`

### 2. `a92d8325` reverts vision work that is already on `main`
Compared with `fb624513`, this commit removes every `vision_token_allowance` (4 → 0) and every `IMAGE_TOKEN` reference (8 → 0) from `src/api/mod.rs`. That reverts #753/#754, most likely during conflict resolution in the health constructors and `llama_server_apply_template`.
- `/v1/health` loses `vision_token_allowance`.
- `/apply-template` no longer refuses image parts for non-vision models or before the projector is ready.
- [`frontend/scripts/web-research-smoke.mjs:656`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/frontend/scripts/web-research-smoke.mjs#L656) will fail once CI reaches it.

Suggested fix: restore main's versions of those hunks.

### 3. K-quant and offloaded CUDA models lose GPU prefill by default
On `main`, when `serial_prefill` was set, or when batched prefill was unsupported, the resident path ran the **serial GPU prefill** (`engine.prefill_from`). `serial_prefill` defaults to `!prefers_batched_prefill()`, which is true for every Q4_K/Q6_K model, and batched prefill is unsupported for offloaded models.

This PR removes the only `prefill_from` call site. That case now returns `Unsupported`, which means a **CPU** prefill:
- [`src/inference.rs:3850`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/inference.rs#L3850)
- This is a large time-to-first-token regression for, e.g., Llama-3.2-3B Q4_K_M and offloaded 8B models.
- It also changes the meaning of the `CAMELID_CUDA_RESIDENT_PREFILL_BATCHED=0` A/B switch. `check-cuda-prefill-parity-gate.mjs` still describes it as "the serial path stays as an A/B escape hatch".

No test catches this, because every new test sets `CAMELID_CUDA_RESIDENT_PREFILL_BATCHED=1`. Please restore the serial GPU branch for the non-paged case, and add a test that runs without that override.

### 4. A warm engine with no reusable prefix fails every later GPU prefill
`main` called `slot.engine.set_filled(reuse)` unconditionally before prefilling. This PR only calls it when `reuse != base_position`, then returns `Err` if the engine's cursor doesn't equal `reuse`:
[`src/inference.rs:3893-3905`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/inference.rs#L3893-L3905)

**Scenario on the default Q8_0 CUDA path:**
1. A request's GPU prefill is skipped. Causes: a prompt of 2 tokens or fewer (`prefill_count > 1` gate), a prompt over capacity, or `CAMELID_CUDA_PREFIX_CONTINUATION=0`.
2. Decode reseeds the GPU KV. `seed_layer` clears `resident_tokens` ([`src/cuda_resident.rs:16874`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/cuda_resident.rs#L16874)), while `filled` rises to N.
3. On the next request, `resident_prefix_len` returns 0, so `reuse == base_position == 0`, `set_filled` is skipped, and the check fails with `expected GPU cursor 0, found N`.
4. That error propagates via `?` at [`src/inference.rs:6422`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/inference.rs#L6422) and fails the request.

Nothing on the single-slot path resets `filled` after that (`clear_kv_slot` only runs when sequence slots exist), so every later GPU-prefill request fails too, until the engine is rebuilt.

Relatedly, at [`:3931`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/inference.rs#L3931), a failed non-paged prefill used to reset `filled`/`resident_tokens` and return `Ok(false)` (CPU fallback). It now returns `Err` and resets neither.

Please restore the unconditional reset and the fallback on the non-paged path, and add a test for "warm engine, request with no shared prefix" and for `CAMELID_CUDA_PREFIX_CONTINUATION=0`.

### 5. `cuda_resident_arena` is never serialized, and the new release smoke requires it
No backend struct emits `cuda_resident_arena`: not `/v1/health`, not `/api/runtime/memory`, not `/props`, not `/slots`. The Prometheus gauges do exist.

[`phase9_read_only_surfaces_expose_cuda_resident_arena`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/api/mod.rs#L28408) passes only because it compares `null` with `null`. As a result, the Arena "capacity two" UI can't activate against a real server; the screenshots use intercepted responses.

More importantly, [`scripts/smoke-release-artifact.mjs:114`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/scripts/smoke-release-artifact.mjs#L114) now **requires** the object, and `release.yml` runs that smoke on four release legs. The next release cut from this tree would fail there, and PR CI never exercises it. Please add the field to both health constructors and the other surfaces, and make the test assert that it is non-null.

### 6. `release_model` returns pool memory before the registries drop (CUDA builds)
In [`release_model`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/api/mod.rs#L16799), the CUDA reset (which ends in `release_async_pool()`) now runs **before** `gemma4_runtimes`, `runnable_runtimes`, `dg_runtimes` and `cached_weights` are dropped. The post-drop reset is now `#[cfg(not(feature = "cuda"))]` ([`:16889`](https://github.com/timtoole02/Camelid/blob/a92d8325f56b85e79e774e7684d0e8490ced90e2/src/api/mod.rs#L16889)).

Memory freed by those drops goes back into the `cuMemAllocAsync` pool and is never trimmed. The next model's fit probe then sees less free VRAM than there really is, which is the host-spill slowdown the retained comment describes. A targeted release also no longer clears the drafter cache. Moving the CUDA block after the drops, or adding a final `release_async_pool()`, should fix it.

## Should-fix
- **GPU prefill / host-mirror failures now fail the request** instead of falling back to the CPU (`main` returned `Ok(false)`). A transient scratch-allocation failure now reaches the user as an error.
- **Slot capacity is decided once at engine start** (`admitted_slots`, `src/api/engine.rs:481`), but `POST /api/runtime/gpu` can switch the GPU on at runtime.
  - A CUDA build started with the GPU off gets 2 slots.
  - After the GPU is switched on, two same-model streams could run cooperatively on the single-sequence resident engine, a case `main` kept exclusive.
  - The routing was traced; KV sharing is suspected, not demonstrated.
- **Timeout semantics changed on the default path.**
  - `stream_completion` now starts the clock at handler entry and adds a pre-`Started` deadline.
  - Cooperative jobs spend queue time out of `CAMELID_GENERATION_TIMEOUT_MS`.
  - Exclusive jobs get queue time plus a fresh full budget.
- **Opt-in path failure blast radius:**
  - One row failing to get or finalize a paged sequence returns a mid-stream 503 to its whole batched-prefill group (`step_paged_prefill_group`).
  - One failed paged prefill (`*guard = None`) tears down every live sequence.
  - A batched row reaching position 512 does the same, because the typed `PositionOutOfRange` isn't treated as fallback-safe.
- **`frontend/package.json`** is a whole-file CRLF→LF rewrite (+42/−41) around a one-line change. Please keep the original line endings.

## Evidence
- **The committed batch-8 result is a hand-written summary JSON**, and the "measured proof" screenshots are an HTML page rendered from that same JSON. There are no raw logs, per-request outputs, or serial-vs-batch token dumps behind it.
- **Its `source_commit` (`f6c546a4`) isn't in this PR.** It is a pre-rebase commit made before the Phase 8 arena commit (+912 lines of `src/`) and all of Phase 9, so it describes a different tree from the one under review.
- **The Phase 8 numbers have no committed receipt:** 9.8124× aggregate, 0.8153 single-request-equivalent wall ratio, and the two-model A-B-A-B gate. A 0.8153 ratio means eight batched requests finished faster than one request alone, so it needs an explanation of the baseline.
- **Per the checklist, the full suite never completed green** after the last fix, and the packaged-artifact check is deferred.

Could you commit raw receipts generated at the PR head, or re-run the gate and bundle them?

## Checked and looks correct
- `rope.rs` stays bit-identical for existing callers.
- The #738 lazy-mirror wiring (`set_cpu_kv_mirror_eager(speculative.is_some())`) is intact.
- Prefix continuation is correctly restricted to non-paged single-slot mode.
- Paged-KV generation checks, prepare/abort rollback and page bounds checks in the kernels are sound.
- Arena pinning and LRU are sound, and the 409 on active unload is checked under engine ownership.
- Per-row cancellation, result scatter and the `Retry-After` path are correct.
- There are no new routes, no sensitive fields on `/v1/health`, and no model labels on metrics.

Suggested order: fix 1 so CI actually runs, then 2 through 6, then refresh the evidence at the final head.

karan68:
Thanks for the review — that was a genuinely useful read. All six blockers are fixed, plus two pre-existing bugs the review indirectly surfaced. Rebased on latest main; the branch is mergeable again.

## Blockers

**1. CI.** The `SHA256SUMS` text entries were hashed over CRLF copies while the committed blobs are LF — regenerated. Re-pinned the SmolLM3 renderer seal (blob sha1 in two scripts + the fixture, then the fixture's own sha256 back in the scripts). I did that re-pin derivationally rather than by hand, because it's chained and I broke it twice by editing `mod.rs` afterwards.

**2. Vision revert.** Restored byte-for-byte from the merge base: the `vision_token_allowance` field and both constructors, the image-token constants, the projector-readiness wording, and the `/apply-template` multimodal + image refusal ladders. Counts are back to main's 4 and 8.

**3. Serial GPU prefill.** Restored. You were right that this hit every Q4_K/Q6_K model by default. `prefill_from` indexes by absolute position, so it can only serve a whole prompt — a partial chunk still declines.

**4. Warm engine.** `set_filled(reuse)` is unconditional again on the non-paged lane, and a failed whole-prompt prefill or host mirror falls back to the CPU instead of failing the request. Continuation chunks keep the strict cursor check, since their KV is mid-flight.

**5. `cuda_resident_arena`.** Now actually serialized on `/v1/health`, `/api/runtime/memory`, `/props` and `/slots`. The test was comparing two absent fields; it pins the shape now.

**6. `release_model`.** The CUDA teardown stays ahead of the registry drops so its 409 still lands while the registries are intact, and a second `release_async_pool()` runs after them to trim what those drops returned. A targeted release clears the drafter cache too.

## Should-fix

Did 1 and 3. For the slot-capacity one, `POST /api/runtime/gpu` can turn the GPU on after `admitted_slots` is frozen, so the check is now per-request rather than at engine start. For the blast radius: `PositionOutOfRange` was being flattened into `Backend(String)` by the slot wrapper, so both dispatch sites couldn't tell it from a dead kernel and tore down the whole arena — one row hitting its context limit ended every live sequence. It's typed through now and classified fallback-safe.

Left 2 (timeout semantics) alone. The pre-`Started` deadline is load-bearing for `phase7_stream_deadline_cancels_a_request_while_it_is_still_queued`, and the queue-time asymmetry between cooperative and exclusive jobs looks like a deliberate call — your view on which way it should go would help.

## Two pre-existing bugs

**The resident CUDA engine never built on CUDA 11.x.** `flash_attention_prefill_tiled_impl` used `constexpr`, which is C++11; NVRTC defaults to C++03 on 11.x, so that one line failed the entire `KERNELS` module. `CudaResidentDecode::new()` returned `Err` for every model and *every* resident GPU decode and prefill silently fell back to the CPU with nothing logged. It's in the base commit too, not new here. `HEAD_DIM` is a template param so `const int` is still an ICE — I avoided `--std=c++14` since this file's float codegen is pinned by `--fmad=false`.

It was undiagnosable because `build_resident_cuda_engine` discarded three separate errors with `.ok()?`; those now report under `CAMELID_RESIDENT_TRACE`, as does whichever gate declines a prompt.

Worth flagging: since the engine couldn't build on 11.x at all, the batch-8 numbers must have come from a CUDA 12 box, and any 11.x machine was running CPU prefill while reporting GPU timings.

**`a_changed_node_set_is_not_answered_from_the_observation_of_the_old_one` was flaky** — 3 failures in 6 runs at `--test-threads=16`. It replaced a node file with one of identical length, and `watch.rs` documents that a same-length rewrite inside one mtime tick is invisible by design. Fixture now differs in length, per that file's own rule. 10/10 now.

## Evidence

Not regenerated. `source_commit` is still `f6c546a4`, which isn't in this PR — you're right and I didn't want to paper over it with a re-run that isn't a real receipt. Given the CUDA 11.x finding I'd rather re-measure from scratch on a known-good box than re-bless the existing numbers. Happy to do that if you tell me which box is authoritative.

## Testing

The three gates you asked for are in, and they execute real GPU prefills on an L4 rather than skipping. Each is causal — re-introducing the regression fails it:

- serial lane → `the default (unset) lane must prefill on the GPU, not fall back to the CPU`
- warm engine → `RuntimeShapeMismatch("resident CUDA prefill continuation expected GPU cursor 0, found 14")`

That second one is the exact failure you predicted from reading the code.

`cargo test --all-targets --features cuda`: 2446 passed, 0 failed. `fmt` and `clippy -D warnings` clean.

The macOS leg needed several fixes of its own — CUDA-only references with no cfg guard, invisible on Linux/Windows because `build.rs` auto-enables `cuda` there, and never caught because `public-scrub` was failing and skipping the whole `rust` job on all three original pushes.

CI at the time of writing: `public-scrub`, `rust (ubuntu --all-features)`, `validation-scripts`, `frontend`, `desktop` and `setup` all green. `rust (macos)` and `rust (windows)` were still running — windows passed on the previous run and nothing since touches it. If macOS is still red I'll chase it; I have a local reproduction that matches it now (the earlier one was being truncated by the `compile_error!` in `cuda.rs`, which aborts before dead-code analysis finishes — that's why it took a few tries).

One flaky job worth knowing about independently of this PR: the frontend `Engine metrics smoke` intermittently exits 139 *after* printing "23 checks passed" — the assertions pass and node segfaults on the way out. Hit it twice, passed the other times, and it reproduces on neither this branch locally nor main.

PR #779, opened 2026-09-17
F3: session KV cache on CUDA - reuse the conversation instead of re-prefilling it

karan68:
## What this does

A chat turn used to re-prefill the entire conversation. This makes the CUDA
resident engine keep the conversation's KV on the device and reuse it, so turn N
pays for its new question and nothing else.

Measured on an L4 (release, Llama-3.2-3B-Instruct-Q4_K_M, 3.4k-token
conversation): **turn 10 prefills 340 tokens instead of 3471**, and
time-to-first-token is **7.6 s against 86.0 s** with the reuse switched off.

Camelid chat on this branch: 287 prompt tokens answered in 509 ms

That screenshot is the shipped UI talking to one `camelid serve` process running
this branch. The third turn reads `tokens in 287 · first token 509ms`. A cold
prefill on this lane runs at roughly 14 ms/token, so those 287 tokens would have
cost about 4 s. The server's own trace for that exact request says why it did
not:

```
prefix continuation: reused  21 of  22 positions, prefilled  1   ->  14 ms
prefix continuation: reused  74 of  97 positions, prefilled 23   -> 302 ms
prefix continuation: reused 263 of 286 positions, prefilled 23   -> 310 ms
```

Full bundle, including the raw trace and an API-only replay with no browser
involved: `qa/evidence-bundles/f3-cuda-session-kv-20260917/`.

## Why it was broken

The engine already knew how to skip leading positions a previous prefill had
built (`resident_prefix_len`). The problem was the record it matched against:
only a *prefill* ever wrote it. Decode advanced `filled` while the record stayed
frozen at the prompt, so turn N reused turn N-1's **prompt** and re-prefilled
every token the model itself had generated.

Two commits fix that, one per lane:

**Single sequence.** Decode now records the row it writes. The rule is narrow on
purpose: extend only a record that already accounts for exactly `[0, position)`,
over a row that is actually written. A gap leaves the record short, which costs a
re-prefill — the failure mode of the opposite choice is answering from KV that
does not hold the prompt's tokens, which is wrong output rather than a slow path.

**Paged / multi-sequence.** F1's concurrency is merged but every one of its knobs
turned this reuse off, because the record was engine-wide and paged KV keeps a
page table per sequence. The page pool already had the right primitive: `fork`
retains pages by refcount and `PendingPageKind::CopyOnWrite` copies only the page
you write into. What was missing was an owner — `release_paged_sequence` freed
every page the moment a session dropped, so the next turn had nothing to fork
however well it matched. It now retains that table as the prefix the next turn
forks from. No kernel changes.

The retained prefix behaves like a cache, not a reservation: when the pool cannot
serve a live sequence, `prepare_paged_append` evicts it and retries rather than
failing a request to protect a speed-up.

Both lanes share one matcher (`resident_prefix_len` is now `pub(crate)`) so they
cannot drift apart.

## Correctness

Reuse is content-addressed: a KV row is a pure function of the token prefix that
produced it, so rows an engine already holds for an identical prefix *are* the
rows a new prompt needs. That argument only holds if it is true byte-for-byte, so
both lanes have an exact-parity gate that runs turn 2 twice — once on the warm
engine, once on a rebuilt one — and requires the two replies to be
token-identical. Greedy decode makes any divergence in the reused K/V observable
as a flipped token at the first near-tie.

Five GPU gates, all passing on debug and release:

| gate | what it pins |
| --- | --- |
| `f3_reusing_generated_kv_answers_identically_to_a_cold_prefill` | warm vs cold, token-identical |
| `f3_paged_reuse_answers_identically_to_a_cold_prefill` | same, through a page fork |
| `f3_generated_rows_extend_the_reusable_prefix_past_the_prompt` | the decode-side record is what causes the win |
| `f3_paged_prefix_survives_the_session_that_built_it` | retention across session teardown |
| `f3_turn_ten_prefills_only_the_new_question` | one user turn prefilled per turn, never the history |

Plus `record_resident_token_extends_only_a_contiguous_accounted_record`, which is
GPU-free and runs in ordinary CI.

## What this does NOT do

**The spec's "turn-10 TTFT within 20% of turn-1" is not met. It is 1.59x.**

The gap is not the cache. With the history resident, turn 10's new tokens still
attend over 3.4k positions where turn 1's attended over 170. Fitting the ten
turns to `c + a · new_tokens · context` splits it as ~4.6 s of per-token prefill
and ~3.0 s of that extra attention.

I chased the obvious fixes and they all lose. Recorded in the doc comment on the
gate so nobody repeats it:

| prefill lane | turn 1 | turn 10 |
| --- | --- | --- |
| serial (default) | 4758 ms | **7573 ms** |
| serial + `CAMELID_FLASH_PREFILL=1` | 4767 ms | 7603 ms |
| `CAMELID_KQUANT_BATCHED_PREFILL=1` | 4083 ms | 12276 ms |
| ...plus flash, `KQUANT_BATCH_TOKENS=4` | 4949 ms | 20132 ms |

`CAMELID_FLASH_PREFILL` is **inert** on a K-quant row — the kernel is reachable
only from `run_batched_layer_stack`, and a K-quant row takes the serial lane, so
the flag cannot fire until batched prefill is also on, where it is slower again.
(`flash_prefill_enabled` asks for sm_89 evidence because none was committed; this
is that evidence, and it does not favour the flag.)

The real ceiling is that the batched K-quant GEMM stages token rows in shared
memory, so `batched_layer_token_cap` clamps the tile to 2 tokens (4 at most
here). Weight reads amortize 2-4x, so prefill runs at roughly *decode* speed.
Closing that needs a K-quant prefill GEMM with real tiles, which is a separate
piece of work.

Worth flagging: fixing it would make this gate's **ratio worse** while making
every user-visible number better, because shrinking the constant leaves the
attention term standing. The ratio is a proxy. The spec's other F3 gate — agent
TTFT p95 under 4 s — is the one that tracks what a user feels, and at 7.6 s it is
also unmet, for the same reason.

So the gate asserts a 2.20x ceiling with the measured 1.59x and the unmet 1.20x
target recorded beside it, rather than asserting a target this lane does not
reach.

## Other limits, stated plainly

- **One conversation at a time.** The device record holds a single conversation,
  so alternating between chats costs a re-prefill. The evidence bundle includes
  that case on purpose (`reused 178 of 693, 7.4 s`) rather than only the flattering
  numbers. This is the spec's "one exact active session" scope.
- **Paged reuse is bounded by `SPLITK_THRESHOLD` (512).** Above it the paged
  decode declines to the CPU. Pre-existing F1 limit, not introduced here.
- **Two-sequence KV without paging still has no reuse** — it has a second KV bank
  rather than page tables, so there is nothing to fork. It now warns once, names
  the cost, and points at `CAMELID_CUDA_PAGED_KV=1`, instead of going quiet.

## Verification

Merged with `upstream/main` (15 commits, touching all five files this changes) and
re-verified on the merged tree:

- `cargo fmt --all -- --check` and `cargo clippy --all-targets --features cuda -- -D warnings`
- `cargo test --all-targets --features cuda`: 2593 lib tests, 57 binaries, 0 failures
- the six CI node gates, including `check-cuda-prefill-parity-gate.mjs` and the
  evidence privacy audit
- the five `f3_*` GPU gates on debug and release

Production change is about 110 lines; the rest of the diff is tests and recorded
measurements.


karan68:
Hey @timtoole02 can you please review and if you , test it also on some other cases if i have missed

